PR #1, opened 2021-04-14
UI Building Blocks and Styles

alice-i-cecile:
[**RENDERED**](https://github.com/alice-i-cecile/rfcs/blob/ui-building-blocks/rfcs/ui-building-blocks.md)

A vision for the basic building blocks, data structures and interaction paradigms for Bevy-native UI. Determining a styling strategy as part of this RFC allows us to make sure the data model works in a more complex use cases.

Previously developed at: https://github.com/alice-i-cecile/bevy-temp-rfc/pull/4

### DIFF
--- a/rfcs/ui-building-blocks.md
+++ b/rfcs/ui-building-blocks.md
@@ -415,6 +415,6 @@ allowing end users to query for style parameter components directly to receive t
 
    1. Widget composition.
    2. Layout.
-   3. Wiring: call-backs, message passing.
+   3. Wiring: call-backs, message passing (see #11).
 
 This RFC defines the underlying UI data structures, so should be settled first.
--- a/rfcs/ui-building-blocks.md
+++ b/rfcs/ui-building-blocks.md
@@ -33,6 +33,12 @@ while the value returned by `.get()` controls the final behavior of the widget.
 Every style parameter has both a `MyStyle<Base>` and a `MyStyle<Final>` variant, stored together on each entity.
 When creating a new style parameter, you must ensure that it implements `StyleParam`, typically achieved with `#[derive(StyleParam)]`.
 
+Styles can be modified directly, by modifying the `Styles` component of the widgets you wish to modify.
+When spawning entities, you may find it convenient to work with the relevant `EntityCommands` methods instead
+to reduce the number of function parameters your systems need.
+Pass in the type of your resource(s) that implements the `StyleResource` trait as type parameters, and the next time commands are processed,
+the entity stored in that resource will be added to (or removed from) the `Styles` component of your widget entity.
+
 When working on complex games or applications, you're likely to want to group your styles into **themes**,
 automatically applying them to large groups of widgets at once.
 In Bevy, themes are applied by adding a generic system that corresponds to that theme to your app's schedule.
@@ -47,15 +53,15 @@ This can work well for quickly toggling between various themes, like you might s
 ### Example: Building a Simple Widget
 
 ```rust
-commands.spawn_bundle(ButtonBundle::default());
+commands.spawn_bundle(ButtonBundle::default()).add_style::<BoldStyle>();
 ```
 
 ### Example: Building a Compound Widget
 
 ```rust
-commands.spawn_bundle(ButtonBundle::default())
+commands.spawn_bundle(ButtonBundle::default()).add_style::<BoldStyle>()
   .with_children(|parent| {
-    parent.spawn_bundle(TextBundle::default())
+    parent.spawn_bundle(TextBundle::default()).replace_style::<BoldStyle, SubtleStyle>()
   });
 ```
 
--- a/rfcs/ui-building-blocks.md
+++ b/rfcs/ui-building-blocks.md
@@ -289,6 +289,14 @@ pub fn toggle_light_dark<M: Component>(
 
 ## Drawbacks
 
+Conceptual concerns:
+
+1. Tightly integrating UI and game state make it harder to split out alternative declarative versions of the UI.
+This is valuable for handling alternative resolutions, input devices and languages.
+We need to carefully design for this from the start, and ensure that as much of this can be handled automatically as possible.
+
+Implementation concerns:
+
 1. Applying stored themes relies on hand-crafted `Commands` magic due to timing issues, rather than being implementable in transparent vanilla Bevy.
 2. Every style parameter component needs its own `propagate_style` and `maintain_style` systems.
 3. The type-system macro magic around `#[style_param]` is complex and mildly cursed.
--- a/rfcs/ui-building-blocks.md
+++ b/rfcs/ui-building-blocks.md
@@ -318,6 +318,32 @@ By making our UI out of vanilla ECS, we can ensure that interoperability and ext
 Keeping the UI in the ECS ensures we can take advantage of its features and improvements,
 enabling a data-driven workflow for UI designers and artists.
 
+### Why not just integrate another UI solution?
+
+Many other game engines and similar GUI tools simply integrate directly into an existing UI framework, commonly the web stack of CSS / HTML / JS.
+This has obvious advantages, namely:
+
+1. It is already built: we could hypothetically save a lot of development effort rather than reimplementing all of the various algorithms and widgets.
+2. It's already proven: such an approach is already used in other engines and AAA games.
+3. Many developers already know how to use the tools, and great learning materials exist.
+
+However, it also comes with serious disadvantages:
+
+1. When designing UIs for games, it is often *very* unclear where gameplay ends and UI begins. Using an external UI library would force a hard separation between the two, resulting in users attempting to do UI-like tasks in the ECS anyways but with minimal support or guidance.
+2. Translating data in and out of the ECS is very common. While a model-view separation may be a useful architecture for maintainability, this translation process is necessarily fairly challenging, resulting in an extended API surface area to learn and maintain and creating boilerplate whenever the boundary is crossed.
+3. Ensuring interoperability is a serious development cost, both upfront and on an ongoing basis.
+4. It greatly increases our dependencies, increasing the churn we must deal with, risking cross-platform compatibility, worsening compile times and possibly increasing binary sizes.
+5. We are beholden to the whims of the larger UI framework we are depending on, and cannot adapt it to fit our needs without forking.
+6. The most commonly known UI framework (the web stack) has serious issues with its spec due to backwards compatibility demands and messy inheritance.
+7. Adding a separate UI solution increases the burden of learning a new tool for all users who are *not* familiar with it already. These users are typically more novice than those coming from an established development career, and can be easily overwhelmed by trying to navigate a complex integration.
+
+Overall, we feel that the disadvantages of doing so outweigh the advantages, even in the short-term.
+Of course, we will be learning and borrowing from existing frameworks aggressively, on both a conceptual and implementation level.
+Like with the rest of Bevy's development, if you feel there are important conceptual lessons we've missed, please join the conversation!
+
+Finally, there's nothing inherent about `bevy_ui` due to Bevy's modular design.
+Integration with other UI frameworks (like the popular nascent [`bevy_egui`](https://crates.io/crates/bevy_egui) crate) is fully possible from an external perspective; we should ensure that this remains feasible for others who wish to use a different stack.
+
 ### Approaches to styling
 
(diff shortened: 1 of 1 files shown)


rfcs/ui-building-blocks.md
```diff
@@ -0,0 +1,413 @@
...
+}
+
+/// Automatically updates the styling for all style parameter components of type `S` whose Styles changed
+///
+/// End users should register this system in your app using `app.add_style::<S>()
+pub fn propagate_styles<S: StyleParam>(mut widget_query: Query<(&S::Base, &mut S::Final, &Styles), 
+  (With<Widget, Without<Style>, Changed<Styles>>)>,
+  style_query: Query<Option<&S::Base>, With<Style>>){
+ 
+ for (widget_param, styles) in widget_query.iter_mut(){
```
NathanSWard, line 167: `widget_query` is a `Query` of 3 things (`Base`, `Final`, and `Styles`), this iterator loop might be a bit of a typo.
It would have to be
```rust
for (sbase, sfinal, styles) in widget_query.iter_mut()
``` 
(also see below in `propagate_style_changes`)

alice-i-cecile: You're correct: this should be fixed up.

It's mostly pseudo-code, but we might as well fix it so it's clearer to readers :)

rfcs/ui-building-blocks.md
```diff
@@ -0,0 +1,413 @@
...
+```rust
+commands.spawn_bundle(ButtonBundle::default());
+```
+
+### Example: Building a Compound Widget
+
+```rust
+commands.spawn_bundle(ButtonBundle::default())
+  .with_children(|parent| {
+    parent.spawn_bundle(TextBundle::default())
```
YohDeadfall, line 58: I would like to have a sugar over it, so instead of spawning bundles I as an user will write less code by invoking a member function `with_text`/`add_text` for `TextBundle`, for example.

```rust
trait SpawnText {
    fn spawn_text(&mut self, /* ... */) -> &mut Self
}

impl<'a, b'>  SpawnText for ChildBuilder<'a, 'b> {
    // ...
}
```
```rust
commands.spawn_bundle(ButtonBundle::default())
  .with_children(|parent| {
    parent.spawn_text("Hello, World!")
  });
```

Not sure that it's possible now due to lack of optional named parameter support.

alice-i-cecile: I agree with your dislike of the parent-child API. It's not great, but I copied the existing syntax to keep the scope of this RFC sane.

I'd like to avoid custom builder functions everywhere (like for `spawn_text`); they're a lot of boilerplate to have to write and maintain, and really obscure what's happening here in a way that makes it harder for people to develop an accurate mental model.

YohDeadfall: So if I correctly understand you, the current API won't be touched by the proposal at all and what I'm asking is for another discussion, right?

alice-i-cecile: Correct. The parent-child API and the strategy for composing widgets is a separate discussion. IMO it should use relations (https://github.com/bevyengine/bevy/pull/1627), but there's a fair bit of design and then implementation work left to do there.

jihiggins:
One thing I've found is it's helpful to have Add/RemoveStyle EntityCommands, so that you don't have to pass Res<MyStyles> around to any widget building functions
(Something like `commands.entity(my_entity).add_style("my_style").remove_style("default_style")` )

YohDeadfall:
While I like the idea of `add_style`/`remove_style`, there's a thing I'm concerned about. These methods may be used for any entity even if it doesn't represent a widget.

alice-i-cecile:
> While I like the idea of add_style/remove_style, there's a thing I'm concerned about. These methods may be used for any entity even if it doesn't represent a widget.

Yes, you see this sort of problem crop up a lot. IMO, kinded entities (https://github.com/bevyengine/bevy/issues/1634) should be up-prioritized and then used aggressively if we end up using this approach to avoid confusing bugs for end users.

anchpop:
I'd like to be able to use this to style things besides UI widgets. For example, if I produce three "texture packs" for all the models in my game, it would be nice if I could use the style system to easily propagate the current texture-pack everywhere I need a texture. 

You can almost think of it as a generic way of propagating data through an application. I could imagine a slightly more general system than the one you describe also working for localization, which is a similar problem - you have one value (the current language) which you want to be able to easily change and have that change propagated to lots of components scattered around the application.

alice-i-cecile:
Thanks @anchpop; I've made the appropriate changes to ensure that everything Just Works for that, and mentioned the application of styling to other domains in Future Work :)

rfcs/ui-building-blocks.md
```diff
@@ -0,0 +1,472 @@
...
+The base version of the style parameter data will be a relation with a target of a unique dummy entity, whose identity is stored in a resource.
+Then, the final version of the data is a vanilla component (a relation with no target) of the same type,
+allowing end users to query for style parameter components directly to receive the current value after all styles have been applied.
+5. Nothing in the styling system limits its use to widgets alone. This opens up opportunities to use the same code and design patterns to game logic, meshes and so on. It hasn't been discussed above due to scope, but it's a powerful tool for end users.
+6. Finally, in order to fully move forward on the second rendition of `bevy_ui`, we also need to solve the following other UI focus areas:
+
+   1. Layout.
+   2. Wiring: call-backs, message passing (see #11).
+
+This RFC defines the underlying UI data structures, so should be settled first.
```
DavidVonDerau, line 472: I strongly disagree with this assertion. _Always write the usage code first_. [1]

When you write the usage code first, that interface will imply algorithms. Those algorithms will then imply the necessary data structures.

When you define data structures first, you end up contorting the algorithms and the user-focused API to fit those data structures. This is almost always less optimal in performance, maintainability, and quality.

The tricky part with a GUI is "what is the usage code"?

I have 2 answers for this.

1. There is the application code for querying the state of the GUI, processing GUI events, and updating or transforming the GUI at run-time.

2. There is the data definition of the GUI. If the intended way to develop a GUI is by writing application code, this is the same 1. However, if the intended way to develop a UI is through some type of non-code data format (whether that's XAML, HTML, XML, RON, etc) with a visual WYSIWYG editor, then this is _different_ from 1.

To that end, I would resolve the tracking issue for paper prototypes of the GUI [2] before this or any other GUI RFC (e.g. [3]) is written.  Each paper prototype should show a full vertical slice of GUI functionality. If you intend to have a data definition for the GUI, then each prototype should include that in the example. If the intent is that everything is defined in code, then the prototypes should show what not just snippets but full pseudocode for what layout, theming, composition, and data flow (data binding, polling, events, callbacks, etc) should look like in the context of that example.

It can be implemented in a straightforward manner. Start a Github repo and as each example is completed, put it through a PR. Reviewers throw stones at the examples _in context_. As each example is created, and new patterns are found or invalidated, the previous examples can be updated with additional PRs. At the end of the process, you'll have a Github repo containing each example with a description of it, annotated screenshots of the UI that you were trying to replicate, and pseudo-code of how you think the equivalent UI code would look for Bevy.

When you then write the RFCs, you _point at the Github repo of paper prototypes and say, "this is our proposal"_. The paper prototypes have already had significant peer review. They cover real use cases. They show full vertical slices. They can be compared to equivalent solutions written in other GUI frameworks, whether that's an entire example or just a particular feature ("here's how Elm UI would handle that").

Going through that process first has a significantly better chance of producing a usable, ergonomic API that can be implemented efficiently (with data structures chosen to fit the API afterward) _and_ it provides immediate feedback during the iterative approach of actually implementing the GUI once the RFCs (pointing to the prototypes!) are accepted: you just write code until each paper prototype isn't paper anymore!

From a scheduling perspective, you're going to go through that process anyway. The only question in my mind is how much time gets spent on these RFCs first.

Which leads to the next perspective -- architecture. These RFCs are not detailed enough for deep evaluation. If we wanted to debate the merits of this approach, the very first thing we'd need to do anyway is start mocking up larger examples and asking "ok, is this what it would look like? well what about this...". 

My last perspective is technical.

It is crucial to focus on usage code with realistic examples for 2 reasons.

1. It demonstrates that the proposed solution will, on paper, work.
2. It will focus the implementation to _exactly what is needed_.

In particular, I'm concerned by this line and other similar statements made with respect to the `relations` feature (not the subject of this RFC):

> Nothing in the styling system limits its use to widgets alone. This opens up opportunities to use the same code and design patterns to game logic, meshes and so on. It hasn't been discussed above due to scope, but it's a powerful tool for end users.

When developing any large codebase, _complexity_ is the enemy. Complexity and abstraction go hand in hand. Every single abstraction introduced in a codebase is a vector for technical debt. The best defense is to only introduce an abstraction when it has a clear need to exist and a clear scope of responsibilities. When you take a solution that was intended to solve one problem and apply it to another problem, it will often accumulate a little bit of complexity to meet the slightly different demands of the second problem. As you do this, for arbitrarily many problems, it becomes increasingly likely that the _one solution_ accumulates enough complexity that it is ultimately more error prone, less efficient, and less maintainable than it would have been to simply have _separate solutions for each problem_.

[1] https://caseymuratori.com/blog_0025
[2] https://github.com/bevyengine/bevy/issues/1974
[3] https://github.com/bevyengine/rfcs/pull/11

alice-i-cecile: Thanks for the feedback. Detail will come, as will end-to-end slices of complex examples.

Please follow https://github.com/bevyengine/bevy/issues/1974 for more updates along these lines, or feel free to start your own RFC outlining your vision for how UI should be done in Bevy.

mirkoRainer:
> I'd like to be able to use this to style things besides UI widgets. For example, if I produce three "texture packs" for all the models in my game, it would be nice if I could use the style system to easily propagate the current texture-pack everywhere I need a texture. 
> 
> 
> 
> You can almost think of it as a generic way of propagating data through an application. I could imagine a slightly more general system than the one you describe also working for localization, which is a similar problem - you have one value (the current language) which you want to be able to easily change and have that change propagated to lots of components scattered around the application. 

Wouldn't this fall outside the scope of UI?
Is there a clear definition of what Bevy UI means?

rfcs/ui-building-blocks.md
```diff
@@ -0,0 +1,472 @@
...
+That however doesn't mean it's a complete free-for-all: `bevy_ui` defines several built-in components, systems and design patterns
+to make working with UI ergonomic, maintainable and interoperable without making it any harder to extend or hook into.
+
+Commonly, UI is defined by three core components, each controlling a certain type of behavior:
+
+1. `ScreenSpace`, which dictates that an entity's transform should be drawn relative to the player's screen, rather than the world.
+Entities with the `ScreenSpace` component are drawn by the UI camera; all others are drawn with ordinary cameras.
+2. `Layout`, which tells Bevy's built-in layout systems to control its positioning relative to other `Layout` entities.
+The values of this component's fields determine exactly how this is done.
+3. `Widget`, a general purpose marker component that designates an entity as "part of the UI".
```
mirkoRainer, line 31: Would `UIComponent` be more appropriate for this naming convention? Or just `UI`? Or... `UserInterface`?
Widget seems here to make this domain specific just to Bevy. Is that desired?

An instance of when disambiguation might be needed with the term "widget":
[iOS widgets](https://support.apple.com/en-us/HT207122) and [Android widgets](https://developer.android.com/guide/topics/appwidgets/overview) already exist. Would there ever be a future for Bevy where both of these "widget" terms come into conflict? Or need disambiguation?

rfcs/ui-building-blocks.md
```diff
@@ -0,0 +1,472 @@
...
+A **widget** is a distinct part of the UI that needs its own data: be that its position, local state or any attributes.
+Widgets in Bevy are represented by entities with a `Widget` marker component.
+
+Each widget has an associated `Styles` component, which contains a `Vec<Entity>` which points to some (possibly 0) number of **styles**, represented as entities with a `Style` marker component.
+
+Each style in that vector is applied, in order, to the widget, overriding the **attributes** that already exist.
+attributes are stored as components, and serve to control some element of its presentation, such as the font, background color or text alignment.
+attributes can be reused across disparate widget types, with functionality achieved through systems that query for the relevant components.
+The type of the attribute components determines which behavior it controls,
+while the value returned by `.get()` controls the final behavior of the widget.
```
mirkoRainer, line 49: It's unclear to me what type this `.get()` method is on?

PR #2, opened 2021-04-14
Make template a github PR template

aevyrie:
Renaming to make this the github template when creating a new file


DJMcNab:
I don't think this is quite right - the way that https://github.com/rust-lang/rfcs does this is to have to the template as `000-template.md` at the directory root, and then. Also, the RFC text shouldn't be in the pull request description, it should be in the new markdown file.

@cart should we have RFC numbering too? I doubt it's going to get ridiculous, but it wouldn't hurt

aevyrie:
> I don't think this is quite right - the way that https://github.com/rust-lang/rfcs does this is to have to the template as `000-template.md` at the directory root, and then. Also, the RFC text shouldn't be in the pull request description, it should be in the new markdown file.
> 
> @cart should we have RFC numbering too? I doubt it's going to get ridiculous, but it wouldn't hurt

Yeah, I just realized this. Will close.

aevyrie:
Cart commented about numbering earlier - I think we'll plan to leave that out for now as feature/rfc names should be unique.

DJMcNab:
It's not just about uniqueness - it's useful for ordering/exploring.

Like low number RFCs are past and are expected to be somewhat outdated.

The numbering is assigned based on the GitHub pull request number, which means that

> Ensuring that we increment numbers correctly / avoid collisions

is easy because GitHub handles it for us

aevyrie:
I'll create an issue and we can discuss there?

PR #3, opened 2021-04-14
Accessibility

aevyrie:
An RFC to capture high level discussion about how we approach accessibility in bevy for both games and application.

It will be natural to spawn more focused, in-depth RFCs and PRs from this high-level one.


DJMcNab:
[RENDERED](https://github.com/aevyrie/rfcs-1/blob/patch-2/rfcs/accessibility.md)

This should be a `.md` file, to match the process and get GitHub to render it nicely.

[OLD RENDERED until ^ is addressed](https://github.com/aevyrie/rfcs-1/blob/patch-2/rfcs/accessibility)

cart:
I think this RFC is too "meta". RFCs should be for specific feature slices, not overarching topics.

cart:
I also think that we should avoid treating RFCs like "feature requests". Creating an RFC is a commitment from the author to come up with a "complete" design for a given feature. I think a first pull request should be more intentional / specific than "Bevy should have this feature". If you think Bevy needs "text to speech" support, I think the RFC should be created with a specific TTS approach in mind and a first stab at the design.

cart:
We should absolutely have high level "accessibility" conversations, but I think those should happen in the main bevy repo. Ex: Github Discussions or general brainstorming and Issues for specific feature requests.

Hopefully at some point during those discussions, Bevy contributors will start developing opinions about what these features should look like. They should then create RFCs so we can discuss the technical merits of specific designs.

aevyrie:
Thanks for the feedback cart, I definitely understand this perspective, and I'm glad you see the value in a high-level accessibility conversation.

I'm not convinced we can scale and have meaningful discussion around "pillars" like accessibility without some central source of truth like an RFC. I'm also not entirely sure Discussions are the best place to be making group decisions like this. I'm worried that discussions are still too ephemeral and difficult to track.

The impetus for this as an RFC was to give the community a way to make _decisions_ about what laws/specs we want to follow, what features we need, etc. Those decisions will drive and shape the subsequent PRs and RFCs.

cart:
The problem is that an RFC _isn't_ a source of truth until it's merged. And we can't merge something this big and nebulous without a ton of very specific conversations that should not happen in a single RFC.

I think RFCs should be for people who firmly believe in a _specific_ implementation for a _specific_ feature, have already had conversations elsewhere with community members, explored the design space, and came to a conclusion that they want to convince others of.

An RFC like this is essentially planting a flag and claiming a huge tract of land with no clear goals for it. I don't want to see a bunch of low effort "high level meta" rfcs pop up. I want people to pick a specific thing, dive deep, and propose a complete solution.

cart:
Or put another way: I would rather see 100 RFCs proposing very specific, sometimes conflicting proposals for text-to-speech, text resizing, etc than a single Accessibility meta-rfc with hundreds of conversations going on and no clear done definition.

PR #6, opened 2021-04-14
Visual Scripting

ValorZard:
This is very incomplete, but i felt that I need to have this out of my head anyways
[Rendered](https://github.com/ValorZard/rfcs/blob/visual_scripting/rfcs/visual_scripting.md)

### DIFF
--- a/rfcs/DELETEME.md
+++ b/rfcs/DELETEME.md
@@ -1,1 +1,0 @@
-Dummy file for git, please delete once the first RFC is merged.
--- a/rfcs/visual_scripting.md
+++ b/rfcs/visual_scripting.md
@@ -1,0 +1,48 @@
+# Feature Name: Visual Scripting
+
+## Summary
+
+In the forthcoming Bevy editor, there should be the ability to do visual scripting, a la Unreal's blueprints system.
+
+## Motivation
+
+Rust, while being fast and modern, is a pretty hard language for first timers to use. Additionally, when you're working on a game, there will be artists and sound designers who will need to use the engine as well. Having a visual scripting solution allows them to work on the game without having to know much programming.
+
+## Guide-level explanation
+
+Here's how I'd think it would work. (For the sake of using a term everyone probably knows, I'll call the visual scripting file a "blueprint")
+
+- Each blueprint that gets created corresponds to an entity/Component Bundle
+- The blueprint itself would display all of the components and systems. You could select systems and see what components it uses in their queries.
+- Looking at this from an artists perspective, say that they want to add a material to the entity/component bundle. They could open up a menu, search up the material component, and then click add. 
+- Then, if they want to have a system to, say, change the color of the material over time, they could right click, spawn in a new system, and then make the system do a loop where one of the variables of the colors could increase by one.
+
+## Reference-level explanation
+
+- I think the best way to handle a visual scripting system would be to just compile it to Rust. The code would be computer generated, but that shouldn't be a problem since its an non-programmer facing tool.
+
+## Drawbacks
+
+Why should we *not* do this?
+
+## Rationale and alternatives
+
+- This design isn't the best way of doing this, but its the best way I can think of right now. I look forward to talking with people about the better way of doing this.
+- Again, this feature is meant to make the engine more accessible to non-programmers and non-rust users while still being a rust centric engine.
+
+## \[Optional\] Prior art
+
+* The biggest success for a visual scripting system, and most definitely the biggest influence, would be Unreal Engine 4's Blueprint system.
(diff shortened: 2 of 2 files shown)


aevyrie:
Under drawbacks, it's definitely worth discussing managing complexity. I'm really not well versed in game dev, however managing complexity seems to be a perennial problem with no-code solutions (I'm looking at you LabView). It could be argued that the long-term cost of creating unmanageable visual scripting spaghetti outweighs the lower barrier to entry of not needing to learn Rust. Additionally, it's worth pointing out that Bevy doesn't really use Rust-specific concepts like lifetimes in the API, and it even seems like Bevy is a nice way to start learning Rust from a subset of the language.

Not trying to be a downer here, by the way, I think a visual scripting language could be great. Perhaps there is a different way to tackle the problem no-code tries to solve: lowering the barrier to entry.

E.g.: What if your code generated interactive system and component graphs? It might be a quicker way to jump into a codebase by seeing relationships and data flow visually, and you could also make quick edits with a graphical mile-high view instead of digging into the source. Instead of having a separate visual language, the goal of this concept would be to augment and visualize existing code.

ValorZard:
That would be cool, but that would probably require making a custom text editor and stuff, which would probably end up having as much dev time and resources as visual scripting. I think the biggest problem regarding rust and scripting is the inherent complexity of many of its ideas, like the borrow checker and mutability. Trying to teach that to a non-coding audience just isnt going to work, so we need to abstract all of that as much as we can.

aevyrie:
> That would be cool, but that would probably require making a custom text editor and stuff

Apologies if I wasn't clear, I'm not suggesting anything more than the UE4 blueprints-esque visual scripting editor you've outlined. I'm bringing up managing complexity because it seems to be a common issue with visual scripting, and something we should at least talk about for a bevy implementation. My example was an idea of (one) way we could constrain scripting to address this problem.

> Trying to teach that to a non-coding audience just isn't going to work, so we need to abstract all of that as much as we can.

In addition to this, cart has expressed is he is not interested in first-party support of scripting, so this visual scripting would need to compile to Rust/Bevy source directly. That seems like a pretty huge task, so starting with a visual editor for system dependencies alone might (?) be a useful place to start, and would still need to answer questions of how to serialize the output from the visual editor into the user's source code.

alice-i-cecile:
> so this visual scripting would need to compile to Rust/Bevy source directly. 

I think there may be some useful sub-problems where we compile to some form of asset / scene file instead. Dialogue, UI and state charts (on both the entity and app level) come to mind, and would present an interesting opportunity for a data-driven workflow for non-programmer team members.

ValorZard:
Mmm, the scene idea isnt actually a bad idea now that I think about it. The only problem would be, could you reuse the same scene files, or would there need to be support for another file type?

jihiggins:
Imo, anything around visual scripting would be better as a third party thing. It would be a huge undertaking in terms of design, programming, support, documentation, etc, and it would require constant maintainance. Solo developers who are looking for something with a visual editor might be better off looking at one of the many game development toolsets built for non-programmers. It's also easy to end up with a visual scripting language with performance issues or an editor that is hard to use for complex projects (more or less what @aevyrie said.) That could cause it to become a trap for new users who rely on it for everything (see: game maker drag and drop.)

Re: artist usage in larger teams, I think the editor ui / maybe some basic FSM editing or visualizing stuff would go a long way towards helping with the things an artist would want to edit themselves. Sort of along the lines of what @aevyrie said re: system / component graphs or (unless I misunderstood) what @alice-i-cecile said re: editors for UI / dialog / state charts.

In the guide level explanation, the first part of what's described is similar to having an asset editor that lets you lookup references to any systems that affect given components / assets.
This would be the part that involves scripting / code:
> Then, if they want to have a system to, say, change the color of the material over time, they could right click, spawn in a new system, and then make the system do a loop where one of the variables of the colors could increase by one.

The way the blueprint editor is designed wouldn't really line up with ECS based stuff, since each blueprint is a class and blueprint nodes are mostly either class methods, events, or static functions. You could probably copy parts of how the implementation is done, though. I think you'd want to be able to specify sets of queries, and you'd want to be able to pass the components / component variables from those queries through static / component impl methods. (I think the "blueprints" in an ecs visual scripting editor would be more analogous to systems than to particular entities / bundles / objects.)

cart:
Switching this to draft mode because currently this RFC doesn't have enough detail to be merge-able.

ValorZard:
(how do i make an rfc a draft?)

ValorZard:
I'm going to close this for now. I think the idea still has merit, but this feels too big for an RFC and honestly im not the person that can hash out this implementation. Hopefully someone else can pick up the torch

PR #13, opened 2021-04-21
Fixed Point support

ValorZard:
[RENDERED](https://github.com/ValorZard/rfcs/blob/fixed_point/rfcs/fixed_point.md)

in response to [this](https://github.com/bevyengine/bevy/issues/1680)


ValorZard:
Added future plans for better idea where this is going

bjorn3:
How would compatibility between plugins that don't opt into fixed-point bevy and games that do opt in worl? Currently floating points are directly exposed, so switching to fixed-point violates the additive nature of cargo feature flags by causing compilation errors when trying to set tranaformations to floating-poimt numbers.

ValorZard:
Honestly, it should be completely fine for plugins to not opt into fixed point bevy. As some people on the discord said, its a rare enough feature that its okay to not support every usecase for it.

Frizi:
I think that on the argument of portability, we have a strong alternative here. Namely, make sure our floating point code is portable across all supported platforms. The biggest problem with that are usually trigonometric functions and some sort of fast math optimization. I think we don't do the latter already.

This could be tested by unit tests that compare operation results with exact bit patterns, and ensure that it's the same bit pattern across all tested platforms.

maniwani:
> Namely, make sure our floating point code is portable across all supported platforms.

This is way easier said than done (although, granted, so is fixed-point support). This problem has been around for ages. Relying on strict IEEE-754-2008 compliance seems kind of fragile. If you configure everything and differences still persist, all you can do is throw up your hands. That cross-platform portability no doubt comes with a measurable performance hit. Fixed-point integers are probably much faster than floats when all their optimizations are disabled.

I'd say "yes, and" to a "compliant" floating-point mode than suggest it as a default or viable alternative to fixed-point math.

ValorZard:
I'm closing this because 
1. I don't think this is a very good RFC. The quality of it is EXTREMELY poor compared to the other rfcs currently in queue.
2. I've lost interest in getting this feature implemented. The work required for this to exist, plus the amount of people who would actually want to use it, is incredibly small. 

If someone disagrees, and does want Fixed Point support implemented, I suggest they make a much better RFC than this.

PR #14, opened 2021-04-21
Internationalization with Project Fluent

alice-i-cecile:
[RENDERED](https://github.com/alice-i-cecile/rfcs/blob/fluent-internationalization/rfcs/fluent-internationalization.md)

This RFC builds on the basic data model in #1, but is agnostic to most of the details in there, like how styling works. You probably want #11 or a competitor to actually make much use of this.

See https://projectfluent.org/ for a good overview of the proposed library integration and the problem domain.


alice-i-cecile:
Obviously, this still needs a lot of work. Input from people with experience in internationalization, localization and accessibility is very welcome!

alice-i-cecile:
Closing for now; may reopen later when I have time. Someone else is free to lead this in the meantime.

PR #16, opened 2021-04-22
Bevy Subworld RFC

NathanSWard:
[Rendered](https://github.com/NathanSWard/rfcs/blob/nward/subworld-rfc/rfcs/Subworlds.md)

This is based on the discord discussion starting [here](https://discord.com/channels/691052431525675048/749335865876021248/834310334264115210).

This is a proposal for a `Subworlds` concept: disjoint sets of entities & their components.

### DIFF
--- a/rfcs/Subworlds.md
+++ b/rfcs/Subworlds.md
@@ -1,0 +1,139 @@
+# Feature Name: `Subworlds`
+
+## Summary
+
+Currently, `bevy` contains a single _global_ `World` which holds all entities/components, which are operated on via `Queries` and `Systems`. However, there are uses for disjoint sets of entities that can operate independently from one another. This RFC proposes a concept of `Subworlds`, which are individually a distinct collection of entities and their components. In this model, `Queries` and `Systems` can operate on these `Subworlds` independently and concurrently. This provides the ability to encapsulate sets of entities into `Subworlds`, reducing some of the complexity and overhead that came with grouping entities in the _global_ `World` (via _marker-components_ for example). By excluding large sets of entities that live in other `Subworlds`, `Systems` and `Queries` can more efficiently perform their logic on a smaller, more concise set of components.  
+
+## Motivation 🏃
+
+The primary motivation for using `Subworlds` to allows systems and queries to operate on disjoint sets of entities and their components.
+
+## Guide-level explanation 📝
+
+### Terms & Definitions 📚
+
+- `Subworld`: This refers to the original definition of a `World` in bevy 0.5. (essentially a collection of components which belong to a set of entities). 
+  - _Note_: I may use `World` and `Subworld` interchangeably.
+- `Worlds`: This can be thought of as a collection of `Subworlds`. 
+
+### Example ⭐
+- Your game is composed of levels. 
+- The player is currently on level 1 and nearing the end, so you would like to begin loading level 2.
+- Level 1 and 2 are completely disjoint and do not share any data (except perhaps the player themselves).
+- The difficulty resides in the fact that entities from levels 1 and 2 contain similar components, but the entities themselves are disjoint.
+  - In order to begin populating level 2's entities, something must exist to distinguish the two sets of entities, so that systems operating for level 1 (e.g. _physics_) do not operate on level 2's entities, even though they may contain physics components.
+- Only having a single `World` requires that systems filter the entities based on what level they belong to, adding overhead and complexity to either the `Query`, the system, the components describing the entities, or a combination.
+- `Subworlds` would provide a clean and efficient solution to this problem. 
+  - The primary systems `[physics, render, etc...]` would only operate on Level 1's `Subworld`.
+  - A `load_next_level` system could be running concurrently on Level 2's `Subworld`. 
+  - Then when Level 2 begins, the primary systems operate the on later `Subworld` while the former one gets saved/dropped.
+
+### Things `Subworlds` would allow 🙂
+- Running systems on subworld `A` on thread `A`, while running the same systems on subworld `B` on thread `B`.
+- Running systems on specific disjoint groups of entities based on some criteria.
+- Running systems on subworld `A` every frame and on subworld `B` every 10 seconds.
+- Helps systems stay clean and not require overuse of _marker-components_ (and possibly _relations_) which can lead to excessive archetypal fracturing which will end up hindering performance. 
+- Give developers the choice to use either a _single_, _global_, world that contains every entity (_note_: for some games, that might makes the most sense and be the most performing), or multiple `Subworlds` if they so desire.
+
+## Reference-level explanation 🧑‍🏫
+
+TODO: expand upon this
+
+```rust
+struct Worlds {
+    subworlds: HashMap<WorldId, World>,
+}
+
+struct World { // aka `Subworld`
+    entities: Entities,
+    components: Components,
+    // more stuff
+}
+```
+
+### Example API (This is not final!) 💻
+```rust
+fn setup(mut commands: Commands) {
+    commands
+        .spawn() // this will spawn the entity into the default world
+        .create_world("World2")
+        .create_world("World3")
+        .set_world("World2") // sets the current world being operated on
+        .spawn(); // will spawn this entity into "World2"
+}
+
+fn on_level_change(mut commands: Commands) {
+    commands
+        .remove_world("World2") // removes the world and all of it's components
+        .add_world("AnotherWorld")
+        .with_system(some_system.system())
+        .unwrap();
+}
+
+fn main() {
+    App::build()
+        // This system will operate on *all* worlds by default
+        .add_system(general_system.system()) 
+        // add this system which only operates on world 2
+        .add_world_system(["World2"], world2_system.system())
+        // add this system which runs on the Default world and World2
+        .add_world_system(["Default", "World2"], two_world_system.system())
+        // provide a RunCriteria
+        .add_world_set(
+            WorldSet::new()
+                .with_run_criteria(some_criteria.system())
+                .with_system(another_system.system())
+                .with_worlds(["World2"])
+        )
+        .run();
+}
+```
+
+- By default, there is a single `World` contained within the `Worlds` (this mirrors the previous behavior of `bevy`).
+- Queries and systems still only operate entities and components from a single `World`.
+  - However, systems can be applied (concurrently) to multiple `Worlds` at a time.
+
+## Drawbacks 🙁
+
+In theory adding a dynamic number of `Subworlds` would add overhead since they must be kept in (something like) a `HashMap<Id, Subworld>`, where as having a single _global_ `World` does not incur this overhead.  
+
+## Rationale and alternatives 📜
+
+- Using _marker-components_ to distinguish entities that reside in disjoint sets. 
+  - This works, however, if `EntitiesA` {E1, E2, E3} are completely disjoint from `EntitiesB` {E4, E5, E6}, you shouldn't have to keep them grouped into a single set (`World`).
+  - Also, when constructing queries, this adds the overhead of need to check all entities instead of the entities of a specific `Subworld` that you are interested in. 
+
+- _Relations_ could also be a potential solution to this problem. 
+  - However the same problems described above apply. 
+  - The overhead of needing to check entity relationships exceeds that of running a query on a specific `Subworld`.  
+  - Overuse of _Relations_ and _marker-components_ could also lead to extreme archetypal fractaling, and worse case a single archetype-per-entities scenario. 
+
+## Prior art 🎨
+
+- See [here](https://discord.com/channels/691052431525675048/749335865876021248/834310334264115210) for the initial discord discussion that sparked this RFC.
+
+- The popular C++ ECS library [EnTT](https://github.com/skypjack/entt) supports this behavior, and even encourages it, via giving you direct access their `World` structure (`EnTT` calls these _registries_).
+```cpp
+auto level1 = entt::registry{}; // registry is analogous to bevy's World.
+auto level2 = entt::registry{};
+
+// depending on the current level, we can run the same system on a different sub-<worlds/registries>.
+some_system(level1);
+some_system(level2);
+```
+  - Currently supporting this is `bevy` is not possible, as systems and the `World` they operate on are closely coupled.
+  - `EnTT` is able to support this since systems are nothing more than free functions.
+
+## Unresolved questions ❓
+
+- How do resources behave in respect to each `Subworld`? Are they shared between them or made exclusive to each. (or both?? E.g. `Res<T>`/`ResMut<T>` vs `Local<T>`).
+- How do we support moving/copying entities and their component from one `Subworld` to another?
+- Should/Would it be possible for `Subworlds` to communicate? Or more specifically, the entities inside them to communicate?
+- What would the API look like for creating/modifying/removing `Subworld`s and how would you prescribe systems to run on specific sets of `Subworld`. E.g. a `SubworldRunCriteria`?
+- We should aim to **not** actually change the name `World` to `Subworld`. And instead introduce some new type `Worlds` (which is a collection of the sub-worlds). 
+- How does this interact with `Stages` and `RunCriteria`?
+
+## Future possibilities 🚀
+
+- Entity move
+- World-specific resources


alice-i-cecile:
After a first skim, I have two other use cases in mind for this:
1. Prefab staging grounds, as raised in https://github.com/bevyengine/bevy/issues/1446. The idea is to have prefab entities ready to go (rather than always needing to load them from file), without having them affected by random systems.
2. Parallel worlds simulation for predictive AI.

Do those sound feasible to you with this design?

rfcs/Subworlds.md
```diff
@@ -0,0 +1,139 @@
...
+  - The primary systems `[physics, render, etc...]` would only operate on Level 1's `Subworld`.
+  - A `load_next_level` system could be running concurrently on Level 2's `Subworld`. 
+  - Then when Level 2 begins, the primary systems operate the on later `Subworld` while the former one gets saved/dropped.
+
+### Things `Subworlds` would allow 🙂
+- Running systems on subworld `A` on thread `A`, while running the same systems on subworld `B` on thread `B`.
+- Running systems on specific disjoint groups of entities based on some criteria.
+- Running systems on subworld `A` every frame and on subworld `B` every 10 seconds.
+- Helps systems stay clean and not require overuse of _marker-components_ (and possibly _relations_) which can lead to excessive archetypal fracturing which will end up hindering performance. 
+- Give developers the choice to use either a _single_, _global_, world that contains every entity (_note_: for some games, that might makes the most sense and be the most performing), or multiple `Subworlds` if they so desire.
```
alice-i-cecile, line 36: This is less of a benefit and more of a literal description? I'm not sure it fits well here :)

rfcs/Subworlds.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Motivation 🏃
+
+The primary motivation for using `Subworlds` to allows systems and queries to operate on disjoint sets of entities and their components.
+
+## Guide-level explanation 📝
+
+### Terms & Definitions 📚
+
+- `Subworld`: This refers to the original definition of a `World` in bevy 0.5. (essentially a collection of components which belong to a set of entities).
```
alice-i-cecile, line 15: I'm not sure I think "Subworld" is a useful term. What is it part of, as implied by sub-?

Would "multiple Worlds" capture the same idea in a clearer way?

rfcs/Subworlds.md
```diff
@@ -0,0 +1,139 @@
...
+    // more stuff
+}
+```
+
+### Example API (This is not final!) 💻
+```rust
+fn setup(mut commands: Commands) {
+    commands
+        .spawn() // this will spawn the entity into the default world
+        .create_world("World2")
```
alice-i-cecile, line 59: We'll want to steal the Labels API from SystemLabels / StageLabel etc. for this.

rfcs/Subworlds.md
```diff
@@ -0,0 +1,139 @@
...
+        .with_system(some_system.system())
+        .unwrap();
+}
+
+fn main() {
+    App::build()
+        // This system will operate on *all* worlds by default
+        .add_system(general_system.system()) 
+        // add this system which only operates on world 2
+        .add_world_system(["World2"], world2_system.system())
```
alice-i-cecile, line 78: I think that we should try to get better composable system insertion syntax before moving forward with this proposal. Swapping to the builder pattern shouldn't be hard, but needs its own RFC.

We're already in a rough place in terms of combinatorial API explosion, and this adds another layer.

rfcs/Subworlds.md
```diff
@@ -0,0 +1,139 @@
...
+- Queries and systems still only operate entities and components from a single `World`.
+  - However, systems can be applied (concurrently) to multiple `Worlds` at a time.
+
+## Drawbacks 🙁
+
+In theory adding a dynamic number of `Subworlds` would add overhead since they must be kept in (something like) a `HashMap<Id, Subworld>`, where as having a single _global_ `World` does not incur this overhead.  
+
+## Rationale and alternatives 📜
+
+- Using _marker-components_ to distinguish entities that reside in disjoint sets.
```
alice-i-cecile, line 102: Marker components don't work well for disjoint sets where you don't know the size at compile time. This is the "dynamic component problem".

Relations can get around this to some degree but have different semantics, and "tag components" or the like are not yet implemented or designed at all.

rfcs/Subworlds.md
```diff
@@ -0,0 +1,139 @@
...
+// depending on the current level, we can run the same system on a different sub-<worlds/registries>.
+some_system(level1);
+some_system(level2);
+```
+  - Currently supporting this is `bevy` is not possible, as systems and the `World` they operate on are closely coupled.
+  - `EnTT` is able to support this since systems are nothing more than free functions.
+
+## Unresolved questions ❓
+
+- How do resources behave in respect to each `Subworld`? Are they shared between them or made exclusive to each. (or both?? E.g. `Res<T>`/`ResMut<T>` vs `Local<T>`).
```
alice-i-cecile, line 129: Given that Resources are now part of `World`, the default behavior would likely be to be isolated to each Subworld. I'm not sure that's the *desired* behavior though.

rfcs/Subworlds.md
```diff
@@ -0,0 +1,139 @@
...
+- How do we support moving/copying entities and their component from one `Subworld` to another?
+- Should/Would it be possible for `Subworlds` to communicate? Or more specifically, the entities inside them to communicate?
+- What would the API look like for creating/modifying/removing `Subworld`s and how would you prescribe systems to run on specific sets of `Subworld`. E.g. a `SubworldRunCriteria`?
+- We should aim to **not** actually change the name `World` to `Subworld`. And instead introduce some new type `Worlds` (which is a collection of the sub-worlds). 
+- How does this interact with `Stages` and `RunCriteria`?
```
alice-i-cecile, line 134: Being able to desynchronize schedules across worlds would be very interesting, but may result in serious scope creep.

rfcs/Subworlds.md
```diff
@@ -0,0 +1,139 @@
...
+some_system(level2);
+```
+  - Currently supporting this is `bevy` is not possible, as systems and the `World` they operate on are closely coupled.
+  - `EnTT` is able to support this since systems are nothing more than free functions.
+
+## Unresolved questions ❓
+
+- How do resources behave in respect to each `Subworld`? Are they shared between them or made exclusive to each. (or both?? E.g. `Res<T>`/`ResMut<T>` vs `Local<T>`).
+- How do we support moving/copying entities and their component from one `Subworld` to another?
+- Should/Would it be possible for `Subworlds` to communicate? Or more specifically, the entities inside them to communicate?
```
alice-i-cecile, line 131: Perhaps via shared resources? This is particularly compelling since Events are stored as resources.

rfcs/Subworlds.md
```diff
@@ -0,0 +1,139 @@
...
+}
+
+fn main() {
+    App::build()
+        // This system will operate on *all* worlds by default
+        .add_system(general_system.system()) 
+        // add this system which only operates on world 2
+        .add_world_system(["World2"], world2_system.system())
+        // add this system which runs on the Default world and World2
+        .add_world_system(["Default", "World2"], two_world_system.system())
```
alice-i-cecile, line 80: I would love to see a "for each world, run this system" API as well. I expect this will be the common case.

Perhaps with added filtering?

rfcs/Subworlds.md
NathanSWard: Yea, this is what I assumed we should use :)

rfcs/Subworlds.md
NathanSWard: Yea that's a fair point. Would you have a link/reference to an issue/PR associated with this task?
I could potentially help move it forward as well 🥳

rfcs/Subworlds.md
NathanSWard: Yes I agree that it's not the _desired_ behavior. I'll update this to better reflect the desire/intention of Resources.

rfcs/Subworlds.md
alice-i-cecile: Apparently I'd only discussed it informally! Here's a fresh one for you: https://github.com/bevyengine/bevy/issues/1978

Should be a fairly straightforward change.

rfcs/Subworlds.md
alice-i-cecile: Excellent. As you polish this up you should call this out explicitly to make the design as clear as possible.

hymm:
How will subworlds interact with the scheduler? Specifically stages, states, and exclusive systems.

rfcs/Subworlds.md
hymm: I would think the asset server at the very least needs to be shared between worlds.  Though storing handles to resources should be per world, so that when the world is dropped the handle can be freed.

jamescarterbell:
I'm wondering if this needs to be an engine level thing or if it can just be its own library with the subworlds as resources. Having it as a library pretty much clears up all the questions about how the resources are shared and also allows for the consumers to set it up as needed for their use case.

alice-i-cecile:
I would be *very* interested to see if we could use this idea to temporarily store removed entities / components / relations in a purgatory world. I think that would clean up both the API and the impl in a satisfying way.

alice-i-cecile:
`@sircarter had an excellent [comment](https://discord.com/channels/691052431525675048/749335865876021248/835335381427683349) that I don't want to get lost:

> If subworlds are going to be a engine feature it'd be nice if we could have queries look like this
`Query<Q, F, W>`
where `W: DerefMut<target = World>`
>
>Or something so you could easily query subworlds. This could also be added to relations so you could have cross world relations

cart:
This has some overlap with my pipelined rendering experiments:

* I have a Game World and a Render World
* I have a Game Schedule and a Render Schedule
* These need to be able to run independently in parallel (at times) and be synchronized (at times)

My current experimental solution is `SubApps`. Note that this is a hack-ey impl that only exists to get the ball rolling. It largely punts the synchronization (and parallel execution) problem, but it requires almost no changes to the current apis or app models, which is nice.

```rust
pub struct App {
    pub world: World,
    pub runner: Box<dyn Fn(App)>,
    pub schedule: Schedule,
    sub_apps: Vec<SubApp>,
}

struct SubApp {
    app: App,
    runner: Box<dyn Fn(&mut World, &mut App)>,
}

impl Plugin for PipelinedRenderPlugin {
    fn build(&self, app: &mut App) {
        /* more render stuff here */
       
        let mut render_app = App::empty();
        let mut extract_stage = SystemStage::parallel();
        // don't apply buffers when the stage finishes running
        // extract stage runs on the app world, but the buffers are applied to the render world
        extract_stage.set_apply_buffers(false);
        render_app
            .add_stage(RenderStage::Extract, extract_stage)
            .add_stage(RenderStage::Prepare, SystemStage::parallel())
            .add_stage(RenderStage::Render, SystemStage::parallel());
        
        app.add_sub_app(render_app, |app_world, render_app| {
            // extract
            extract(app_world, render_app);

            // prepare
            let prepare = render_app
                .schedule
                .get_stage_mut::<SystemStage>(&RenderStage::Prepare)
                .unwrap();
            prepare.run(&mut render_app.world);

            // render
            let render = render_app
                .schedule
                .get_stage_mut::<SystemStage>(&RenderStage::Render)
                .unwrap();
            render.run(&mut render_app.world);
        });
}

impl Plugin for PipelinedSpritePlugin {
    fn build(&self, app: &mut App) {
        // 0 is currently implicitly the "render app". A more complete solution would use typed labels for each app.
        app.sub_app_mut(0)
            .add_system_to_stage(RenderStage::Extract, extract_sprites.system())
            .add_system_to_stage(RenderStage::Prepare, prepare_sprites.system()));
        
        /* more render stuff here */
    }
}
```

cart:
I'm definitely not advocating for SubApps as the solution at this point. Just sharing where my head is at / what my requirements currently are.

alice-i-cecile:
From [Discord](https://discord.com/channels/691052431525675048/743664267856838736/837842890080124938): we could use this as a tool to get closer to ["Instant" command processing](https://github.com/bevyengine/bevy/issues/1613).

Have three default worlds (or more likely two shadow worlds for "normal" world):
1. The main world, which works as we have now.
2. The Commands world, where components are immediately added / entities are instantly spawned before being reconciled.
3. The purgatory world, where removed components and entities live for two frames under a double buffering scheme before being permanently deleted.

This would be fantastic for prototyping and ergonomics, and has a nice clear mental model.

rfcs/Subworlds.md
```diff
@@ -0,0 +1,139 @@
...
+- Helps systems stay clean and not require overuse of _marker-components_ (and possibly _relations_) which can lead to excessive archetypal fracturing which will end up hindering performance. 
+- Give developers the choice to use either a _single_, _global_, world that contains every entity (_note_: for some games, that might makes the most sense and be the most performing), or multiple `Subworlds` if they so desire.
+
+## Reference-level explanation 🧑‍🏫
+
+TODO: expand upon this
+
+```rust
+struct Worlds {
+    subworlds: HashMap<WorldId, World>,
```
cart, line 44: If we use types for World identifiers (ex: `MainWorld`, `RenderWorld`, etc) and include those in Query types (which has been discussed elsewhere in this RFC / on discord), then we could remove the current runtime Query WorldId checks (that ensure the query matches the created world) in favor of compile time type safety.

cart: Eh actually maybe not, unless we make it impossible to create two worlds with the same type identifier.

alice-i-cecile: That seems like a perfectly sensible restriction to have.

tower120:
I would like to say that having separate worlds alone may be not enough.

Imagine that you have some simulation - not only you want to isolate it entities and systems/queries, it is also desirable to tick/update it at specific rate. You may want to have that simulation real-time synced, or may want to jump forward in time, by simulating hundreds of world frames at once (probably while not locking main world). In other words, tick world manually may be necessary, fixed step is not always an option.

So I would vote for something close to separate apps... That would also make development of certain game aspects (like mini-games, smart 3D HUD/interface, etc...) more isolated/modular. You can make your mini game as separate project, then just attach it to main one. (This could be achieved with plugins as well - but you can touch something in main world you want/should not)

jamescarterbell:
@tower120 Subworlds may not be enough for ever use case, but they are still a useful additional and one that can be added independent of multiple executors/synced apps. Although for your examples I can imagine some good solutions that don't require multiple apps.

cart:
Agreed. I think we'll want "multiple ECS schedules" and "multiple worlds" with the ability to pass an arbitrary number of worlds into a given schedule. I think "multiple apps" currently implies a tight [Schedule, World] coupling, which I think isn't what we want. Instead, I think we want to allow "arbitrary orchestrator logic" to determine how Schedules execute on Worlds. That way you don't need to create a bunch of empty Schedules when all you need is "staging worlds" for your app logic.

tower120:
@cart Ok, that sounds even nicer. But what about Time resource? Doesn't schedulers coupled with time?

And I think I want to have some kind of "manual" scheduler, where you can do `scheduler.update(delta_time)`. Like for case of some sort of turn-based simulation, or offline-simulation.

alice-i-cecile:
> And I think I want to have some kind of "manual" scheduler, where you can do scheduler.update(delta_time). Like for case of some sort of turn-based simulation, or offline-simulation.

For this, the current [custom runner](https://github.com/bevyengine/bevy/blob/main/examples/app/custom_loop.rs) abstraction works pretty well; I think that we could integrate that nicely into what Cart's proposing to get a really lovely abstraction.

cart:
Lots of things rely on specific System/World pairs. As long as the Time used by app logic is stored in the "app world" and it is ticked using the "app schedule", things will work as expected.

tower120:
@alice-i-cecile Looks like that "runner" blocking.... It would be much nicer to `update` app/scheduler directly. Like to "move" world when we're want that.

BTW, why you want "subworlds" (like hierarchical?), not just "worlds"(plain) ?

alice-i-cecile:
> BTW, why do you want "subworlds" (like hierarchical?), not just "worlds"(plain) ?

The initial implementation was easier if you stored them in Resources :P From my perspective, I'd enjoy a bit of hierarchy to have nice "purgatory" or "command staging" worlds associated with each "proper" world.

cart:
We can always have freestanding Worlds that live inside of components / resources. But imo if we're integrating with the scheduler it should only care about "global top level worlds"

tower120:
BTW, Unity ECS has interesting feature of world merging. In Unity ECS archetype's storage consist of 16Kb chunk linked list. So merge is close to just changing pointers in list.
That very useful for async loading / world cells streaming. You load everything what you need in stage/temporary world, then when load is done - you just merge that world into active/main one..

Maybe bevy could utilize something like that with new multi-world system...

cart:
Ooh thats very clever. We'd need to do that at the table level, not the archetype level (because archetypes are just metadata in Bevy), but its definitely possible.

I'm starting to think we need both:

## Sub Worlds

Shared entity id space, shared (or easily mergable) metadata ids like ComponentId/ArchetypeId/BundleId, etc. These would be _extremely_ useful for a number of scenarios:
  * Replacing Commands: Commands are allocation heavy right now and can bottleneck op-heavy things. SubWorlds would allow us to move more work into systems, re-use Component allocations across runs (and/or directly move the allocated data into the main world), handle "archetype moves" within a system, run queries on entities that were just spawned, etc
  * "Baked Scenes": Prepare entities in a "subworld" and "merge" it when its ready. However this is slightly different than Commands in that Scenes might be instantiated multiple times. Ids shouldn't be reused, Component values might need to be remapped (ex: for unique Asset or Physics handles), etc.
 
The major missing pieces here:
* Apis for creating and interacting with subworlds (should largely mirror the World api)
* Ability to share / sync archetype, component, and bundle information with source World. Direct sharing would likely involve locks. Indirect sharing would be memory-hungry. This will require some careful thought. It might require "reconciliation" logic for Archetypes created in SubWorlds that aren't yet in the main World, or that _do_ exist in the main World, but were created elsewhere.
* Entity Id space management: ideally we don't need to "remap" entities created in subworlds (because users will want to store+use EntityIds created for them and those entities should remain valid when merged with the main world). _allocating_ entities in parallel is already possible. The missing piece is handling metadata tracking (which is currently stored in a dense array). How should subworlds store this metadata, which won't be densely packed at zero? How will the main world account for entity ids allocated, but not yet merged into the main world? I'm guessing it will likely be something like "Make EntityMeta an `enum { Live(Meta), Allocated }` .... when `entities.flush()` is called, set subworld ids to Allocated". This will add an additional branch to entity location lookup, but its probably fine, given the benefits we're getting.
* Ability to efficiently merge storages. Maybe it uses pages like @tower120 mentioned to allow complete reuse of allocations. Maybe it just does the most efficient copy possible.
* Ensure SubWorlds _cannot_ access data they shouldn't be able to in the context of a schedule (ideally they just don't have any references to this data). Id also like to avoid doing expensive "access control" on subworlds (ex: what legion does for a similar api). SubWorlds shouldn't have access to the main World data / they shouldn't be a "filter" on the main world data. 

## Multiple Worlds

Completely separate Worlds without any shared state. Useful for running parallel tasks/contexts (ex: AppWorld, RenderWorld, NetworkWorld, OtherAppWorld, EditorWorld, etc).

The major missing pieces here:
* High level api for constructing these worlds
* Scheduler integration to support running logic on multiple worlds at the same time
* System integration to enable accessing Components and Resources from multiple worlds at the same time.
* User-configurable orchestration logic to determine when "multi-world schedules" run.

## Summary

I think "sub worlds" could yield some pretty significant performance and functionality wins that would impact all Bevy apps, but they are a more complicated endeavor.

"Multiple worlds" are only useful in a few specific contexts, but the scope is much smaller.

I don't think theres a good way to unify these two scenarios (and I don't think theres much value in doing so). We should probably start two separate efforts to design them.

alice-i-cecile:
I completely agree with your analysis above @cart. Both are very useful, but have entirely distinct use cases and design challenges. Getting the high-level API right is about the only bit that I think should be done in unison.

NathanSWard:
> I'm starting to think we need both:

I also completely agree with this.
And if I'm being honest, when I initially created the RFC, I was thinking more `Multiple Worlds` than what has turned into `Sub Worlds`
I guess it was just a bad naming choice on my half.....

tower120:
@cart What about "Universe" conception? Universe is basically EntitySpace (virtually, EntityId generator). Each world constructed_within/constructed_with universe and cannot be moved between universes.
According to what you described: subworlds - is worlds within the same "universe", multiworlds - worlds in different "universes".
Worlds within the same Universe can be merged and traversed by the same system(/scheduler?).

P.S. What kind of ECS storage bevy use? (I thought it is archetype-centic...)

bjorn3:
Bevy supports both the table layout, which is often called archetype in other ECS'es and the sparse layout. The default is table, but you can override it to sparse on a per-component basis.

Waridley:
Just being able to run the same `SystemStage` on multiple worlds would be extremely helpful for my use case: killcams with rollback netcode. Everything else in this discussion would just be bonus convenience features for me, but not being able to run the same `SystemStage` on two separate worlds makes killcams so awkward to implement that I probably won't bother until this feature is added to Bevy. A separate world is how Overwatch implements its killcams as well. It lets the game keep ticking in the background while the replay is being rendered.

alice-i-cecile:
My current gut feeling is that the basic design should be:
- each app stores a `Vec<World>`
- each app also stores a `Vec<(Schedule, WorldId)>`
- by default, there is only one of each
- schedules run on the world they are pointing to each tick (if any), and all schedules must complete before the next tick can begin (allowing for cross-world synchronization at the end of the tick)
- we add a `Commands`-like tool to transfer ECS data and ordinary `Commands` between worlds, create and delete worlds, and enable / disable schedules
- `Schedules` become clone-able

TheBlckbird:
Any updates on this? Is it still planned to implement this?

alice-i-cecile:
A solution for something in this space is still desired, but the ECS crew largely have their hands full pushing towards relations at the current time.

jpedrick:
I've briefly read through this and I'm not sure this solution has been proposed:

```rust

#[derive(Component)]
struct SomeComponent;

#[derive(World, Invisible)]
struct AlternateInvisibleWorld;

#[derive(World, Visible)]
struct AlternateVisibleWorld;

fn some_system(commands: &mut Commands){
    // Default is MainWorld
    for _ in (0..3) {
        commands.spawn(SomeComponent).with_children(|parent| parent.spawn(...));
    }
    commands.spawn(SomeComponent).in_world(AlternateVisibleWorld).with_children(|parent| parent.spawn(...));
    for _ in (0..2) {
        commands.spawn(SomeComponent).in_world(AlternateInvisibleWorld).with_children(|parent| parent.spawn(...));
    }
}

fn query_all_worlds(
    some_main_world_components: Query<SomeComponent, In<MainWorld> >,
    some_visible_components: Query<SomeComponent, In<AlternateVisibleWorld> >,
    some_invisible_components: Query<SomeComponent, In<AlternateInvisibleWorld> >,
){
    assert!(some_components.iter().count() == 3);
    assert!(some_visible_components().count() == 1);
    assert!(some_invisible_components().count() == 2);
}

// Queries by default use MainWorld
fn query_only_main_world(
    some_main_world_components: Query<SomeComponent>,
){
    assert!(some_components.iter().count() == 3);
}

fn also_query_all_worlds(
    some_main_world_components: Query<SomeComponent, In<AllWorlds> >,
){
    assert!(some_components.iter().count() == 6);
}

fn query_union_alternate_worlds(
    some_main_world_components: Query<SomeComponent, (In< AlternateVisibleWorld >, In<AlternateInvisibleWorld>) >,
){
    assert!(some_components.iter().count() == 3);
}

```

So, ultimately, the design would utilize `World`, but would have components specifying what world they exist in. It could even allow some entities to exist in multiple worlds(What for? I do not know.).

couyit:
I am developing a node-based modeling tool, which I believe could be one possible use case for Subworld.
Vertices are stored in Subworld, and the node-based procedure is an exclusive system in Subworld that runs in parallel with Mainworld systems.
Thanks to Subworld, I can keep them isolated while running the procedure in parallel.

I'm not very familiar with ECS, so I might make mistakes; however, I have devised a solution to address the challenges in implementing Subworld.

As far as I understand, queries in ECS proceed like this:

- Identify archetypes that match the query.
- Retrieve tables corresponding to the archetypes.

As I'm not confident in my English, I'm including a figure to better clarify my explanation.

IMG_0377

The solution transforms the above structure into the following:

IMG_0381

In this case, queries proceed as follows:

- Determine which archetype list to use.
- Identify archetypes in `archetype2` that match the query.
- Retrieve tables corresponding to the archetypes.

The merged archetypes could be structured like this:

IMG_0379

Alternatively, it could be structured like this:

IMG_0380

In the initial structure, each archetype is associated with a list of components. A similar approach is likely used in Unity ECS.
In the second structure, the same archetype is duplicated and stored in the archetypes list. I believe this structure aligns better with Bevy's architecture, but I am not sure if this structure is achievable.
No matter the structure, Entity Id space and ComponentID are always shared between Subworld and Mainworld.
If my understanding of ECS is correct, this approach should work, but I am not entirely confident.



Edit: I am also considering scheduling in Subworld.
Using the same scheduler could be a viable approach.
As far as I know, Bevy executes systems in parallel when there are no conflicting data accesses.
Ideally, if two systems access separate Subworlds, the scheduler would automatically execute them in parallel.
Users would still be able to add systems as before, but queries may need to be adjusted to support a format similar to `Query<Q, F, W>`, where `W: DerefMut<target = World>` as mentioned in the previous comments.

Merging Subworlds is also a system, but since it accesses both worlds, the scheduler determines that other systems accessing the worlds targeted for merging cannot run in parallel due to data access conflicts. This is similar to how the current scheduler prevents systems with data access conflicts from running in parallel.

As I lack in-depth knowledge of the scheduler, I am uncertain whether this method can be implemented.

TBye101:
Hi all, I'm developing a game that is similar in many ways to Rimworld but with more of a fantasy aim.

I'd like to add my scenario that I believe this thread's concept could solve: 

Allowing for multiple "dimensions", some of which operate faster or slower than others (tick rate) and have different rules (systems) as to how they function. 

In my simulation, it would be very nice to be able to render only one `World` which the camera is seeing, but run the systems of and simulate multiple `World`s as necessary. Secondary worlds might have different sets of systems, and a desire to receive more or less updates than the main world. The camera can move between `World`s, influencing what the player sees on screen.

I'm a ways away from actually wanting such a thing, but hopefully by then Bevy will have a baked in way to handle such a thing!

PR #17, opened 2021-04-23
Automatic Event Tracking

jihiggins:
### [Implementation PR](https://github.com/bevyengine/bevy/pull/1776)

### DIFF
--- a/rfcs/Automatic event tracking.md
+++ b/rfcs/Automatic event tracking.md
@@ -12,9 +12,16 @@ The current double buffering approach is error prone when used in systems with r
 
 https://github.com/bevyengine/bevy/pull/1776
 
-## Guide-level explanation
+## API Changes
 
-*TBD, may not require api changes*
+No longer need to worry about _when_ an event is read. As long as the system / EventReader parameter was setup before the event was created, the event is guaranteed to live until that system / EventReader has looked at it.
+
+The current draft implementation has manual readers accessed like this:
+```rust
+let app_exit_event_reader = app_exit_events.get_reader("app_runner");
+```
+The string input functions as a tag so that manual_reader access can be tracked.
+Other than this, it is basically the same.
 
 ## Reference-level explanation
 
@@ -29,6 +36,8 @@ Events will need a way to tell which part of the buffer has been read by all rel
 ## Rationale and alternatives
 
 This would involve minimal / no api changes, and would allow us to avoid any lost events. If the user forgets to run Events::update, the current method of making sure events are not lost can lead to unbounded memory usage.
+
+It might be better to just leave manual event readers unchanged and have their reads not be tracked. If this is the case, the manager should avoid removing events until the end of the frame (or until the end of the next frame), even if they have been read by all subscribers.
 
 ## Unresolved questions
 
--- a/rfcs/Automatic event tracking.md
+++ b/rfcs/Automatic event tracking.md
@@ -1,0 +1,43 @@
+# Feature Name: `automatic_event_tracking`
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
I'm still working through [Differential Dataflows](https://timelydataflow.github.io/differential-dataflow/) but I am increasingly concerned that @coleporier is right that we may need something to this effect to handle all of the various event patterns in a principled way :p I will chew on this more later...

tower120:
I made some progress with my EventQueue.

According to benchmarks, read performance is stunning https://github.com/tower120/lockless_event_queue/blob/91383baff8aeda56bdd80ab60984daae6f3fd992/benches/read_bench.rs#L8 :
```
EventQueue worst case scenario: 100ms
EventQueue best case scenario: 40ms
Deque: 60ms
Vec: 28ms
```
Benchmark is single-threaded, to show overhead. Since there is no locks on read operations, in MT case - it is as fast as vec/deque read in MT (namely, memory bandwidth bounded).
I didn't benchmark write overhead yet.

I think, I found how to deal with clean operation without additional locks.
Event contains `start_point = *chunk + index`  and `start_point_epoch`. On `clear` call `start_point_epoch` increased and `start_point` updated with queue head position. 
Each Reader contains `start_point_epoch`. On start of read session (getting iterator), reader's compared with atomically read event's `start_point_epoch`. If they mismatch -  and event's `start_point` > reader's current position - reader position updated, and all chunks in between mark as read.

Thus, reader (within read session quants) always work with actual queue.
The only problem that remains - is sleepy readers. To combat them, it is enough to just update their positions, like in read session start. In Bevy, it should be possible to access system's local resources (event readers) once in a while, without actually running systems. Just calling `update` on readers after `clear` will force queue to drop its tail. 

So, I see it as following:
_Events run in `AUTO_CLEANUP = false` mode, which means that it does not clean not needed chunks automatically. We'll call shrink_to_fit manually._
* Arc<Event<T>> constructed and shared as local res among all systems that push events. Each pushing system also has LocalRes Vec<T>. When you push event - you actually push to that vec first. After system run - that vec mass_pushed to Event with just one lock.
* Each pulling system contains EventReader, constructed from Event. Before system run iterator acquired and send to systems. Then iterator destroyed.
* [Each N frames] After all systems run:
   - All events, with too big length cut.
   - On all readers `update` is called.
* [Each frame] After all systems run:
   - On each event `shrink_to_fit` called.

So.... What do you say? For me, looks like it deal with both performance and memory issues, while being able to pass messages reliably between frames.

colepoirier:
@tower120 Are you sure your units are correct? Numbers in the 10s of milliseconds seems outrageously high for the time per operation. Unless this is total time taken? If so I’m sorry I’m not sure I understand the benchmark. Can you please explain it further?

tower120:
@colepoirier that's time of reading 100`000 items. Look at Vec as baseline. Nothing can't be faster than it.

colepoirier:
@tower120 Ah thanks that makes more sense. I would suggest you add information like this and some further context because as it stands right now you’re explanation is lacking in what makes this a good solution and relies on the reader to properly understand your benchmark and implementation.

alice-i-cecile:
@tower120 Impressive! I'd be interested in seeing an RFC (or PR with nice clear explanatory text) for this. This type of tool may be very useful to have in place as we tackle UI next, so I'd love a proper write-up.

tower120:
I would like to make this `EventQueue` as separate crate, cuz I think it have common functionality.
I'll describe principle of operation in a documentation.
I'll make separate RFC describing `EventQueue` integration into Bevy.

alice-i-cecile:
> I would like to make this EventQueue as separate crate, cuz I think it have common functionality.
> I'll describe principle of operation in a documentation.
> I'll make separate RFC describing EventQueue integration into Bevy.

Awesome, sounds like a great choice. It's always nice to help grow the whole Rust ecosystem.

PR #18, opened 2021-04-23
Entity Relations

BoxyUwU:
RFC mostly written by alice with some help from me

[Rendered](https://github.com/BoxyUwU/rfcs/blob/min-relations/rfcs/min-relations.md)

[Partial Impl](https://github.com/bevyengine/bevy/pull/1627)

### DIFF
--- a/rfcs/min-relations.md
+++ b/rfcs/min-relations.md
@@ -153,6 +153,8 @@ fn new_friends(query: Query<&mut Excitement, Added<Relation<FriendsWith>>>){
 
 ```
 
+TODO: mention relation cleanup on despawn.
+
 Sometimes, we may instead want to look for "entities that are targets of some relation".
 We can do so by iterating over all entities with that relation, and then filtering by target.
 
--- a/rfcs/min-relations.md
+++ b/rfcs/min-relations.md
@@ -244,15 +244,23 @@ macro_rules Relation {
 
 ## Future possibilities
 
-1. Data on the target entity.
-2. GRAPH
-3. Kinded entities
-4. Use with UI.
-5. Frustum culling.
-6. Replace parent-child.
-7. Automatically symmetric relations.
-8. `Noitaler`
-9. Non-self-referential relations using archetype invariants. `Query<&mut Money, Relation<Owes, &mut Money>>`
+Relation enhancements:
+
+1. Sugar for accessing data on the target entity in a single query. `Relation<T: Component, Q: WorldQuery>`
+2. Graph traversals.
+3. Arbitrary target types instead of just `Entity` combined with `KindedEntity` proposal to ensure your targets are valid.
+4. Automatically symmetric relations.
+5. `Noitaler`*
+6. Graph shape guarantees using archetype invariants.
+7. Streaming iters to allow for queries like: `Query<&mut Money, Relation<Owes, &mut Money>>` and potentially use invariants on graph shape to allow for skipping soundness checks for the aliasing `&mut Money`'s
+8. Assorted Performance optimizations.
+9. Relation orders to enable things like `Styles`.
+
+Relation applications in the engine:
+
+1. Replace existing parent-child.
+2. Applications for UI: styling, widget wiring etc.
+3. Frustum culling.
 
 AUTHOR'S NOTE: Much like `yeet`, `Noitaler` is the bikeshed-avoidance name for the inverse of `Relation`, and will never be used.
 EDITOR's NOTE: `Noitaler` may very much be used :3
--- a/rfcs/min-relations.md
+++ b/rfcs/min-relations.md
@@ -80,21 +80,21 @@ Just like with ordinary components, you can request them in your queries:
 ```rust
 // Relations can be queried for immutably
 fn friendship_is_magic(mut query: Query<(&mut Magic, &Relation<FriendsWith>), With<Pony>>) {
-	// Each entity can have their own friends
+    // Each entity can have their own friends
     for (mut magic, friends) in query.iter_mut(){
-		// Relations return an iterator when unpacked
-		magic.power += friends.count();
-	}
+        // Relations return an iterator when unpacked
+        magic.power += friends.count();
+    }
 }
 
 // Or you can request mutable access to modify the relation's data
 fn interest(query: Query<&mut Relation<Owes>>){
-	for debts in query.iter(){
-		for (owed_to, amount_owed) in debts {
-			println!("we owe {} some money", owed_to);
+    for debts in query.iter(){
+        for (owed_to, amount_owed) in debts {
+            println!("we owe {} some money", owed_to);
             amount_owed *= 1.05; 
-		}
-	}
+        }
+    }
 }
 
 // You can query for entities that target a specific entity
@@ -102,36 +102,38 @@ fn friend_of_dorothy(
     mut query: Query<Entity, With<Relation<FriendsWith>>>, 
     dorothy: Res<Dorothy>,
 ) {
-	// Setting relation filters requires that the query is mutable
-	query.set_relation_filters(
+    // Setting relation filters requires that the query is mutable
+    query.set_relation_filters(
         RelationFilters::new()
             .add_target_filter::<FriendsWith, _>(dorothy.entity)
     );
 
-	for source_entity in query.iter(){
-		println!("{} is friends with Dorothy!", source_entity);
-	}
+    for source_entity in query.iter(){
+        println!("{} is friends with Dorothy!", source_entity);
+    }
 
     // RelationFilters are reset at the end of the system, make sure to set them every time your system runs!
-	// You can also reset this manually by applying a new `RelationFilters` to the query
-	// They override existing relation filters, rather than adding to them
+    // You can also reset this manually by applying a new `RelationFilters` to the query
+    // They override existing relation filters, rather than adding to them
 }
 
 // Or even look for some combination of targets!
 fn caught_in_the_middle(mut query: Query<&mut Stress, With<Relation<FriendsWith>>>,
-	dorothy: Res<Dorothy>,
-	elphaba: Res<Elphaba>){
-	
-	// You can directly chain this as set_relation_filters returns &mut Query
-	query.set_relation_filters(
-		// Note that we can set relation filters even if the relation is in a query filter
+    dorothy: Res<Dorothy>,
+    elphaba: Res<Elphaba>){
+    
+    // You can directly chain this as set_relation_filters returns &mut Query
+    query.set_relation_filters(
+        // Note that we can set relation filters even if the relation is in a query filter
         RelationFilters::new()
+            // Note: add_target_filter is additive- i.e. entities this query
+            // matches must be friends with *both* dorothy AND elphaba
             .add_target_filter::<FriendsWith, _>(dorothy.entity)
-			.add_target_filter::<FriendsWith, _>(elphaba.entity)
+            .add_target_filter::<FriendsWith, _>(elphaba.entity)
     ).for_each(|mut stress| {
-		println!("{} is friends with Dorothy!", source_entity);
-		stress.val = 1000; 
-	})
+        println!("{} is friends with Dorothy and Elphaba!", source_entity);
+        stress.val = 1000; 
+    })
 }
 
 /// If we want to combine targets in an "or" fashion, we need a slightly different syntax
@@ -141,16 +143,16 @@ TODO: write example
 fn not_alone(
     commands: mut Commands, 
     query: Query<Entity, Without<Relation<FriendsWith>>>,
-	puppy: Res<Puppy>
+    puppy: Res<Puppy>
 ) {
-	for lonely_entity in query.iter(){
-		commands.insert_relation(FriendsWith, puppy.entity);
-	}
+    for lonely_entity in query.iter(){
+        commands.insert_relation(FriendsWith, puppy.entity);
+    }
 }
 
 // So does change detection with Added and Changed!
 fn new_friends(query: Query<&mut Excitement, Added<Relation<FriendsWith>>>){
-	query.for_each(|mut excitement|{excitement.value += 10});
+    query.for_each(|mut excitement|{excitement.value += 10});
 }
 
 ```
@@ -165,14 +167,14 @@ fn all_children(
     player_query: Query<Entity, With<Player>>,
     child_query: Query<&Name, (With<Relation<ChildOf>, Without<Player>)>>,
 ) {
-	let player_entity = player_query.single.unwrap();
-	
+    let player_entity = player_query.single.unwrap();
+    
     child_query.set_relation_filters(
         RelationFilters::new()
             .add_target_filter::<ChildOf, _>(player_entity)
     );
 
-	for name in child_query.iter() {
+    for name in child_query.iter() {
         println!("{} is one of my children.", name) 			
     }
 }
@@ -186,17 +188,17 @@ fn debts_come_due(
     mut debtor_query: Query<(&mut Money, Relation<Owes>), Without<Shark>>, 
     mut lender_query: Query<&mut Money, With<Shark>>,
 ) {
-	for (mut debtor_money, debt) in debtor_query.iter(){
-		for (lender, amount_owed) in debt {
-			let mut lender_money = lender_query.get(lender).unwrap();
-			if debtor_money.dollars >= amount_owed.dollars {
-				debtor_money.dollars -= amount_owed.dollars;
-				lender_money.dollar += amount_owed.dollar;
-			} else {
-				panic!("Nice kneecaps you got there...");
-			}
-		}
-	}
+    for (mut debtor_money, debt) in debtor_query.iter(){
+        for (lender, amount_owed) in debt {
+            let mut lender_money = lender_query.get(lender).unwrap();
+            if debtor_money.dollars >= amount_owed.dollars {
+                debtor_money.dollars -= amount_owed.dollars;
+                lender_money.dollar += amount_owed.dollar;
+            } else {
+                panic!("Nice kneecaps you got there...");
+            }
+        }
+    }
 }
 ```
 
--- a/rfcs/min-relations.md
+++ b/rfcs/min-relations.md
@@ -255,6 +255,7 @@ Relation enhancements:
 7. Streaming iters to allow for queries like: `Query<&mut Money, Relation<Owes, &mut Money>>` and potentially use invariants on graph shape to allow for skipping soundness checks for the aliasing `&mut Money`'s
 8. Assorted Performance optimizations.
 9. Relation orders to enable things like `Styles`.
+10. Generalized `despawn_recursive`.
 
 Relation applications in the engine:
 
--- a/rfcs/min-relations.md
+++ b/rfcs/min-relations.md
@@ -1,0 +1,73 @@
+# Feature Name: `min-relations`
+
+## Summary
+
+One paragraph explanation of the feature.
+
+## Motivation
+
+Why are we doing this? What use cases does it support?
+
+## Guide-level explanation
+
+Explain the proposal as if it was already included in the engine and you were teaching it to another Bevy user. That generally means:
+
+- Introducing new named concepts.
+- Explaining the feature, ideally through simple examples of solutions to concrete problems.
+- Explaining how Bevy users should *think* about the feature, and how it should impact the way they use Bevy. It should explain the impact as concretely as possible.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, explain how this feature compares to similar existing features, and in what situations the user would use each one.
+
+## Reference-level explanation
+
+This is the technical portion of the RFC. Explain the design in sufficient detail that:
+
+- Its interaction with other features is clear.
+- It is reasonably clear how the feature would be implemented.
+- Corner cases are dissected by example.
+
+The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work.
+
+## Drawbacks
+
+Why should we *not* do this?
+
+## Rationale and alternatives
+
+- Why is this design the best in the space of possible designs?
+- What other designs have been considered and what is the rationale for not choosing them?
+- What is the impact of not doing this?
+- Why is this important to implement as a feature of Bevy itself, rather than an ecosystem crate?
+
+## \[Optional\] Prior art
+
+Discuss prior art, both the good and the bad, in relation to this proposal.
+This can include:
+
+- Does this feature exist in other libraries and what experiences have their community had?
+- Papers: Are there any published papers or great posts that discuss this?
+
+This section is intended to encourage you as an author to think about the lessons from other tools and provide readers of your RFC with a fuller picture.
+
+Note that while precedent set by other engines is some motivation, it does not on its own motivate an RFC.
+
+## Unresolved questions
+
+- What parts of the design do you expect to resolve through the RFC process before this gets merged?
+- What parts of the design do you expect to resolve through the implementation of this feature before the feature PR is merged?
+- What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
+
+## \[Optional\] Future possibilities
+
+Think about what the natural extension and evolution of your proposal would
+be and how it would affect Bevy as a whole in a holistic way.
+Try to use this section as a tool to more fully consider other possible
+interactions with the engine in your proposal.
+
+This is also a good place to "dump ideas", if they are out of scope for the
+RFC you are writing but otherwise related.
+
+Note that having something written down in the future-possibilities section
+is not a reason to accept the current or a future RFC; such notes should be
+in the section on motivation or rationale in this or subsequent RFCs.
+If a feature or change has no direct value on its own, expand your RFC to include the first valuable feature that would build on it.
--- a/rfcs/min-relations.md
+++ b/rfcs/min-relations.md
@@ -187,7 +187,7 @@ fn debts_come_due(
 	for (mut debtor_money, debt) in debtor_query.iter(){
 		for (lender, amount_owed) in debt {
 			let mut lender_money = lender_query.get().unwrap();
-			if debtor_money >= amount_owed {
+			if debtor_money.dollars >= amount_owed.dollars {
 				debtor_money.dollars -= amount_owed.dollars;
 				lender_money.dollar += amount_owed.dollar;
 			} else {
--- a/rfcs/min-relations.md
+++ b/rfcs/min-relations.md
@@ -186,7 +186,7 @@ fn debts_come_due(
 ) {
 	for (mut debtor_money, debt) in debtor_query.iter(){
 		for (lender, amount_owed) in debt {
-			let mut lender_money = lender_query.get().unwrap();
+			let mut lender_money = lender_query.get(lender).unwrap();
 			if debtor_money.dollars >= amount_owed.dollars {
 				debtor_money.dollars -= amount_owed.dollars;
 				lender_money.dollar += amount_owed.dollar;
--- a/rfcs/min-relations.md
+++ b/rfcs/min-relations.md
@@ -45,7 +45,10 @@ The section should return to the examples given in the previous section, and exp
 
 ## Drawbacks
 
-Why should we *not* do this?
+1. Introduces another core abstraction to users that they will want to learn.
+Rewriting our parent-child code in terms of relations is important, but will introduce the concept to users quite quickly.
+2. Even if relations are more ergonomic than the Entity-in-component approach to accomplish the same tasks, relations will be held to a higher bar for quality.
+3. Heavy use of relations can seriously fragment archetypes. We need to be careful that this doesn't badly impact our performance.
 
 ## Rationale and alternatives
 
--- a/rfcs/min-relations.md
+++ b/rfcs/min-relations.md
@@ -2,11 +2,26 @@
 
 ## Summary
 
-One paragraph explanation of the feature.
+Relations connect entities to each other, storing data in these edges. This RFC covers the minimal version of this feature, allowing for basic entity grouping and simple game logic extensions.
 
 ## Motivation
 
-Why are we doing this? What use cases does it support?
+Entities often need to be aware of each other in complex ways: an attack that targets another unit, groups of entities that are controlled en-masse, or complex parent-child hierarchies that need to move together.
+
+We *could* simply store an `Entity` in a component, then use `query.get(my_entity)` to access the relevant data.
+But this quickly starts to balloon in complexity.
+
+- How do we ensure that these components are cleaned up when the entity they're pointing to is?
+- How do we handle pointing to multiple entities in similar ways?
+- How do we look for all entities that point to a particular entity of interest?
+- How do we despawn an entity and all of its children (with many types of child-like entities)?
+- How do we quickly access data on the entity that we're attached to?
+- How do we traverse this graph of entities?
+- How do we ensure that our graph is acylic?
+
+We *could* force our users to solve these problems, over and over again in each individual game.
+
+This powerful and common pattern deserves a proper abstraction that we can make ergonomic, performant and *fearless*.
 
 ## Guide-level explanation
 
--- a/rfcs/min-relations.md
+++ b/rfcs/min-relations.md
@@ -63,15 +63,15 @@ commands.entity(alice).remove_relation(FriendsWith, boxy) // not uwu :(
 // You can use direct World access!
 world.entity_mut(alice).insert_relation(FriendsWith, boxy); // uwu:3
 
-// Relations are directed
-commands.entity(boxy).insert_relation(FriendsWith, alice); // one way friendship sad oh wait you reinserted your one nvm rip my joke :(
+// Relations are directed; a relation in one direction does not imply the same in the other direction
+commands.entity(boxy).insert_relation(FriendsWith, alice); // one way friendship sad :(
 
 // You can mix different relation kinds on the same entity!
-world.commands.entity(alice).insert_relation(Owes(9999), boxy); // :)))))
+commands.entity(alice).insert_relation(Owes(9999), boxy); // :)))))
 
 // You can add mulitple relations of the same kind to a single source entity!
 let cart  = world.spawn().id();
-world.commands.entity(alice).insert_relation(FriendsWith, cart); // Hi!
+commands.entity(alice).insert_relation(FriendsWith, cart); // Hi!
 ```
 
 Once your relations are in place, you'll want to access them in some form.
@@ -87,7 +87,7 @@ fn friendship_is_magic(mut query: Query<(&mut Magic, &Relation<FriendsWith>), Wi
 	}
 }
 
-// Or you can request mutable access to modify their data
+// Or you can request mutable access to modify the relation's data
 fn interest(query: Query<&mut Relation<Owes>>){
 	for debts in query.iter(){
 		for (owed_to, amount_owed) in debts {
@@ -124,10 +124,12 @@ fn caught_in_the_middle(mut query: Query<&mut Stress, With<Relation<FriendsWith>
 	
 	// You can directly chain this as set_relation_filters returns &mut Query
 	query.set_relation_filters(
+		// Note that we can set relation filters even if the relation is in a query filter
         RelationFilters::new()
             .add_target_filter::<FriendsWith, _>(dorothy.entity)
 			.add_target_filter::<FriendsWith, _>(elphaba.entity)
     ).for_each(|mut stress| {
+		println!("{} is friends with Dorothy!", source_entity);
 		stress.val = 1000; 
 	})
 }
--- a/rfcs/min-relations.md
+++ b/rfcs/min-relations.md
@@ -59,15 +59,9 @@ Note that while precedent set by other engines is some motivation, it does not o
 
 ## \[Optional\] Future possibilities
 
-Think about what the natural extension and evolution of your proposal would
-be and how it would affect Bevy as a whole in a holistic way.
-Try to use this section as a tool to more fully consider other possible
-interactions with the engine in your proposal.
-
-This is also a good place to "dump ideas", if they are out of scope for the
-RFC you are writing but otherwise related.
-
-Note that having something written down in the future-possibilities section
-is not a reason to accept the current or a future RFC; such notes should be
-in the section on motivation or rationale in this or subsequent RFCs.
-If a feature or change has no direct value on its own, expand your RFC to include the first valuable feature that would build on it.
+1. GRAPH
+2. Kinded entities
+3. Use with UI.
+4. Frustum culling.
+5. Replace parent-child.
+6. Reverse relations.
--- a/rfcs/min-relations.md
+++ b/rfcs/min-relations.md
@@ -1,5 +1,10 @@
 # Feature Name: `min-relations`
 
+// FIXME
+// Make sure to talk about these:
+// - Reverse relations.
+// - Or support for add_target_filter.
+
 ## Summary
 
 Relations connect entities to each other, storing data in these edges. This RFC covers the minimal version of this feature, allowing for basic entity grouping and simple game logic extensions.
@@ -9,6 +14,7 @@ Relations connect entities to each other, storing data in these edges. This RFC 
 Entities often need to be aware of each other in complex ways: an attack that targets another unit, groups of entities that are controlled en-masse, or complex parent-child hierarchies that need to move together.
 
 We *could* simply store an `Entity` in a component, then use `query.get(my_entity)` to access the relevant data.
+(In this RFC, we call this the **Entity-in-component pattern**.)
 But this quickly starts to balloon in complexity.
 
 - How do we ensure that these components are cleaned up when the entity they're pointing to is?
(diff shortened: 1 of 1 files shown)


rfcs/min-relations.md
```diff
@@ -0,0 +1,273 @@
...
+
+TODO: mention relation cleanup on despawn.
+
+Sometimes, we may instead want to look for "entities that are targets of some relation".
+We can do so by iterating over all entities with that relation, and then filtering by target.
+
+```rust
+fn all_children(
+    player_query: Query<Entity, With<Player>>,
+    child_query: Query<&Name, (With<Relation<ChildOf>, Without<Player>)>>,
```
alice-i-cecile, line 168:
```suggestion
    child_query: Query<&Name, With<Relation<ChildOf>>>,
```

rfcs/min-relations.md
```diff
@@ -0,0 +1,273 @@
...
+        println!("{} is one of my children.", name) 			
+    }
+}
+```
+
+Finally, you can use the `Entity` returned by your relations to fetch data from the target's components by combining it with `query::get()` or `query::get_component<C>()`.
+
+```rust
+// We need to use a marker component to avoid having multiple mutable references
+fn debts_come_due(
```
alice-i-cecile, line 187: Replace with:
fn debts_come_due(
    mut debt_query: Query<Relation<Owes>>, 
    mut money_query: Query<&mut Money>,
) {}

Courtesy of @TheRawMeatball.

rfcs/min-relations.md
```diff
@@ -0,0 +1,270 @@
+# Feature Name: `min-relations`
+
+// FIXME
+// Make sure to talk about these:
+// - Reverse relations.
+// - Or support for add_target_filter.
+
+## Summary
+
+Relations connect entities to each other, storing data in these edges. This RFC covers the minimal version of this feature, allowing for basic entity grouping and simple game logic extensions.
```
Frizi, line 10:
```suggestion
Relations allow creating logical connection between entity to multiple other entities. They also allow storing data associated with those individual connections, known as edges. Queries areexpanded to allow efficient filtering of entities that relate to specified group of entities.

This RFC covers the minimal version of this feature, allowing for basic dynamic entity grouping and simple game logic extensions.
```

rfcs/min-relations.md
```diff
@@ -0,0 +1,270 @@
...
+
+## Motivation
+
+Entities often need to be aware of each other in complex ways: an attack that targets another unit, groups of entities that are controlled en-masse, or complex parent-child hierarchies that need to move together.
```
Frizi, line 14: I think the `InFrustum(camera_id)` would be especially understandable here. It's easy to see why it can't be handled by simple tag component here.

alice-i-cecile: I worry that multiple camera frustums are bit too obscure of a topic for the average gameplay programmer referring to this RFC. I agree that they're a clear example though!

rfcs/min-relations.md
```diff
@@ -0,0 +1,270 @@
...
+(In this RFC, we call this the **Entity-in-component pattern**.)
+But this quickly starts to balloon in complexity.
+
+- How do we ensure that these components are cleaned up when the entity they're pointing to is?
+- How do we handle pointing to multiple entities in similar ways?
+- How do we look for all entities that point to a particular entity of interest?
+- How do we despawn an entity and all of its children (with many types of child-like entities)?
+- How do we quickly access data on the entity that we're attached to?
+- How do we traverse this graph of entities?
+- How do we ensure that our graph is acylic?
```
Frizi, line 26: I'm not sure if this particular point is handled in any way by proposed relations. Afaict cycles are still an issue (in cases where you explicitly don't want them).

rfcs/min-relations.md
alice-i-cecile: Relations allow users to create logical connection between entities. Each source entity can have multiple kinds of relations, each with their own data, and multiple relations of the same kind pointing to a different target entity.
Queries are expanded to allow efficient filtering of entities by the relations they have or the targets of these relations.

rfcs/min-relations.md
alice-i-cecile: The framing could be clearer. These problems are *possible* to address in a cohesive way at an engine-level by including relations as a feature. Min-relations won't hit all of these though :(

rfcs/min-relations.md
```diff
@@ -0,0 +1,270 @@
...
+
+We *could* force our users to solve these problems, over and over again in each individual game.
+
+This powerful and common pattern deserves a proper abstraction that we can make ergonomic, performant and *fearless*.
+
+## Guide-level explanation
+
+**Relations** are components that connect entities together, pointing from the entity they are stored on to the entity that they target.
+Just like other components, they can store effectively arbitrary data, or they can be data-less unit structs that serve as **markers**.
```
Frizi, line 38:
```suggestion
**Relations** are similar to components that connect entities together, pointing from the entity they are stored on to the entity that they target.
Just like components, they can store effectively arbitrary data, or they can be data-less unit structs that serve as **markers**. Unlike components, they don't store the target entity directly, as that connection is managed by the engine.
```

rfcs/min-relations.md
Frizi: Okay, so it's more of a future prospect. I guess that's fine.

rfcs/min-relations.md
```diff
@@ -0,0 +1,273 @@
...
+    puppy: Res<Puppy>
+) {
+    for lonely_entity in query.iter(){
+        commands.insert_relation(FriendsWith, puppy.entity);
+    }
+}
+
+// So does change detection with Added and Changed!
+fn new_friends(query: Query<&mut Excitement, Added<Relation<FriendsWith>>>){
+    query.for_each(|mut excitement|{excitement.value += 10});
```
Frizi, line 158:
```suggestion
    query.for_each(|(mut excitement, friends)|{
    	excitement.value += 10 * friends.count();
	});
```

rfcs/min-relations.md
```diff
@@ -0,0 +1,273 @@
...
+
+```
+
+TODO: mention relation cleanup on despawn.
+
+Sometimes, we may instead want to look for "entities that are targets of some relation".
+We can do so by iterating over all entities with that relation, and then filtering by target.
+
+```rust
+fn all_children(
```
Frizi, line 171: This example is equivalent to `friend_of_dorothy` and could probably be removed.

rfcs/min-relations.md
```diff
@@ -0,0 +1,273 @@
...
+    for name in child_query.iter() {
+        println!("{} is one of my children.", name) 			
+    }
+}
+```
+
+Finally, you can use the `Entity` returned by your relations to fetch data from the target's components by combining it with `query::get()` or `query::get_component<C>()`.
+
+```rust
+// We need to use a marker component to avoid having multiple mutable references
```
Frizi, line 191: There is no marker component in the example.

rfcs/min-relations.md
alice-i-cecile: I think I agree. We were originally going to use this to demonstrate reverse relations, but we cut those from min_relations.

rfcs/min-relations.md
```diff
@@ -0,0 +1,273 @@
...
+
+### Entity hierarchies
+
+## Reference-level explanation
+
+TODO: Boxy explains the magic.
+
+## Drawbacks
+
+1. Introduces another core abstraction to users that they will want to learn.
```
Frizi, line 217:
```suggestion
1. Introduces another core abstraction to users that they will have to learn.
```

I think it was a little too optimistic :P

rfcs/min-relations.md
alice-i-cecile: Yeah; Meatball fixed this example for us! We can remove that comment now.

rfcs/min-relations.md
alice-i-cecile: Sadly you're right.

rfcs/min-relations.md
```diff
@@ -0,0 +1,273 @@
...
+Rewriting our parent-child code in terms of relations is important, but will introduce the concept to users quite quickly.
+2. Even if relations are more ergonomic than the Entity-in-component approach to accomplish the same tasks, relations will be held to a higher bar for quality.
```
Frizi, line 219: Is this a drawback? Easier and higher quality at the same time sounds like a very good thing.

alice-i-cecile: Hmmm... It's more work on the devs though! I agree though, this can probably be cut.

rfcs/min-relations.md
```diff
@@ -0,0 +1,273 @@
...
+9. Relation orders to enable things like `Styles`.
+10. Generalized `despawn_recursive`.
+
+Relation applications in the engine:
+
+1. Replace existing parent-child.
+2. Applications for UI: styling, widget wiring etc.
+3. Frustum culling.
+
+AUTHOR'S NOTE: Much like `yeet`, `Noitaler` is the bikeshed-avoidance name for the inverse of `Relation`, and will never be used.
```
Frizi, line 274:
```suggestion
[bikeshed]: AUTHOR'S NOTE: Much like `yeet`, `Noitaler` is the bikeshed-avoidance name for the inverse of `Relation`, and will never be used.
```

rfcs/min-relations.md
```diff
@@ -0,0 +1,273 @@
...
+## Prior art
+
+[`Flecs`](https://github.com/SanderMertens/flecs), an advanced C++ ECS framework, has a similar feature, which they call "relationships".
+These are somewhat different, they use an elaborate query domain-specific language along with being more feature rich.
+You read more about them in the [corresponding PR](https://github.com/SanderMertens/flecs/pull/358).
+
+## Unresolved questions
+
+1. How do we put relations in Bundles?
+2. What do we actually call `Noitaler`? Candidates so far are:
```
Frizi, line 238:
```suggestion
2. What do we actually call [`Noitaler`][[bikeshed]? Candidates so far are:
```

rfcs/min-relations.md
alice-i-cecile: Clarified!

rfcs/min-relations.md
alice-i-cecile: Removed.

rfcs/min-relations.md
alice-i-cecile: Fixed.

rfcs/min-relations.md
alice-i-cecile: Cut.

rfcs/min-relations.md
```diff
@@ -0,0 +1,438 @@
...
+This would enable the `Styles` proposal from #1 to use relations.
+10. Generalized `despawn_recursive` by parameterizing on relation type.
+
+Relation applications in the engine:
+
+1. Replace existing parent-child API.
+2. Applications for UI: styling, widget wiring etc.
+3. Frustum culling with multiple cameras.
+
+[bikeshed]: AUTHOR'S NOTE: Much like `yeet`, `Noitaler` is the bikeshed-avoidance name for the inverse of `Relation`, and will never be used.
```
Frizi, line 437:
```suggestion
<b id="bikeshed">AUTHOR'S NOTE</b>: Much like `yeet`, `Noitaler` is the bikeshed-avoidance name for the inverse of `Relation`, and will never be used.

```

rfcs/min-relations.md
```diff
@@ -0,0 +1,438 @@
...
+## Prior art
+
+[`Flecs`](https://github.com/SanderMertens/flecs), an advanced C++ ECS framework, has a similar feature, which they call "relationships".
+These are somewhat different, they use an elaborate query domain-specific language along with being more feature rich.
+You read more about them in the [corresponding PR](https://github.com/SanderMertens/flecs/pull/358).
+
+## Unresolved questions
+
+1. How do we put relations in Bundles?
+2. What do we actually call [`Noitaler`][[bikeshed]? Candidates so far are:
```
Frizi, line 397:
```suggestion
2. What do we actually call [`Noitaler`](#bikeshed)? Candidates so far are:
```

rfcs/min-relations.md
```diff
@@ -0,0 +1,438 @@
...
+let cart  = world.spawn().id();
+commands.entity(alice).insert_relation(FriendsWith, cart); // Hi!
+```
+
+Once your relations are in place, you'll want to access them in some form.
+Just like with ordinary components, you can request them in your queries:
+
+```rust
+// Relations can be queried for immutably
+fn friendship_is_magic(mut query: Query<(&mut Magic, &Relation<FriendsWith>), With<Pony>>) {
```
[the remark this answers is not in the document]
Frizi: Currently reference is necessary in query directly, but not used in combinators like `With<Relation<T>>`. It is used to determine if access is read-only or read-write. This is similar to how references are used on component types here.

rfcs/min-relations.md
```diff
@@ -0,0 +1,438 @@
...
+    elphaba: Res<Elphaba>){
+
+    // Note that we can set relation filters even if the relation is in a query filter
+    query.filter_relation::<FriendsWith, _>(
+        RelationFilter::any_of()
+        .target(dorothy.entity)
+        .target(elphaba.entity)
+    // You can directly chain this as set_relation_filters returns &mut Query
+    ).build().for_each(|mut stress| {
+        println!("{} is friends with Dorothy and Elphaba!", source_entity);
```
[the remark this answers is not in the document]
Frizi: Yep, I believe this should be `RelationFilter::all_of()` to make sense in this example.

rfcs/min-relations.md
```diff
@@ -0,0 +1,438 @@
...
+    });
+}
+
+```
+
+You can use the `Entity` returned by your relations to fetch data from the target's components by combining it with `query::get()` or `query::get_component<C>()`.
+
+```rust
+fn debts_come_due(
+    mut debt_query: Query<(Entity, Relation<Owes>)>,
```
Frizi, line 160:
```suggestion
    mut debt_query: Query<(Entity, &Relation<Owes>)>, 
```

rfcs/min-relations.md
```diff
@@ -0,0 +1,438 @@
...
+
+Each relation kind + target combination is treated like a unique type for the purposes of archetypes.
+This makes can be relevant to the performance of your game; accelerating target entity filtering but creating overhead if you have a very large number of distinct targets.
+As a result, changing the source or target of a relation can only be done while the app has exclusive access to the world, via `Commands` or in exclusive systems.
+Here's an example of how you might do so:
+
+```rust
+fn love_potion(
+    commands: mut Commands,
+    query: Query<Entity, (With<Relation<&Loves>>, With<Cute>)>,
```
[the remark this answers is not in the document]
Frizi: That looks wrong indeed.
```suggestion
    query: Query<Entity, (With<Relation<Loves>>, With<Cute>)>,
```

rfcs/min-relations.md
```diff
@@ -0,0 +1,438 @@
...
+        // these command applies to any relations that match,
+        // and overwrite any conflicting relations
+        commands.entity(victim).change_target::<Loves>(player); // This is unethical!!
+    }
+}
+
+fn adoption(
+    commands: mut Commands,
+    // As with components, you can query for relations that may or may not exist
+    query: Query<(Entity, &NewOwner, Option<Relation<Owns>>), With<Kitten>>,
```
Frizi, line 204:
```suggestion
    query: Query<(Entity, &NewOwner, Option<&Relation<Owns>>), With<Kitten>>,
```

rfcs/min-relations.md
```diff
@@ -0,0 +1,438 @@
...
+```
+
+### Advanced relation filters
+
+As the savvy reader may have guessed, we can combine relation filters in arbitrarily complex ways.
+While most of you may never need to plumb the dizzying heights of compound relation filtering,
+the DBAs in the audience will surely both thank us for our efforts and deride our naive implementations.
+
+Fortunately, there are several simple and convenient API extensions that you might want to use.
+The first of these is the ability to filter by the source entity as well:
```
Frizi, line 224: I'm not sure this whole paragraph makes much sense in the RCF concext. This is imho a bit too blog-like for techical documentation.

```suggestion
For those cases where previous examples weren't enough, there are several additional API extensions that you might want to use. Using those, Relation filters can be combined in arbitrarily complex ways.

The first of these is the ability to filter by the source entity as well:

```

rfcs/min-relations.md
```diff
@@ -0,0 +1,438 @@
...
+```rust
+
+```
+
+Finally, for when you have *truly* complex relation filtering needs, you can turn to **compound relation filters**.
+By combining `RelationFilter::not` with the power of `RelationFilter::inner_join`, `RelationFilter::left_join`, `RelationFilter::right_join`, `RelationFilter::full_join`, you can extract the information you desire from even the most convoluted of architectures!
+As you might expect, these are `(RelationFilter, RelationFilter) -> RelationFilter` functions,
+and act like the [standard database joins](http://www.sql-join.com/sql-join-types) of the same names.
+
+No example is given: if you need these, you will know what to do.
```
Frizi, line 277: Is this a part of `min-relations`? Also at the point when we do actually specify them, there should be an example given. But I think this should be only lightly mentioned under future work right now.

rfcs/min-relations.md
```diff
@@ -0,0 +1,438 @@
...
+`any_of` and `all_of` can be used to collect groups of entities into a single filter.
+`any_of` uses **or semantics**, returning any entity if any of the filters are met,
+while `all_of` uses **and semantics**, rejecting entities who do not meet every specified filter.
+
+```rust
+fn 
+
+```
+
+We can filter on multiple types of relations at once by chaining together our `.filter_relation` methods before we call `.build()`.
```
Frizi, line 256: The `or/and` semantics of multiple filter relations have to be clarified.
```rust
query.filter_relation(a).filter_relation(b).build()
```
Is it "return everything that matches either `a` or `b`" or is it "return everything that matches both `a` and `b` at the same time"?

rfcs/min-relations.md
```diff
@@ -0,0 +1,438 @@
...
+
+```rust
+fn 
+
+```
+
+Chaining filters in this way uses "and" semantics, just like `all_of`.
```
Frizi, line 263:
````suggestion
Chaining filters in this way uses "and" semantics, just like `all_of`. Following query will only return entities that match both `a` and `b` relation filters at the same time.
```rust
query.filter_relation(a).filter_relation(b).build()
```

````

File: rfcs/min-relations.md [hunk 25]
```diff
@@ -0,0 +1,438 @@
...
+
+While relations filters are reset each time you leave a system, you may want to reset them sooner,
+commonly when performing multiple subqueries in bevy. To do so, simply call `query.reset_filter::<R>()` or `query.reset_filters()`.
```
Frizi, line 266: Why do filters need a reset? What kind of state do they hold? I feel like starting fresh iterator on the query should always give me the same thing.

I presume the idea here is to "remove previously added relation filters". In that case, how about `query.remove_relation_filter::<R>()` and `query.clear_relation_filters()`?

Also this remove method accepts a relation type and nothing else. Does that mean I can only add a single filter per given relation type in the first place? Or does it remove all added filters with this relation type?

rfcs/min-relations.md
```diff
@@ -0,0 +1,438 @@
...
+// You can query for entities that target a specific entity
+fn friend_of_dorothy(
+    mut query: Query<Entity, With<Relation<FriendsWith>>>, 
+    dorothy: Res<Dorothy>,
+) {
+    // Setting relation filters requires mutable access to the query
+    query.filter_relation::<FriendsWith, _>(
+        RelationFilter::target(dorothy.entity)
+    // .build() causes the relation filters to be applied
+    ).build();
```
Frizi, line 108: Why is builder pattern necessary here? If we do that, then the builder should return something.

```rust
let mut filtered_query = query.filter_relation(..).build();
```

Otherwise, adding more filters would be possible after building. And also querying things without calling `build` would lead to bugs, as the filters would be not applied.

rfcs/min-relations.md
```diff
@@ -0,0 +1,438 @@
...
+            *money_query.get(lender).unwrap() += amount_owed;
+        }
+    }
+}
+```
+
+### Nuances of relations
+
+Relations between entities are automatically cleaned up when either the source or target entity is removed.
+Just like components, you can view which relations were removed in the last frame through the `RemovedRelation<T>` system parameter, which also returns their data.
```
Frizi, line 177: Why is `RemovedRelation` a separate thing? Should we use `Removed<Relation<T>>` instead?

rfcs/min-relations.md
```diff
@@ -0,0 +1,438 @@
...
+    commands: mut Commands,
+    // As with components, you can query for relations that may or may not exist
+    query: Query<(Entity, &NewOwner, Option<Relation<Owns>>), With<Kitten>>,
+) {
+   for (kitten, new_owner, previous_ownership) in query.iter(){
+       // Changing the target or the source will fail silently if no appropriate relation exists
+       match previous_ownership {
+           // The first parameter is the data, the second is the old target
+           Some(_,_) => commands.entity(kitten).change_source::<Owns>(new_owner); // uwu :3
+           None => commands.entity(kitten).insert_relation(Owns::default(), new_owner); // uwu!!
```
Frizi, line 211: I'm not sure about this one, but is the relation reversed here?
```suggestion
           None => commands.entity(new_owner).insert_relation(Owns::default(), kitten); // uwu!!
```

rfcs/min-relations.md
```diff
@@ -0,0 +1,438 @@
...
+You read more about them in the [corresponding PR](https://github.com/SanderMertens/flecs/pull/358).
+
+## Unresolved questions
+
+1. How do we put relations in Bundles?
+2. What do we actually call [`Noitaler`][[bikeshed]? Candidates so far are:
+   1. `ReverseRelation`
+   2. `Inverse`
+   3. `CoRelation`
+   4. `TargetOf`
```
Frizi, line 401:
```suggestion
   4. `TargetOf`
   5. 'ConverseRelation`
```

rfcs/min-relations.md
```diff
@@ -0,0 +1,438 @@
...
+These are somewhat different, they use an elaborate query domain-specific language along with being more feature rich.
+You read more about them in the [corresponding PR](https://github.com/SanderMertens/flecs/pull/358).
+
+## Unresolved questions
+
+1. How do we put relations in Bundles?
+2. What do we actually call [`Noitaler`][[bikeshed]? Candidates so far are:
+   1. `ReverseRelation`
+   2. `Inverse`
+   3. `CoRelation`
```
Frizi, line 400: I know this isn't exactly the time to bikeshed, but I think we should remove this one. It's too similar to "corelation", which means totally diffrent thing.

rfcs/min-relations.md
Frizi, line 401: I know this isn't exactly the time to bikeshed, but I think we should remove "CoRelation". It's too similar to "corelation", which means totally diffrent thing.

```suggestion
   1. `ReverseRelation`
   2. `Inverse`
   3. `ConverseRelation`
   4. `TargetOf`
```

rfcs/min-relations.md
alice-i-cecile: I would prefer this, but I'm following the convention from components.

rfcs/min-relations.md
alice-i-cecile: Right, the relation should be owned by...

rfcs/min-relations.md
alice-i-cecile: This will be cleaned up shortly to a much simpler paradigm thanks to Boxy.

rfcs/min-relations.md
alice-i-cecile: IMO this is better as a separate option. "Co-relation" is the obvious term for readers with a background in advanced mathematics, even if they are a tiny minority.

rfcs/min-relations.md
alice-i-cecile: You're right; making the fix.

rfcs/min-relations.md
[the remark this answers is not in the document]
alice-i-cecile: Fixed!

rfcs/min-relations.md
alice-i-cecile: I'm planning to add more in depth examples here.

rfcs/min-relations.md
Frizi: I mean, I don't have anything strong to say here. In fact i've made a typo there, I wanted to point out similarity to "correlation" :P If you are not convinced to remove or replace this potential name, ignore this comment.

rfcs/min-relations.md
Frizi: I see, it's not a query combinator, but a system param. Then the convention really should be `RemovedRelations<T>`, which probably would be an iterator over `(Entity, RemovedTargets)` where `RemovedTargets` is another iterator over `Entity` (or whatever the associated Target type will be in the future).

rfcs/min-relations.md
```diff
@@ -0,0 +1,473 @@
...
+
+### Advanced relation filters
+
+For those cases where previous examples weren't enough, there are several additional API extensions that you might want to use. Using those, Relation filters can be combined in arbitrarily complex ways.
+
+The first of these is the ability to filter by the source entity as well:
+
+
+```rust
+fn
```
bjorn3, line 260:
```suggestion
// TODO complete this
fn 
```

rfcs/min-relations.md
```diff
@@ -0,0 +1,479 @@
...
+fn reattach_springs(mut commands: Commands,
+                   selected_mass1: Res<FirstSelected>,
+                   selected_mass2: Res<SecondSelected>,
+                   query: Query<(Entity, Relation<Spring>)>
+                  ) {
+    
+    let m1 = selelected_mass1.entity;
+    let m2 = selelected_mass2.entity;
+
+    for (source_mass, spring) in query.iter_mut(){
```
Frizi, line 231: this example could use `query.get_mut(source_mass)` to avoid the outside loop. We know which entity's relations we are interested in directly.

rfcs/min-relations.md
alice-i-cecile: > I wanted to point out similarity to "correlation" :P

Yeah, this is the main reason I dislike CoRelation and don't think it's adequate :(. Capitalizing it like `CoRelation` is better than `Corelation` for sure, but it doesn't help enough.

rfcs/min-relations.md
alice-i-cecile: Ah, you're right there. I'll add the `s`. Down the line, I'd prefer to see Removed* move to subworlds, but that's out of scope here.

rfcs/min-relations.md
alice-i-cecile: @BoxyUwU was explaining to me that the builder pattern was a necessary API sacrifice for performance in order to... something about archetype iteration? I'll add a to-do in rationale.

I think I agree with the comments about the return value though; I'll make that change.

alice-i-cecile: Oh hey, using assignment allows us to avoid needing mutable access *and* eliminates the need for a reset_filter API.

I'm in favor of this.

rfcs/min-relations.md
```diff
@@ -205,7 +215,7 @@ fn adoption(
        match previous_ownership {
            // We can change sources by controlling which entity owns the relation
            // move_relation is directly analagous to move_component
-           Some(_, old_owner) => commands.entity(old_owner).move_relation::<Owns>(new_owner, kitten); // uwu :3
+           Some(old_owner, _) => commands.entity(old_owner).move_relation::<Owns>(new_owner, kitten); // uwu :3
```
Frizi, line 218:
```suggestion
           Some((old_owner, _)) => commands.entity(old_owner).move_relation::<Owns>(new_owner, kitten); // uwu :3
```

rfcs/min-relations.md
```diff
@@ -238,24 +248,123 @@ fn reattach_springs(mut commands: Commands,
...
+            // This is just a dummy to be replaced
+            let best_move = Entity::new();
+            let current = current_move.compute_heuristic()
+
+            if current > current_best{
+                current_best = current;
+                best_move = current_move;
+            }
+        }
+        commands.insert_relation::<NextMove>(piece, best_move);
```
Frizi, line 284: this `for` loop could be replaced with simple `.max_by_key(..)`;
```suggestion
		let best_move = valid_moves
			.max_by_key(|(potential_position, _)| potential_position.compute_heuristic());
			
		if let Some((_, best_move)) = best_move {
			commands.insert_relation::<NextMove>(piece, best_move);
		}
```

also changed `compute_heuristic` to be a method of `potential_position`, as it made more sense to me and also uses both tuple arguments. Please correct me if that's wrong for some reason.

rfcs/min-relations.md
```diff
@@ -422,11 +422,27 @@ We can even combine positive and negative filters by including multiple copies o
...
+    plants_query: Query<Entity, With<Plant>>,
+    animals_query: Query<Enity, With<Animal>>,
+    species_query: Query<Entity, (With<Relation<Consumes>, Without<Relation<Consumes>>)>
+){
+    let plants = plants_query.iter().collect();
+    let animals = animals_query.iter().collect();
+
+    species_query
+        // The second type parameter controls which relation filter of that type is being modified
+        .filter_relation::<Consumes, InFilter<{1}>>(RelationFilter::any_of().targets(plants))
```
Frizi, line 436: I'm not sure about this one. Adding extra generic param to that function call will always make it required to list, even if this is unambiguous, e.g. like thtat
```rust
        .filter_relation::<Owes, _>>(target)
```
which just doesn't look good. Preferably we would structure it such that this generic can be auto-inferred somewhere, which should make the compiler solve the unambiguous case (given relation type is used only once in the query) without any extra help.

I know it's possible, because that's how `frunk` crate infers the `Indices` to operate on. You don't have to do anything if there is only one valid and provable set of indices for given operation. But it might require some extra gymnastics around generics for the relation filter itself or something. Also it would be nice if the index could be specified at runtime, but this is secondary.

Frizi: Alternative syntax that avoid `_`:
```rust
        .filter_relation::(RelationFilter::<Owes>::target(target))
```

rfcs/min-relations.md
```diff
@@ -371,10 +371,29 @@ From time-to-time, we may care about *excluding* entities who have relations to
 To do so, we set our relation filter on a `Without<Relation<R>>` query parameter.
 
 ```rust
-fn blackball(){
+// Fraternizing with communists is a strict disqualifcation!
```
bjorn3, line 390:
```suggestion
// Fraternizing with communists is a strict disqualification!
```

rfcs/min-relations.md
alice-i-cecile: This second type parameter is definitely my largest complaint with the current design as well.

rfcs/min-relations.md
alice-i-cecile: Awesome. I agree with all of those changes.

alice-i-cecile: Applied now.

alice-i-cecile: This change has been applied <3

rfcs/min-relations.md
```diff
@@ -0,0 +1,644 @@
...
+2. The number of different groups you need is unknowable at compile time.
+
+Here are some concrete examples where this pattern works well:
+
+- you're storing multiple parts (such as meshes or sprites with their own `Transform`s) of the same in-game-object across multiple entities
+- you want to dynamically group units together, like in a real-time-strategy game
+- the effects created by a single entity, like the bullets they're spewing out
+- you want a way to mark the controller of a unit in a queryable way, but the number of possible controllers is unknown at compile time
+- you have multiple cameras, and want to mark which frustum each entity is in so they can be culled independently
+
```
alice-i-cecile, line 483: Per feedback, this needs a simple worked example to show users what this looks like in practice :p

rfcs/min-relations.md
```diff
@@ -0,0 +1,644 @@
...
+
+```rust
+// You can use the `Entity` target returned by your relations to fetch data from the 
+// target's components by combining it with `query::get()` or `query::get_component<C>()`.
+fn debts_come_due(
+    mut debt_query: Query<(Entity, &Relation<Owes>)>,
+    mut money_query: Query<&mut Money>,
+) {
+    for (debtor, debt) in debt.query() {
```
tigregalis, line 178: Sorry, meant to do this as a review.
```suggestion
    for (debtor, debt) in debt_query.iter() {
```

tigregalis:
> Heavy use of relations can seriously fragment archetypes. We need to be careful that this doesn't badly impact our performance.

Why is this the case? Relations could build on SparseSet - I guess the question is, now does every entity have a Relation?

An alternative approach: just as the ECS is essentially a columnar database with tables for each archetype, borrowing from SQL you could store a "junction table" for each Relation kind separate from the Components, that is to say, not on the entity itself. I don't think you'd need exclusive world access anymore either. To illustrate, something like:

#### FriendsWith
| Source        | Target        | Data         |
| ------------- | ------------- |:------------:|
| alice         | boxy          | FriendsWith  |
| boxy          | alice         | FriendsWith  |
| alice         | cart          | FriendsWith  |

#### Owes
| Source        | Target        | Data         |
| ------------- | ------------- |:------------:|
| alice         | boxy          | Owes(9999)   |

(both of these are external to the components/resources/archetypes)

I'm not familiar with the code but I think it might require a bit of work to make `Query<(Entity, &Relation<FriendsWith>)>` work here but I think it's doable?

BoxyUwU:
> Why is this the case? Relations could build on SparseSet - I guess the question is, now does every entity have a Relation?

@tigregalis relations can be in any `StorageType` just like components- regardless of Table/SparseSet storage this still fragments archetypes as in bevy "archetypes" means something subtly different than other ECS'- in bevy archetypes are just a pile of metadata for what tables and sparse sets all the components are stored in for that set of entities

It's still bad to fragment our archetypes here though as non-dense queries (queries that interact with SparseSet stored components/relations) have to iterate over archetypes so keep the amount of archetypes low is good for performance.

> (both of these are external to the components/resources/archetypes)

Storing relations in such a way that they dont fragment archetypes is something I would very much like to support- however, fragmenting archetypes isnt *always* bad. I [wrote up some thoughts](https://hackmd.io/taA-uPxJTyGXcOIgjFNpHQ?view) a few days ago about the problems with fragmenting archetypes and also why we probably want to support doing it

rfcs/min-relations.md
```diff
@@ -0,0 +1,644 @@
...
+
+## Motivation
+
+Entities often need to be aware of each other in complex ways: an attack that targets another unit, groups of entities that are controlled en-masse, or complex parent-child hierarchies that need to move together.
+
+We *could* simply store an `Entity` in a component, then use `query.get(my_entity)` to access the relevant data.
+(In this RFC, we call this the **Entity-in-component pattern**.)
+But this quickly starts to balloon in complexity.
+
+- How do we ensure that these components are cleaned up when the entity they're pointing to is?
```
aevyrie, line 18:
```suggestion
- How do we ensure that these components are cleaned up when the entity they're pointing to is removed?
```

rfcs/min-relations.md
```diff
@@ -0,0 +1,644 @@
...
+### Why is my favorite feature relegated to Future Work?
+
+It's already many thousands of words long, and the associated PR has thousands of lines of code.
+Done is better than perfect, and we had to draw the line somewhere.
+As with all things, we can gradually improve this over time.
+
+The set of features expressed here is useful enough to be a clear improvement over the existing patterns in at least some use cases.
+There's a lot of value in getting this feature in front of users to see the response at scale.
+
+### Why is the implementation for this so *slow*?
```
aevyrie, line 570: It could be helpful to provide some measure of how slow this is, and what operations are particularly bad.

- Are queries fast? 
- Is it only adding/changing relations that is slow?

A starting point could be to run the ECS benchmark as-is with some relation filters on to see if there is a tangible impact. Perhaps seeing how adding/removing relations compares to component add/removes. Having *some* information, even if it isn't complete or perfect, would help users understand the order of magnitude they should expect.

> Many uses of relations are not performance-critical,

I only say this because "performance-critical" is a hugely relative metric.

rfcs/min-relations.md
aevyrie: `SharedRelation` seems equivalent in meaning to my untrained eyes (co -> shared) and seems to match the way I would describe this if I was teaching it to someone.

rfcs/min-relations.md
```diff
@@ -0,0 +1,644 @@
...
+7. `Noitaler`*, for relations that point in the opposite direction.
+8. Streaming iters to allow for queries like: `Query<&mut Money, Relation<Owes, &mut Money>>` and potentially use invariants on graph shape to allow for skipping soundness checks for the aliasing `&mut Money`'s
+9. Assorted performance optimizations. For example:
+   1. Reducing the cost of having many archetypes.
+   2. Relation storage type options to cause archetype fragmentation based on whether *any* relations of that type are present.
+   3. Index-backed relations look-up.
+10. Relative ordering between relations of the same kind on the same entity.
+This would enable the `Styles` proposal from #1 to use relations.
+11. Compound relation filters, letting you nest logic arbitrarily deep. These are just sugar / perf for very advanced users, and can wait for a while.
+12. \[Controversial\] A full graph constraint solver DSL ala [Flecs](https://github.com/SanderMertens/flecs) for advanced querying.
```
alice-i-cecile, line 634:
```suggestion
12. \[Controversial\] A full graph constraint solver DSL ala [Flecs](https://github.com/SanderMertens/flecs) for advanced querying.
13. Entity-to-entity event channels using relations (blocked on [bevy #2116](https://github.com/bevyengine/bevy/pull/2116)
```

rfcs/min-relations.md
```diff
@@ -0,0 +1,645 @@
...
+and a tangible implementation to benchmark against.
+There's also nothing stopping you from doing things the old ways in your elaborate graph-traversing systems.
+
+See also: the question directly above.
+
+## Prior art
+
+[`Flecs`](https://github.com/SanderMertens/flecs), an advanced C++ ECS framework, has a similar feature, which they call "relationships".
+These are somewhat different, they use an elaborate query domain-specific language along with being more feature rich.
+You read more about them in the [corresponding PR](https://github.com/SanderMertens/flecs/pull/358).
```
alice-i-cecile, line 585:
```suggestion
You can read more about them in the [docs](https://github.com/SanderMertens/flecs/blob/master/docs/Queries.md) or [corresponding PR](https://github.com/SanderMertens/flecs/pull/358).
```

alice-i-cecile:
Status update! A [branch](https://github.com/bevyengine/bevy/tree/entity_relations) on the Bevy repo has been opened up for this work. Plan is:

1. Do basic review on https://github.com/bevyengine/bevy/pull/1627. 
2. Merge in that PR so others have a stable base to work off of.
3. Put `bevy_ecs` in "slow mode" to reduce the level of merge conflicts we get while this is live.
4. Work on that branch as a group to benchmark relations, improve performance and add features before they go live.

From @cart on [Discord](https://discord.com/channels/691052431525675048/692572690833473578/845093291824775219), his main concerns were:

> 1. I want to see perf comparisons between "manually implemented reactive indexing" (which represents the theoretical upper limit to a real reactive system) and relations, with benches that cover the spectrum of best and worst case performance. Ideally we have some "real world" scenario tests too.
> 2. I want a summary of those tests in the context of "how we would report the current perf limitations to users".
> 
> We can then use these to determine if we merge the current relations impl (or what the criteria for merging will be). If relations are the general purpose indexing api we're committing to, but the current impl doesn't meet those needs yet, we need to come up with a concrete plan for the future with a very high chance of success and few unknowns (edit for clarity:  ... if we want to merge the current impl). 

This work should be able to be distributed nicely once we have a basic implementation in place on that branch

SanderMertens:
FYI, documentation on flecs relations:
https://github.com/SanderMertens/flecs/blob/master/docs/Relations.md

Relevant sections from the query manual on relations:
https://github.com/SanderMertens/flecs/blob/master/docs/Queries.md#relation
https://github.com/SanderMertens/flecs/blob/master/docs/Queries.md#wildcard
https://github.com/SanderMertens/flecs/blob/master/docs/Queries.md#variables
https://github.com/SanderMertens/flecs/blob/master/docs/Queries.md#set-substitution
https://github.com/SanderMertens/flecs/blob/master/docs/Queries.md#transitivity

LegNeato:
Facebook calls these "associations" in TAO (their graph db) and their data access application code layer FYI:

(Pdf warning)

https://www.usenix.org/system/files/conference/atc13/atc13-bronson.pdf&ved=2ahUKEwif9qnSyYzyAhVFgp4KHYhIAqEQFjABegQICRAC&usg=AOvVaw3IBhg6Xn4t8igN25axxu-h

LegNeato:
Also check out differential dataflow, which makes storing and operating on graphs extremely fast and performant:

https://github.com/TimelyDataflow/differential-dataflow

alice-i-cecile:
I've been reading about differential dataflow :) Seems like an advanced high-performance abstraction we can add on later and use internally. It will likely be too complex for most end user cases, but I'm curious to see what sort of high-performance APIs for common operations we can create.

LegNeato:
As I used to work at FB...all I think about and have experience with are graphs :-). Happy to talk about what worked and what didn't at scale. I think exposing graph semantics at extremely powerful and let framework authors optimize a lot while keeping a simple interface for product code (mainly entities and associations / edges and how to traverse them).

HackerFoo:
I'm using relations in my [CSG editor](https://youtu.be/4sPx02P_JPs) to represent operations on two (or more) shapes. For this, binary relations (graph edges) are not enough, because there are (at least) three entities involved: the two operands, and the result.
It also useful to iterate through all relations of a specific type (for computing the operations), which implies that each relation should be an entity.

This is how relations are represented in relational databases, where each relation is a table.

Here's a [gist](https://gist.github.com/HackerFoo/24ff001d217d1efad1944eb08c9dd916), but much more work would be required to make this easy to use. Note that this uses a `Component` containing nothing but `Entity`s, so it's the opposite approach :)

Since each relation is an entity, this supports higher-order relations, as well as using an existing entity as a relation, e.g. parent/child relations could be represented with the RelationIndex pointing to the inverse relation, so that a child's `RelationIndex` points to the parents Children relation, containing the child's entity.

vultix:
@alice-i-cecile What is the current plan for this feature set?

alice-i-cecile:
@vultix good question! There are two competing paths forward, very roughly:

1. Implement some form of reliable [indexing](https://github.com/bevyengine/bevy/discussions/1205). At the moment I expect that atomic ordering constraints are probably the most likely strategy here.
2. Build out a very basic prototype that uses an resource to track the graph centrally and the reliable indexing tech to keep this in sync.
3. Build out a pretty API on top of that prototype, which should be able to mostly look like the design in this RFC.

Alternatively:
1. Forbid direction mutations of parent-child components using type-level permissions.
2. Only allow modification via commands.
3. Extend the parent / child commands API to be generic over user-generated relation types.

The first is my preference, but the second may be easier to get to (and the rest of the team can disagree with me!). The gist of the problem is that we can't efficiently store lookups of the relations graph within the archetype graph, because of the large non-localized performance implications of such widespread archetype fragmentation.

Most of the relevant ECS folks are tied up with trying to [improve scheduling APIs](https://github.com/bevyengine/bevy/discussions/2801) so this feature has been on the backburner.

PR #21, opened 2021-04-25
Improve RFC template

alice-i-cecile:
[RENDERED](https://github.com/colepoirier/rfcs/blob/rfc-section-update/template.md)

 I've worked with the RFC process a fair bit now, and while I generally like it, there are a few obvious warts that I'd like to get out of the way sooner than later.
 
 The basic problems I'm trying to solve are:

1. Guide-level and Reference-level are nonsensical and unclear as section names in the context of Bevy. These have been changed to User-facing Explanation and Implementation Strategy
2. Implementation Strategy's guidance was vague and fuzzy, and conflicted with Cart's directives for very detailed designs, as given in the README
3. Work done in RFCs is not easily translated to maintainable bits of Bevy. I've added a dedicated Examples section to discuss what should be in the examples folder, and pushed to make the User-Facing Explanation more portable to the Bevy book. 

This means that there's a rough mapping of sections onto durable work:
1. User-facing explanation: chopped up and moved to the Bevy book or other documentation.
2. Examples: used to improve the doc examples and the examples folder.
3. Implementation strategy: useful for making and reviewing the implementation PR.
4. Future work: moved to tracking issues on the main Bevy repo so valuable ideas aren't lost forever.

Everything else is ephemeral; preserved only to record why we made the decisions we did.

### DIFF
--- a/template.md
+++ b/template.md
@@ -20,7 +20,26 @@ Explain the proposal as if it was already included in the engine and you were te
 
 ## \[Optional\] Examples
 
-This is the technical portion of the RFC. Explain the design in sufficient detail that:
+Lay out the examples that Bevy should add to either the examples folders or the doc comments to help new users discover and learn about this feature.
+These can either be completely new, or simply changes to existing examples.
+When completing this section, you can but do not have to write all of these examples yourself.
+Describing them in sufficient detail that another developer can do so is perfectly fine.
+If your examples are particularly long, please provide a brief description and link to a [Gist](https://gist.github.com/) for each.
+
+Most examples in Bevy aim to clearly demonstrate a single feature, group of closely related small features,
+or show how to accomplish a particular task (such as asset loading, creating a custom shader or how to test your app).
+In rare cases, creating new "game" examples is justified to demonstrate new features that open a complex class of functionality in a way that's hard to showcase in isolation or requires additional integration testing.
+
+Examples in Bevy should be:
+
+1. Working. They must compile and run, and should fail in obvious ways.
+2. Clear. They must use descriptive variable names, have reasonable code-quality, be formatted, and be appropriately commented.
+3. Relevant. They should use game-relevant fluff and explain why what they're demonstrating is useful.
+4. Minimal. They should be no larger or more complex than is needed to meet their other goals.
+This reduces maintenance burden and improves clarity when used as a reference.
+
+More complex demonstrations of functionality are also welcome, but for now belong in community tutorials or template games.
+
 ## Implementation strategy
 
 This is the technical portion of the RFC.
--- a/template.md
+++ b/template.md
@@ -28,7 +28,7 @@ and then focus in on the tricky details so that:
 - It is reasonably clear how the feature would be implemented.
 - Corner cases are dissected by example.
 
-When necessary, this section should return to the examples given in the previous section, and explain more the details that make those examples work.
+When necessary, this section should return to the examples given in the previous section and explain the implementation details that make them work.
 
 When writing this section be mindful of the following [repo guidelines](https://github.com/bevyengine/rfcs):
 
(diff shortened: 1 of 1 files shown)


template.md
```diff
@@ -18,15 +18,45 @@ Explain the proposal as if it was already included in the engine and you were te
...
 
-## Reference-level explanation
+## \[Optional\] Examples
 
-This is the technical portion of the RFC. Explain the design in sufficient detail that:
+Lay out the examples that Bevy should add to either the examples folders or the doc comments to help new users discover and learn about this feature.
+These can either be completely new, or simply changes to existing examples.
+When completing this section, you can but do not have to write all of these examples yourself.
+Describing them in sufficient detail that another developer can do so is perfectly fine.
+If your examples are particularly long, please provide a brief description and link to a [Gist](https://gist.github.com/) for each.
```
Nilirad, line 27: What about suggesting to wrap the code into a collapsible? A gist could be accidentally deleted and difficult to track and be kept in sync with the RFC.

alice-i-cecile: I agree with this; I'd forgotten about this functionality! This looks to be a pretty good link that explains how to do so: https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab

alice-i-cecile, line 27:
```suggestion
If your examples are particularly long, please use a [collapsible section](https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab) with a summary for each example.
```

template.md
```diff
@@ -18,15 +18,45 @@ Explain the proposal as if it was already included in the engine and you were te
 - If applicable, provide sample error messages, deprecation warnings, or migration guidance.
 - If applicable, explain how this feature compares to similar existing features, and in what situations the user would use each one.
 
-## Reference-level explanation
+## \[Optional\] Examples
```
cart, line 21: I'm not sure we should encourage RFC developers to come up with complete self-contained examples (on top of the author-selected example snippets provided in the user-facing explanation):

1. Coming up with a compilable example without an implementation of the feature to run through a compiler is a big ask
2. Choosing the best holistic example to teach a given feature isn't necessarily the same as picking the best code snippets to introduce a feature in an RFC.
3. Full examples do take up space and create new logistics problems like requiring collapsible sections. 

To me this feels a bit like "additional" work that often can't be efficiently completed without a working impl. I'm pretty cool with just asking RFC implementers to come up with examples (when they are relevant).

cart:
In general I dig these changes. I'm just a little uncertain about the "examples" section (see my comment above).

template.md
alice-i-cecile: I agree with 1 and 2 for sure, and 3 can be a bit tricky. This is why I marked it as Optional; I think it's not a great fit for many RFCs but critical for others.

In particular, this was inspired by the UI work, where complex vertical slices are really important for seeing how everything fits together. I think this applies in a number of other domains too (e.g. physics, subworlds or scheduler APIs) where a lot of the challenge in a good design is in complexity management. The idea was to structure those slices by relating them to work that Bevy itself can use directly so then it doesn't go to waste.

People can always add bonus sections I guess, but I figured a cue and some concrete advice might be helpful for when it applies :)

Regardless on what we decide, I think the advice in this section on "What makes a good Bevy example" is valuable and if it doesn't end up here we should drop it in CONTRIBUTING.md or the like :)

cart: I think I'd prefer to just let this line from the "User-facing explanation" section to be open-ended based on the scope of the rfc: "Explaining the feature, ideally through simple examples of solutions to concrete problems."

Something like a UI RFC without UI examples isn't very useful, so we can/should encourage people to provide more examples where relevant.

My preference is to just remove the example section here and add the "what makes a good Bevy example" section to the Contributing section of the Bevy Book. I want to keep the RFC template "lean", and a large section requesting complete compilable / Bevy-example-folder-ready examples feels a bit "heavy", even if it has the optional label.

alice-i-cecile: Done. I think this should be ready to merge now then @cart.

PR #27, opened 2021-06-15
Require `#[derive(Component)]` for component types

Frizi:
[Rendered](https://github.com/Frizi/rfcs/blob/derive-component/rfcs/27-derive-component.md)

[Early discussion issue](https://github.com/bevyengine/bevy/issues/1843)
[Implementation PR](https://github.com/bevyengine/bevy/pull/2254)

### DIFF
--- a/rfcs/27-derive-component.md
+++ b/rfcs/27-derive-component.md
@@ -86,14 +86,6 @@ Remove all registration of components. Make sure to correctly annotate component
   struct MyComponent { .. }
 ```
 
-Explain the proposal as if it was already included in the engine and you were teaching it to another Bevy user. That generally means:
-
-- Introducing new named concepts.
-- Explaining the feature, ideally through simple examples of solutions to concrete problems.
-- Explaining how Bevy users should *think* about the feature, and how it should impact the way they use Bevy. It should explain the impact as concretely as possible.
-- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
-- If applicable, explain how this feature compares to similar existing features, and in what situations the user would use each one.
-
 ## Implementation strategy
 
 - remove the existing blanket impl
--- a/rfcs/27-derive-component.md
+++ b/rfcs/27-derive-component.md
@@ -7,15 +7,15 @@ Extend the `Component` trait to declare more information statically, instead of 
 
 ## Motivation
 
-`Component` trait is being used inside Bundles and Queries. It's usage in Queries is enforced by the type system.
+The `Component` trait is being used inside Bundles and Queries. It's usage in Queries is enforced by the type system.
 Unfortunately, because it's automatically derived, almost every type can be treated as a component, including Bundles.
 That means `Component` APIs don't really have any way to prevent misuse, like `commands.insert(bundle)`. Right now
 it's very easy to end up with that code because of a typo, as the `Bundle` api methods are very similarly named.
 
 There are also other pitfalls, like primitive types being automatically `Component`s, but without clear meaning what they represent.
-This is an especially sever issue for function pointer types. It's easy to use an enum variant or newtype constructor as
-an component, like `.insert(MyEnum::Variant)`,  without realising it. This leads to to very hard to spot bugs without any hint
-of either a compiler or runtime that something went wrong. This is easily preventable by not implementing `Component` trait for those types.
+This is an especially severe issue for function pointer types. It's easy to use an enum variant or newtype constructor as
+a component, like `.insert(MyEnum::Variant)`,  without realising it. This leads to very hard to spot bugs without any hint
+by the compiler or at runtime that something went wrong. This is easily preventable by not implementing the `Component` trait for those types.
 
 We also already have `#[derive(Bundle)]` and others. Adding that to `Component` keeps the syntax similar across the board.
 
@@ -29,11 +29,11 @@ into the world.
 struct MyComponent(..);
 ```
 
-This enables compiler to perform much more optimizations in the query iteration code, as a lot of the code paths can be statically eliminated.
+This enables the compiler to perform much more optimizations in the query iteration code, as a lot of the code paths can be statically eliminated.
 
 ## User-facing explanation
 
-In order to define your own component type, you have to implement a `Component` trait. The easiest way to do that is using a `derive` macro.
+In order to define your own component type, you have to implement the `Component` trait. The easiest way to do that is using a `derive` macro.
 
 ```rust
 #[derive(Component)]
@@ -41,7 +41,7 @@ struct HitPoints(u32);
 ```
 
 If you forget to properly annotate your component type, the compiler will remind you
-of that as soon as you try to insert that component into the world or query for it. Apart from type safety, this also provides a visual indication that given type
+of that as soon as you try to insert that component into the world or query for it. Apart from type safety, this also provides a visual indication that the given type
 is intended to be directly associated with entities in the world.
 
 By default, your component will be stored using Dense storage strategy.
--- a/rfcs/27-derive-component.md
+++ b/rfcs/27-derive-component.md
@@ -104,7 +104,29 @@ Requiring extra annotation on every component type adds some boilerplate.
 - instead of `derive`, use an attribute macro.
 - Do nothing and keep registering metadata at runtime. Live with the branching inside query iterators.
 - Require manual implementation of `Component` trait without providing derive macro.
+### Why not use a manual implementation of `Component` to set the storage type?
 
+As shown in [this playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ca85c5d3c561ef9d1feec143bbcd9611), we could force users to write out a manual implementation of the `Component` trait when they wanted to change the storage type.
+
+This has a couple advantages:
+
+1. More direct and less magical, at least for advanced Rust users.
+2. No configuration DSL to maintain.
+3. Better type checking.
+
+However, it has several more serious disadvantages, which outweigh those:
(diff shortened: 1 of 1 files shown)


rfcs/27-derive-component.md
```diff
@@ -0,0 +1,123 @@
+# Feature Name: `derive_component`
+
+## Summary
+
+Remove blanket impl from `Component` trait, and require it to be manually implemented or derived using `#[derive(Component)]`.
+Extend the `Component` trait to declare more information statically, instead of requiring runtime registration.
+
+## Motivation
+
+`Component` trait is being used inside Bundles and Queries. It's usage in Queries is enforced by the type system.
```
MinerSebas, line 10:
```suggestion
The `Component` trait is being used inside Bundles and Queries. It's usage in Queries is enforced by the type system.
```

rfcs/27-derive-component.md
```diff
@@ -0,0 +1,123 @@
...
+- world.register_component(
+-     ComponentDescriptor::new::<MyComponent>(StorageType::SparseSet)
+- ).unwrap();
+
++ #[derive(Component)]
++ #[component(storage = "SparseSet")]
+  struct MyComponent { .. }
+```
+
+Explain the proposal as if it was already included in the engine and you were teaching it to another Bevy user. That generally means:
```
BoxyUwU, line 89: did you mean to keep this in? :sweat_smile:

rfcs/27-derive-component.md
```diff
@@ -0,0 +1,123 @@
...
+
+`Component` trait is being used inside Bundles and Queries. It's usage in Queries is enforced by the type system.
+Unfortunately, because it's automatically derived, almost every type can be treated as a component, including Bundles.
+That means `Component` APIs don't really have any way to prevent misuse, like `commands.insert(bundle)`. Right now
+it's very easy to end up with that code because of a typo, as the `Bundle` api methods are very similarly named.
+
+There are also other pitfalls, like primitive types being automatically `Component`s, but without clear meaning what they represent.
+This is an especially sever issue for function pointer types. It's easy to use an enum variant or newtype constructor as
+an component, like `.insert(MyEnum::Variant)`,  without realising it. This leads to to very hard to spot bugs without any hint
+of either a compiler or runtime that something went wrong. This is easily preventable by not implementing `Component` trait for those types.
```
MinerSebas, line 18:
```suggestion
This is an especially severe issue for function pointer types. It's easy to use an enum variant or newtype constructor as
a component, like `.insert(MyEnum::Variant)`,  without realising it. This leads to very hard to spot bugs without any hint
of either by the compiler or at runtime that something went wrong. This is easily preventable by not implementing the `Component` trait for those types.
```

rfcs/27-derive-component.md
```diff
@@ -0,0 +1,123 @@
...
+One way to use that is to declare component's Storage type statically, instead of requiring manual registration of that metadata
+into the world.
+
+```rust
+#[derive(Component)]
+#[component(storage = "SparseSet")]
+struct MyComponent(..);
+```
+
+This enables compiler to perform much more optimizations in the query iteration code, as a lot of the code paths can be statically eliminated.
```
MinerSebas, line 32:
```suggestion
This enables the compiler to perform much more optimizations in the query iteration code, as a lot of the code paths can be statically eliminated.
```

rfcs/27-derive-component.md
```diff
@@ -0,0 +1,123 @@
...
+
+## User-facing explanation
+
+In order to define your own component type, you have to implement a `Component` trait. The easiest way to do that is using a `derive` macro.
```
MinerSebas, line 36:
```suggestion
In order to define your own component type, you have to implement the `Component` trait. The easiest way to do that is using a `derive` macro.
```

rfcs/27-derive-component.md
```diff
@@ -0,0 +1,123 @@
...
+
+```rust
+#[derive(Component)]
+struct HitPoints(u32);
+```
+
+If you forget to properly annotate your component type, the compiler will remind you
+of that as soon as you try to insert that component into the world or query for it. Apart from type safety, this also provides a visual indication that given type
```
MinerSebas, line 44:
```suggestion
of that as soon as you try to insert that component into the world or query for it. Apart from type safety, this also provides a visual indication that the given type
```

rfcs/27-derive-component.md
Frizi: No :)

cart:
I'd like it if the RFC compared "api design choices" a bit more:

* [Derive DSLs vs manual impls](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ca85c5d3c561ef9d1feec143bbcd9611) 
* [Methods of defining storage type in Component trait](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4a2d47173f68137a34c31987d9d46bfa)

alice-i-cecile (approved):
This is compelling enough for me as is, and I agree with everything stated in it.

Ixentus:
I agree with everything in this RFC about `#[derive(Component)]`. It offers a ton of benefits.

I don't like `#[component(storage = "SparseSet")]`. `#[derive(Component)]` is alright since most Components have derives anyway, but this storage one doesn't seem necessary. It looks like cart has some alternatives for this. Perhaps this change could be removed from the RFC and moved to a new issue, so the RFC can be accepted easily.

alice-i-cecile:
@Ixentus the new component storage configuration approach is much nicer than the old strategy. It will remain an optional optimization; simply omitting any configuration will use the default table storage.

wilk10:
Coming from a user perspective, I'd like to add that I really don't mind the (limited) extra boilerplate. More than once I found myself scanning through my structs wondering: "Is this a Component?", so besides the technical benefits listed above, I would enjoy the (already mentioned) visual indication at first glance as well.

NathanSWard:
> More than once I found myself scanning through my structs wondering: "Is this a Component?

I agree. This is one of the big upside to this RFC that is also really like.

Something else, I would like a world where change detection has an opt-out path, but that'll only be possibly once we have `#[derive(Component)]` and `#[derive(Resource)]`

BoxyUwU (approved):
I'm not sold on `#[component(storage = "SparseSet")]` but other than that I quite like this, personally I'd consider not having `derive(Component)` to be a blocker for relations

NathanSWard:
> I'm not sold on `#[component(storage = "SparseSet")] `

@BoxyUwU I'm curious about what part of this you're not sold on. Is it the static storage type, the syntax, etc...?

BoxyUwU:
ah sorry yes should have clarified, it's just the syntax yeah. not a fan of using a string here and it's also a lot of text to write out. I'm not familiar with proc macros though so have no idea how actionable that is. Either way the proposed syntax seems totally workable and I wouldn't hate writing it :)

Frizi:
I actually tried without the string, but that would be kinda non-standard. The [attribute parser provided with syn crate](https://docs.rs/syn/1.0.73/syn/struct.Attribute.html#method.parse_meta) only allows string literal at that position. We could roll our own if that's really desired, but it's just more work and definitely less common pattern.

rfcs/27-derive-component.md
```diff
@@ -0,0 +1,115 @@
...
+## Drawbacks
+
+Requiring extra annotation on every component type adds some boilerplate.
+
+## Rationale and alternatives
+
+- instead of `derive`, use an attribute macro.
+- Do nothing and keep registering metadata at runtime. Live with the branching inside query iterators.
+- Require manual implementation of `Component` trait without providing derive macro.
+
```
alice-i-cecile, line 107:
```suggestion
### Why not use a manual implementation of `Component` to set the storage type?

As shown in [this playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ca85c5d3c561ef9d1feec143bbcd9611), we could force users to write out a manual implementation of the `Component` trait when they wanted to change the storage type.

This has a couple advantages:

1. More direct and less magical, at least for advanced Rust users.
2. No configuration DSL to maintain.
3. Better type checking.

However, it has several more serious disadvantages, which outweigh those:

1. Manual implementations of `Component` will break (often in dozens of places scattered around the code base) whenever we make breaking changes to the `Component` trait. Derive + attribute macros will continue working flawlessly, except where they directly touch the breaking changes.
2. Longer boilerplate, especially as the `Component` trait grows. This is particularly frustrating as you need to add / remove this repeatedly when benchmarking perf as the actual net impact is very fact-specific.
3. Exposes internal-ish implementation details to users who really don't care.

### Comparison of methods of defining storage type in `Component` trait

There are several possible ways to define the storage type, as shown in [this playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4a2d47173f68137a34c31987d9d46bfa).

While any approach would work, an associated type is safer and offers better discoverability, due to the use of the type checker.
In addition, you can't assert specific associated const value in the trait bounds, which may be limiting in the future.
The slight increase in verbosity is worth it for these benefits.
```

I didn't want to deal with a fork-of-a-fork of a repo I already have far too many copies of 😅

Ixentus:
Thanks for that write-up @alice-i-cecile , I am now much happier with this DSL macro.

alice-i-cecile (approved):

cart:
I think this is probably good to go. I do wish this was a bit more prescriptive about implementation details (ex: what should the Storage associated type look like? How will that be consumed in things like Fetch impls? Will Component::Storage be constrained to a Storage trait?)

But i've also dragged my feet on this enough and we do have an implementation that we all like.

PR #28, opened 2021-06-22
CPU Particles RFC

james7132:
[RENDERED](https://github.com/james7132/rfcs/blob/cpu-particles/rfcs/cpu-particles.md)

This RFC proposes adding a fully featured CPU particle system implementation to Bevy.

### DIFF
--- a/rfcs/cpu-particles.md
+++ b/rfcs/cpu-particles.md
@@ -1,0 +1,398 @@
+# Feature Name: `cpu-particles`
+
+Prototype Implementation: https://github.com/james7132/bevy_prototype_particles
+
+## Summary
+Provide a modular set of components and systems for authoring and controlling
+CPU-simulated particle systems.
+
+## Motivation
+Games frequently require visual effects that are more fluid in nature, like fire,
+lightining, or smoke. These effects are difficult, or impossible to simulate with
+static or skinned meshes.
+
+This RFC aims to detail an efficient CPU-bound implementation for modularly
+creating, simulating, and rendering these kinds of effects.
+
+## User-facing explanation
+Each particle system is a collection of simple billboarded quads with the following
+additional properties: position, 2D rotation, velocity, angular velocity, color,
+texture UVs, and lifetime. When spawned, a particle is assigned a lifetime, which
+ticks down frame by frame, until it's destroyed when it reaches 0.
+
+Particles themselves are not entities, but are instead contained within a
+`Particles` component, which stores the live state of all particles in a given
+particle system as the general configuration for particle initialization.
+Particles do not have a stable ID and attempting to store one is meaningless.
+
+`ParticleMaterial` is a material component for rendering each particle. Main hook
+for into the rest of the render graph. Adding this to a entity with `Particles`
+will render the inner state of `Particles`. In this case, `Particles` does the
+same job that a `Handle<Mesh>` component does in the normal rendering flows.
+
+`ParticleEmitter` is a component for controlling the emission behavior of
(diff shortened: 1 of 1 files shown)


JonahHenriksson:
Hey, I was really interested in creating something like this as well, but after looking at your code it seems like Bevy is going through a lot of changes with the render system, so I think I'm going to wait a little while until it's more stable. But I have a few ideas for how users can control values for things like position, 2D rotation, color, etc. So I figured you might be interested, especially if you are going to be the driving force behind this (I'm not too involved, so I don't know about who is doing what, but you seem way more experienced with this stuff). Essentially, it's a builder pattern for interpolation curves. It would look like so:

```
InterpolationCurve::build()
  .mode(mode: InterpolationMode::Linear)
  .push(value: 0.4, length: 20)
  .push(value: 0.2..0.3, left: 10, right: 20)
  .mode(mode: InterpolationMode::Cardinal)
  .push(value: 0.6, length 30);
```
The idea is that the lengths are relative (they don't have to add to 1); 20 will become 20/(20+30+30)=0.25. This could work with things like your `CurveVariable::<Color>` to control how long it takes to mix colors. Lengths can either be the midpoint (just `length`) or be represented by different parts for each side (`left` and `right`). `mode` can be used to change the `InterpolationMode` for that moment onward. This doesn't have to be just for particles, I'm sure it would work well for animation.

Also, I think `ParticleEmitter` should be a bit more generic. For instance, in my design I used a trait, so users can create their own emitters, in case they want to create patterns and what not. I know that your code is a prototype, so those built in functions could be placeholders, but who knows! I'm coming from seeing other game engines, like Unity, feel kind of bloated, when the tools aren't flexible enough for tasks that go outside of what it was made for. Kudos!

james7132:
> Essentially, it's a builder pattern for interpolation curves. It would look like so:

Seems good, the referenced [Curve PR in the RFC](https://github.com/bevyengine/bevy/pull/1837) could really use a better builder-like public interface. Though I think the eventual goal is to have it available in some general editor eventually instead of authoring it from code.

> Also, I think ParticleEmitter should be a bit more generic. For instance, in my design I used a trait, so users can create their own emitters, in case they want to create patterns and what not. I know that your code is a prototype, so those built in functions could be placeholders, but who knows! I'm coming from seeing other game engines, like Unity, feel kind of bloated, when the tools aren't flexible enough for tasks that go outside of what it was made for. Kudos!

This is actually probably better to just keep it simple. `Particles` already has a public interface for spawning particles via `ParticleParams`. Nothing is stopping a developer from writing their own emitters that does not rely on the existing `ParticleEmitter` infrastructure.

djeedai:
This very much look like the kind of feature that can be entirely investigated and implemented in an external crate without an RFC. Case in point is `bevy_hanabi` for GPU particles (I'm the author). What is the rationale for trying to add this as built-in @james7132 ?

james7132:
Yes, this can be experimented with externally, and I am via https://github.com/james7132/bevy_prototype_particles (though that is slowly getting out of date with the rendering implementation), but ultimately I want this to be a first party crate at some point. The rationale is in the RFC: no competitive game engine is considered feature complete without some kind of implementation for particle systems. Also having a common implementation/interface allows the ecosystem to develop tools against it more readily (i.e.  CPU-side modifiers). IMO, we need both a first party CPU and a GPU implementation to round out the rendering offering that Bevy has to be competitive with pretty much any other major game engine.

cBournhonesque:
Just mentioning that this crate: https://github.com/abnormalbrain/bevy_particle_systems
is a good example of CPU particles.
(however particles are spawned as separate entities)

PR #31, opened 2021-09-02
System builder syntax rework

thechubbypanda:
Existing PR: [#2736](https://github.com/bevyengine/bevy/pull/2736)
Original issue: [#1978](https://github.com/bevyengine/bevy/issues/1978)

### DIFF
--- a/rfcs/2736-system_builder_syntax.md
+++ b/rfcs/2736-system_builder_syntax.md
@@ -1,0 +1,72 @@
+# Feature Name: (fill me in with a unique ident, `system_builder_syntax`)
+
+## Summary
+
+Rewrite the system creation workflow to remove the excess of `add_startup_system_to_stage`-like methods and instead use the `system.stage().label().startup()` builder-like system.
+
+## Motivation
+
+Mostly described in [#1978](https://github.com/bevyengine/bevy/issues/1978).
+
+- Allows us to unify the existing `.label()`, `.after()`, etc. methods and the stage positioning of systems.
+- App would only have 3 methods to add systems: `add_system`, `add_exclusive` and `add_system_set`.
+- Arguably more intuitive design (somewhat backed up by those who commented on #1978).
+
+## User-facing explanation
+
+- `add_startup_system(system)` becomes `add_system(system.startup())`
+- `add_system_to_stage(stage, system)` becomes `add_system(system.stage(stage))`
+- `add_startup_system_to_stage(stage, system)` becomes `add_system(system.startup().stage(stage))`
+- `add_system(system.exclusive_system())` becomes `add_exclusive(system.exclusive_system())`
+- `add_startup_system(system.exclusive_system())` becomes `add_exclusive(system.exclusive_system().startup())`
+- `add_startup_system(system.label(x).before(y))` becomes `add_system(system.startup().label(x).before(y))`
+
+The rest should follow.
+
+## Implementation strategy
+
+This has already been implemented in pr [#2736](https://github.com/bevyengine/bevy/pull/2736).
+
+Similar to what was suggested by @cart in #1978:
+
+```
+trait System {
+  /* same functions as before */
+  fn config(&mut self) -> &mut SystemConfig;
+}
+
+trait LabelSystem {
+  fn label(self, label: impl SystemLabel) -> Self
+}
+
+impl<T: System> LabelSystem for T {
+  fn label(self, label: impl SystemLabel) -> Self {
+    self.config().add_label(label);
+    self
+  }
+}
+```
+
+- Each system impl stores a `struct SystemConfig` which stores data such as stage labels, run criteria, insertion points, etc.
+- A trait is created for each type of Config(urability) such as a `trait StageConfig` which has a `stage(StageLabel)` function that sets the given StageLabel in the system's config struct.
+- All config types are added to the prelude so that no more imports are required in the examples at the very least.
+- All legacy system adding functions are removed and replaced with the basic ones described in the motivation. They will handle everything the same as the original ones did but in fewer functions.
+
+## Drawbacks
+
+- Requires changes over a large surface area.
+- Changes the way end users interact with the API, requires a (short) adjustment period.
+- Removes the current system descriptor system (existing prs may need reworking)
+- Adds a minor memory overhead to each system.
+- Minor boilerplate increase for each system impl.
+- "Forces systems to own "schedule config", which muddles the "dependency tree" a bit. Systems currently exist a level "below" shcedules, which means other schedule apis could theoretically be built on top. This "mixes" the levels a bit." - @cart #1978
+
+## Rationale and alternatives
+
+- Alternative: Somehow write this on top of the existing system descriptors.
+    - Would be signifcantly less "clean", where clean is a mix of line count and code "layers".
+- Not implementing this rfc would likely hinder future users and devs of bevy in the long run since this improvement is more modular and can be built upon easily.
+
+## Unresolved questions
+
+- Are we happy to fundamentally alter the end user experience of Bevy and basically make all existing docs out of date.
--- a/rfcs/2736-system_builder_syntax.md
+++ b/rfcs/2736-system_builder_syntax.md
@@ -20,6 +20,10 @@ Mostly described in [#1978](https://github.com/bevyengine/bevy/issues/1978).
 - `add_system(system.exclusive_system())` becomes `add_exclusive(system.exclusive_system())`
 - `add_startup_system(system.exclusive_system())` becomes `add_exclusive(system.exclusive_system().startup())`
 - `add_startup_system(system.label(x).before(y))` becomes `add_system(system.startup().label(x).before(y))`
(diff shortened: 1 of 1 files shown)


Ratysz:
[Rendered](https://github.com/bevyengine/rfcs/blob/9623eaa041aba2c8e227479166dcf71080753a21/rfcs/2736-system_builder_syntax.md).

alice-i-cecile:
> Are we happy to fundamentally alter the end user experience of Bevy and basically make all existing docs out of date.

Yep! This change is only going to get more painful the longer we wait.

rfcs/2736-system_builder_syntax.md
```diff
@@ -0,0 +1,72 @@
...
+## User-facing explanation
+
+- `add_startup_system(system)` becomes `add_system(system.startup())`
+- `add_system_to_stage(stage, system)` becomes `add_system(system.stage(stage))`
+- `add_startup_system_to_stage(stage, system)` becomes `add_system(system.startup().stage(stage))`
+- `add_system(system.exclusive_system())` becomes `add_exclusive(system.exclusive_system())`
+- `add_startup_system(system.exclusive_system())` becomes `add_exclusive(system.exclusive_system().startup())`
+- `add_startup_system(system.label(x).before(y))` becomes `add_system(system.startup().label(x).before(y))`
+
+The rest should follow.
```
alice-i-cecile, line 24: I'd like to see a demonstration of how this works with system sets, and another with states as well.

alice-i-cecile:
IMO there's an important bit of **Future work** that should be included: this would provide an excellent unified foundation for "label descriptors", a much-discussed future feature that would allow users to specify schedule and run criteria information on labels, which would then be dispatched to each system with that label.

rfcs/2736-system_builder_syntax.md
thechubbypanda: Please see the updated examples :)

thechubbypanda:
> IMO there's an important bit of **Future work** that should be included

By included do you mean the changes required for label descriptors be made on top of this pr before merging it?

Ratysz:
"Future work" refers to a subsection of the RFC where you note down stuff that could happen after this RFC is accepted and implemented.

I don't think label properties are relevant here, though. It's mostly an orthogonal feature, in both API and implementation.

alice-i-cecile:
Alright, I accept @Ratysz's judgement on that.

rfcs/2736-system_builder_syntax.md
alice-i-cecile: Awesome, looks good!

alice-i-cecile (approved):

Nilirad:
I have a couple of ideas about the `add_exclusive(system.exclusive_system())` API:

1. Changing it to `add_exclusive(system)`: Yeeting `.exclusive_system()` since it is assumed that the system is exculsive.
2. Changing it to `add_system(system.exclusive_system())`: Removing `add_exclusive` method and specifying the exclusiveness through the builder-like pattern.

I'm quite ignorant to that part of the codebase so I don't know if the proposed APIs are possible or the implementation effort broaden too much the scope of the RFC/PR.

alice-i-cecile:
>Changing it to add_exclusive(system): Yeeting .exclusive_system() since it is assumed that the system is exclusive.

I'd be quite happy with this, although I'd prefer `add_exclusive_system` for clarity. @DJMcNab, you ran into some issues with yeeting `.exclusive_system` though right?

Ratysz:
> @DJMcNab, you ran into some issues with yeeting .exclusive_system though right?

That was a while ago, I've fixed that. This breaks it again; haven't had the time to try and fix it yet.

DJMcNab:
I didn't even try to get rid of `.exclusive_system` - to be honest, my PR was mostly just about recognising the `for<'a> &'a mut F: FnMut(...)` was the way to remove needing `.system`

rfcs/2736-system_builder_syntax.md
```diff
@@ -0,0 +1,76 @@
...
+## Summary
+
+Rewrite the system creation workflow to remove the excess of `add_startup_system_to_stage`-like methods and instead use the `system.stage().label().startup()` builder-like system.
+
+## Motivation
+
+Mostly described in [#1978](https://github.com/bevyengine/bevy/issues/1978).
+
+- Allows us to unify the existing `.label()`, `.after()`, etc. methods and the stage positioning of systems.
+- App would only have 3 methods to add systems: `add_system`, `add_exclusive` and `add_system_set`.
```
cart, line 12: I'm not convinced that giving exclusive systems their own api  / making them more separate is the direction we should be going in (at least, not yet). Exclusive systems are just systems with arbitrary side effects (new archetypes, new components, etc). Our plans for "stageless schedules" need to have some way to accommodate "normal system" side effects (ex: applying Commands). I personally think solving that problem well will involve blurring the lines between exclusive and normal systems, potentially to the point that distinguishing between the two isn't worthwhile. Given that we don't _yet_ know what this will look like in practice, I'd rather stick with what we currently have (which also happens to align better with the future that I think is coming). @Ratysz what are your thoughts on this?

Ratysz: I think separation is absolutely the right way to go. In fact, this PR doesn't go as far as it probably should - a lot of the weird API and implementation somersaults had to happen precisely to accomodate exclusive systems.

We should stop trying to merge parallel and exclusive systems until we actually have the ability to do so (which is, indeed, a part stageless), because right now we definitely can't blur the lines. Or so I [keep saying](https://github.com/bevyengine/bevy/pull/1144#issuecomment-755269848).

Ratysz: I forgot to mention: we tried here. It looks like a Rust limitation, it refuses to infer types across one of the boundaries.

cart: Haha I figured you'd take that stance. I don't have strong opinions here because exclusive systems are relatively uncommon and there are definitely good reasons to keep them separate in the current implementation. This change just creates more churn for existing exclusive system users and creates a situation where in the future we'll probably break them again in favor of something that will probably be closer to the current api. To me the (possibly temporary) marginal wins we get from this separation don't outweigh the cost of breaking everyone, but I won't push back too hard on this if everyone else thinks differently.

cart: > I forgot to mention: we tried here. It looks like a Rust limitation, it refuses to infer types across one of the boundaries.

For what its worth, I made some changes to exclusive systems in #2777 that probably enable this.

Ratysz: > This change just creates more churn for existing exclusive system users and creates a situation where in the future we'll probably break them again in favor of something that will probably be closer to the current api. To me the (possibly temporary) marginal wins we get from this separation don't outweigh the cost of breaking everyone, but I won't push back too hard on this if everyone else thinks differently.

Don't get me wrong, I also subscribe to this line of thinking; it's too early to merge them, too late to not merge them - hence the dredged-up post. It's hard to not be a little bitter after time and time again seeing a shiny new API get completely blown up and polluted with faux specialisation and pseudo-HKT noise as soon as it needs to account for exclusive systems.

> For what its worth, I made some changes to exclusive systems in #2777 that probably enable this.

Unlikely, the issue is the same that lead to `enum SystemDescriptor`: we need to have a concrete type somewhere between generic system and its consumer. Only this time the boundary is smeared across half a dozen traits with 3-4 implementors each.

cart: > It's hard to not be a little bitter after time and time again seeing a shiny new API get completely blown up and polluted with faux specialisation and pseudo-HKT noise as soon as it needs to account for exclusive systems.

Yup I totally get that and concede the point that the current implementation would be much cleaner if we embraced a hard split. But from my perspective its hard not to be bitter that we keep investing in a "detour" in the interest of simplicity of the current executor. We wouldn't need to care about ugly faux specialization and pseudo-HKT noise  if we just decided to treat exclusive systems as "normal systems" like we did in the old executor. Which I'm relatively certain is where we are headed in the future too. I pushed back pretty hard on the exclusive system split when the new executor landed and honestly its becoming clear to me that this split really is a "detour" that optimizes for a specific executor implementation that is lacking a feature that we all know we want. If cleanness and simplicity are the goals, imo our time is best spent on re-unifying exclusive systems and normal systems / formally acknowledging that Systems can have side effects (which again, is what the old impl did and what I pushed for in the new executor).

Im being a bit overly dramatic here. Not trying to be combative / turn this into an argument.  I'm very happy with the improvements the new executor brought and we may very well have been better served by embracing fully separate exclusive systems in the short term. Just trying to illustrate that I also have some strong feelings about this situation / trying to qualify why I think investing further in the system/exclusive system split isn't what I want personally.

Ratysz: For the most part, I agree; I don't think being conservative about splitting is unreasonable. I keep trying to advocate for it because I believe that the amount of headache it saves us in the short term is far, far greater than the amount of headache we will endure when mending this schism further down the line.

thechubbypanda:
So how's this going. Has cart's prerequisite stuff been merged?

alice-i-cecile:
Fixed by #45.

PR #32, opened 2021-09-09
rc_event_queue 

tower120:
This is rather true RFC (request for comments), than specification, at least for now.

[rc_event_queue](https://github.com/tower120/rc_event_queue) is not a cargo crate yet.

It is somewhat close to https://github.com/bevyengine/rfcs/pull/17 ... but implementation is different.

### DIFF
--- a/rfcs/reader_counted_event_queue.md
+++ b/rfcs/reader_counted_event_queue.md
@@ -39,12 +39,12 @@ Maintenance (reclaim memory):
 * Each frame - for each `EventQueue` call `cleanup`. 
 * **[SAFETY MEASURE]** Each Nth frame - check length of each `EventQueue`. If it is too big, truncate first chunks/clear (this will not free memory yet). 
   Next (this will move all EventReaders to new position immediately):
-  - Either. On all associated EventReaders should be called `EventReader::iter()` or `update_position()` (EventReaders should be presumably "queried" from systems locals, that read that EventType). 
+  - Either. On all associated EventReaders should be called `update_position()` (EventReaders should be presumably "queried" from systems locals, that read that EventType). 
   - Either. Each frame, in scheduler, for systems with EventReaders, that not run,  `update_position()` should be called.
     This is very fast operation.
 
 If there is time window, where nothing happens, and we just like wait for v-sync, it is a good place to put maintenance there. 
-Maintenance can be skipped few frames, and sliced between several frames. But I think this should be not necessary, since I expect
+Maintenance can be skipped few frames, and sliced between several frames. But I think this should not be necessary, since I expect
 maintenance to be very fast.
 
 ## Drawbacks
--- a/rfcs/reader_counted_event_queue.md
+++ b/rfcs/reader_counted_event_queue.md
@@ -1,0 +1,77 @@
+# Feature Name: `reader_counted_event_queue`
+
+## Summary
+
+Eliminate 2 frame event limitation, eliminate need in ManualEventReader. Without performance sacrifice.
+
+Achieve through integration of https://github.com/tower120/rc_event_queue.
+
+## Motivation
+
+Current Event implementation may be considered as logically dangerous, since systems that rely on events may simply 
+ not receive messages. And this could happen silently. ManualEventReader solution ... is not ideal.
+
+This RFC attempts to solve this problem completely.
+
+## User-facing explanation
+
+It looks like, there should be no changes to current state of affairs.
+Except that events will survive 2 frames window.
+
+## Implementation strategy
+
+[How "reader counted event queue" works.](https://github.com/tower120/rc_event_queue/blob/master/doc/principal-of-operation.md)
+
+`rc_event_queue` should work in `AUTO_CLEANUP = false` mode. Meaning that we will call `cleanup` manually.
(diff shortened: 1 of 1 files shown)


DJMcNab:
Thanks for creating this RFC - how events interact with 'asynchronous' systems is a real problem area and I'm glad to have more suggests in this space. 

However there is a licensing issue which needs to be resolved before we can proceed - `rc_event_queue` currently does not have a license available.

Note for the author: Please provide a license for `rc_event_queue`. 
In general, I want us to avoid introducing new dependencies not available under MIT/Apache2.0, so I would recommend you use this combination of licenses - this is what `bevy` itself is licensed under.
The easiest guide to doing so is [C-permissive](https://rust-lang.github.io/api-guidelines/necessities.html#crate-and-its-dependencies-have-a-permissive-license-c-permissive) from the Rust API guidelines. 
Please note that bevy chooses not to include a copyright notice line in [our MIT license](https://github.com/bevyengine/bevy/blob/main/docs/LICENSE-MIT) following the lead of the Rust project, although I have no opinion for or against it for `rc_event_queue` (although if you made me choose I'd err on the side of not including one).

For other readers: I should strongly advise against reading the code of `rc_event_queue` until this is resolved out of an abundance of caution. This also applies to the docs linked in the RFC, which from my cursory reading (of the RFC) seem central to explaining how the proposed change would work.

tower120:
> In general, I want us to avoid introducing new dependencies not available under MIT/Apache2.0, so I would recommend you use this combination of licenses - this is what bevy itself is licensed under.

What's a point of dual licensing? Isn't MIT more permissive than Apache2.0? Initially I was going to license just under MIT.

> Please note that bevy chooses not to include a copyright notice line in our MIT license...

What is that mean?

bjorn3:
> What's a point of dual licensing? Isn't MIT more permissive than Apache2.0? Initially I was going to license just under MIT.

Apache-2.0 has extra rules regarding patents. IANAL, but I think it roughly prevents contributors from contributing code to which a patent they own applies without giving a patent grant to all users of the code. It is incompatible with GPL-2.0 according to some lawyers though, hence the dual license with MIT which is clearly compatible.

tower120:
License added. If everything goes well - it should become cargo crate. So it can be just another dependency for project, like every else.

tower120:
By the way, some cargo crates with MIT/APACHE2 have dependencies on MIT only crates (including bevy and rc_event_queue)... Is this ok?

bjorn3:
https://law.stackexchange.com/questions/6081/can-i-bundle-mit-licensed-components-in-a-apache-2-0-licensed-project

tl;dr: yes

tower120:
@alice-i-cecile You said something about that rc_event_queue could be useful for something UI specific. Did you mean Entity Events?

Cuurently, EventQueue must be `Pinned`, because Readers need to talk back when passing inter-chunk boundary, so its stored in heap, to have stable address.

But I think I can make it non-pinned, you'll just have to provide Reader &EventQueue on each read session. Also, I think, it is possible to built-in one-two small chunks into EventQueue, thus having some form of small_vec....

alice-i-cecile:
> @alice-i-cecile You said something about that rc_event_queue could be useful for something UI specific. Did you mean Entity Events?

Ah right! Yes, [per-entity events](https://github.com/bevyengine/bevy/pull/2116) were particularly blocked on solving the questions around event lifespan, as I found they tended to have more unusual access patterns. And then in turn, I felt that storing events on entities might be quite useful for handling UI event dispatching.

> Cuurently, EventQueue must be Pinned, because Readers need to talk back when passing inter-chunk boundary, so its stored in heap, to have stable address.

Interesting, that makes sense. FWIW, `NonSend` components should be quite feasible; we've just never had a pressing need for them before. The overhead of just doing that may be better than your non-pinned variant, but I'll leave that judgement call to you.

tower120:
Main overhead with Arc is construction time, everything else is mostly non-observable, at least for current Bevy Events use-case. 

But if you really plan to put EventQueue into Component and add/remove components dynamically, that overhead is a thing.
Considering small buffer optimization implementation, memory-wise it could be approx. 64byte + chunk size, non-heap (until fits small chunk), and you can move it around. As a price for that, you'll need to pass `&EventQueue` to get `Reader::iter()`.

So I would say, realistically it would be 128-256 bytes on stack. Could that work for your cause?

alice-i-cecile:
> So .... I thought, what about having some new EventsV2 alongside with the current one? This way - new events will not be a braking change....
Is that acceptable?

I would be hesitant to do that; this should work fine as an external crate if we want early usage feedback.

Keeping both in the engine leads to the sort of caveats and confusion you tend to see in e.g. Unity, where you have two very similar but subtly different tools for the job, whose usage is mixed throughout the code base.

alice-i-cecile:
FYI @tower120, we're mostly just blocked on "decision-making attention": there are a large number of other pressing ECS needs that have been getting more attention right now (https://github.com/bevyengine/bevy/discussions/2801) and Cart is focusing on rendering to get 0.6 released <3

tower120:
I see... And API changes, like making `EventReader::iter` return `Iterator`, instead of `DoubleEndedIterator`, is out of the question. Right?

PR #35, opened 2021-10-12
(Hierarchical) Stage Labels, Conditions, and Transitions

maniwani:
[RENDERED][1]

###  Summary

Reorganizes the system scheduling API with better separation of responsibilities. Introduces a single schedule executor, reincarnates stages into topological system labels, and splits run criteria into conditions and transitions.

### Motivation

The current system scheduling API packs great functional coverage into a tasteful number of types, but there are some key things missing, and the overall cohesion isn't as good as it could be. 

See #2801 for more background.

[1]: https://github.com/maniwani/rfcs/blob/stage-labels-rfc/rfcs/35-stages_are_labels.md


maniwani:
Closing since `@alice-i-cecile` is spearheading an RFC for a design a bunch of us discussed in the Discord that incorporates a fair bit of what's in here—only immutable fetches in run criteria, run criteria returning `bool`, reusable system descriptors, one executor for the schedule. 

Some differences are the complete removal of stages (with the addition of an explicit sync point system) and a simpler state machine (no stack whatsover). Any looping and nesting will involve running a nested schedule inside an exclusive system.

PR #36, opened 2021-10-20
Encoding Side Effect Lifecycles as Subgraphs

hymm:
This is an alternative to the scheduling api from https://github.com/bevyengine/rfcs/pull/34 that covers more types of side effects.

[Rendered](https://github.com/hymm/rfcs/blob/side-effect-lifecycle-scheduling/rfcs/side-effect-lifecycle-scheduling.md)


rfcs/side-effect-lifecycle-scheduling.md
```diff
@@ -0,0 +1,124 @@
...
+* Readers and writers must be after the producer and before the consumer. They can run in any order in relation to other readers and writers. If there needs to be an order between readers and writers, use the normal before and after labeling.
+* Consumers are run after producers, readers, and writers. There should only be one consumer of a type of side effect.
+
+### Some Simple Examples
+
+#### Scheduling around command application
+```rs
+fn main() {
+    App::new()
+        .add_system(apply_commands_buffer.consumes("commands"))
```
alice-i-cecile, line 34: I'd prefer to use strongly typed labels here when possible, especially for standard concepts like Commands.

rfcs/side-effect-lifecycle-scheduling.md
```diff
@@ -0,0 +1,124 @@
...
+        // if we don't care about when the commands are applied
+        // we can ignore the new api
+        .add_system(create_commands_A)
+        
+        // if we need to schedule a system after commands are applied
+        .add_system(create_commands_A.label("system A").produces("commands"))
+        .add_system(after_commands_A.after("system A").after("commands"))
```
alice-i-cecile, line 41: I'm not sure this will give the desired behavior: it will run after *all* of the available systems with the "commands" label are complete, when we want "at least one".

alice-i-cecile: This can be particularly nasty if you want to chain several of these things back-to-back within a frame.

rfcs/side-effect-lifecycle-scheduling.md
```diff
@@ -0,0 +1,124 @@
...
+    .before(writes(SideEffectLabel))
+    .before(comsumes(SideEffectLabel))
+)
+```
+
+> TODO: add desugaring of readers, writers, and consumers.
+
+## Drawbacks
+
+I expect that these relationships will mostly be added automatically when using run criteria, app state, and the like with a higher level api. Users would only be expected to interact with this API when things can't be done automatically or the default behavior doesn't match their needs.
```
alice-i-cecile, line 109: This feels more like a rationale than a drawback?

rfcs/side-effect-lifecycle-scheduling.md
```diff
@@ -0,0 +1,124 @@
...
+    .label(produces(SideEffectLabel))
+    .before(reads(SideEffectLabel))
+    .before(writes(SideEffectLabel))
+    .before(comsumes(SideEffectLabel))
+)
+```
+
+> TODO: add desugaring of readers, writers, and consumers.
+
+## Drawbacks
```
alice-i-cecile, line 107: IMO the big drawback to this is that it further increases the complexity of the scheduling graph.

This is unavoidable IMO, but users need better tools to inspect and debug this.

rfcs/side-effect-lifecycle-scheduling.md
```diff
@@ -0,0 +1,124 @@
...
+
+* The example above for expressing the relationships between side effects using just labeling and system sets has one major problem. There is no current api for adding systems to system sets after the system set has been created. If such an api did exist we could potentially just add systems to the correct system set. This didn't seem better than the proposed API and has some implementation complexity since system sets are mostly just api sugar for labeling multiple systems and does not exist at runtime.
+
+## Future Work
+
+* This is part of a bigger redesign of the ECS trying to solve problems with scheduling around hard sync points and data consistency between states. This RFC by itself doesn't solve those problems, but it does encode information about dependencies that other RFCs might need to actually solve the data consistency problems. 
+* Side effects are effectively typed. Build an API to pass and send side effects and enforce the typing between producers and consumers. This could either be with data stored in the executor or a Resource that manages access through access tokens (what Boxy is working on).
+* How to specify an order to how commands are applied?
```
alice-i-cecile, line 123: I don't think this is particularly related.

rfcs/side-effect-lifecycle-scheduling.md
```diff
@@ -0,0 +1,124 @@
...
+* This API doesn't really directly address data consistency problems due to changes to run criteria. Ideally we'd figure out a way to enforce that the consumer needs to run if side effects have been produced.
```
alice-i-cecile, line 124: I started to chew on this in my pre-draft RFC here: https://github.com/alice-i-cecile/rfcs/blob/system-graph-flavors/rfcs/34-system-graph-flavors.md#causal-ties-systems-that-must-run-together

rfcs/side-effect-lifecycle-scheduling.md
```diff
@@ -0,0 +1,124 @@
...
+
+We can express this required ordering with labels, before, and after, but they lack semantic meaning. This is an api for creating subgraphs that express the required ordering between producers, consumers, readers, and writers.
+
+## User Facing Description
+
+### Scheduling Side Effects API
+* `.produces(Label)` System produces a side effect of type `Label`
+* `.consumes(Label)` No read/modifies systems can run after this system. An example that would use this is the system that applies commands to the world.
+* `.reads(Label)` system has read access to side effect and must run after the producers and before the consumer. An example of this type of system would be systems that depend on a run criteria.
```
alice-i-cecile, line 19: Currently, reads and writes have identical behavior. They should either be merged, or readers should always run after writers (a fairly standard pattern).

rfcs/side-effect-lifecycle-scheduling.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Open questions
+
+* With looping criteria it might be possible to create a valid dependency cycle that cannot be resolved in the current API. If that is the case then we might need an `.after_can_break` type of edge that could be ignored in dependency cycle checking.
+
+## Alternatives Considered
+
+* The example above for expressing the relationships between side effects using just labeling and system sets has one major problem. There is no current api for adding systems to system sets after the system set has been created. If such an api did exist we could potentially just add systems to the correct system set. This didn't seem better than the proposed API and has some implementation complexity since system sets are mostly just api sugar for labeling multiple systems and does not exist at runtime.
+
+## Future Work
```
alice-i-cecile, line 119: Indexing and relations are two other highly visible use cases that are enabled by solving the data consistency problem.

PR #37, opened 2021-10-20
Dynamic Queries

GarettCooper:
[RENDERED](https://github.com/GarettCooper/rfcs/blob/dynamic_query/rfcs/37-dynamic_query.md)

Expands the query interface of `World` to support queries with parameters which are determined at runtime, not compile time, to enable future scripting and hot reloading.

I wrote this RFC as a way of learning how Bevy's ECS worked, so I had to implement most of it to make sure what I was proposing was possible. [You can find my branch here.](https://github.com/GarettCooper/bevy/tree/dynamic_query/crates/bevy_ecs/src/query/dynamic)


rfcs/37-dynamic_query.md
```diff
@@ -0,0 +1,168 @@
...
+                        .optional_component(layer_component_id)
+                        .without(frozen_component_id)
+                        .build();
+
+let query_state = world.query_dynamic(&query);
+```
+
+A static query returns the various components it has retrieved in a tuple. This isn't an option for a Dynamic Query
+however, since the number of components being returned isn't known at compile time. Instead, a Dynamic Query returns 
+a `DynamicQueryEntity` struct, which is both indexable and iterable to access the individual `DynamicItem`s it
```
alice-i-cecile, line 45: Allowing indexing and iteration over query items to static queries might be a nice bit of API consistency. I've seen users ask for an alternative to tuple unpacking before in favor of a more "named" API.

rfcs/37-dynamic_query.md
```diff
@@ -0,0 +1,168 @@
...
+```rust
+pub enum DynamicItem {
+    Entity(Entity),
+    Component(DynamicComponentReference),
+    MutableComponent(DynamicMutComponentReference),
+    ComponentNotPresent
+}
+```
+The two inner types, `DynamicComponentReference` and `DynamicMutComponentReference` provide safe access to 
+`&`/`&mut` and change tracking functionality while also enabling users to have access to the underlying pointer for
```
alice-i-cecile, line 59: Presumably the underlying pointer API will have to be extremely unsafe?

rfcs/37-dynamic_query.md
```diff
@@ -0,0 +1,168 @@
...
+                        .mut_component(position_component_id)
+                        .without(frozen_component_id)
+                        .build();
+
+for result in world.query_dynamic(&query).iter_mut() {
+    info!("Updating the position of Entity {}", result[0].unwrap_entity().id())
+    result[2].unwrap_mut_component().downcast_mut::<Position>().unwrap() += result[1].unwrap_component().downcast_mut::<Velocity>.unwrap();
+}
+```
+
```
alice-i-cecile, line 75: IMO this really needs to show a more representative use case after the existing "basic usage" example.

Right now, all of the functionality demonstrated is simply less safe and slower than the existing tools :p

rfcs/37-dynamic_query.md
```diff
@@ -0,0 +1,168 @@
...
+## Unresolved questions
+
+These questions remain unresolved:
+- How much of the type's information should the `DynamicComponentReference` types carry with them (Layout?)?
+- How can change tracking information be captured in dynamic queries?
+- Should dynamic queries be hidden behind a feature?
+- Which traits should the `DynamicItem`/`DynamicItemSet` types implemented for maximum convenience?
+- Do we even want maximum convenience?
+
```
alice-i-cecile, line 155: How do modding / scripting use cases access resources?

rfcs/37-dynamic_query.md
```diff
@@ -0,0 +1,168 @@
...
+queries. In that one small way, the builder pattern employed by this RFC offers a more ergonomic method for
+constructing queries, though the merits of those ergonomics for a feature only itended for advanced usage is
+questionable.
+
+## Unresolved questions
+
+These questions remain unresolved:
+- How much of the type's information should the `DynamicComponentReference` types carry with them (Layout?)?
+- How can change tracking information be captured in dynamic queries?
+- Should dynamic queries be hidden behind a feature?
```
alice-i-cecile, line 152: Given that many of the other bits of special-case functionality (e.g. audio), I think it should be reasonable to do so.

If it is, this should not be a default feature. Ideally it's named in such a way that we can easily add other related bits of functionality to the same feature flag.

This would help reduce compile times slightly (I think), but more importantly, make it much more challenging for beginners to accidentally stumble across.

rfcs/37-dynamic_query.md
```diff
@@ -0,0 +1,168 @@
...
+- Which traits should the `DynamicItem`/`DynamicItemSet` types implemented for maximum convenience?
+- Do we even want maximum convenience?
+
+## Future possibilities
+Dynamic queries lay the groundwork for future expansion of the ECS to fully support runtime behaviour changes. 
+Plugins could dynamically define systems at runtime based on configuration files or scripts, since they now have the 
+tools to interact with the Entity Component System freely. This addition of this single RFC would still leave Bevy a
+long way from full dynamic support though. Future RFCs should set up implementations for inserting new systems and
```
alice-i-cecile, line 160: Link to https://github.com/bevyengine/bevy/pull/2507

CGMossa:
What's happened?

GarettCooper:
I haven't had the time to pursue this and it seems like there community has been happy to move on without it with a lot of great work being done in the dynamic/reflection space. I figured this was just adding noise to the open RFCs list and was better off closed.

zicklag:
If anybody is looking for dynamic query support, there is now the work-in-progress, but very promising, [`bevy_ecs_dynamic`](https://github.com/jakobhellermann/bevy_ecs_dynamic) crate. It's being used to help implement a JavaScript scripting API in [`bevy_mod_js_scripting`](https://github.com/jakobhellermann/bevy_mod_js_scripting) which is also being put to actual use in the [Punchy](https://github.com/fishfolks/punchy) and [Jumpy](https://github.com/fishfolks/jumpy).

It's all a little work-in-progress and under development, but so far it looks like it's going to turn out great.

PR #41, opened 2021-11-06
Generic UI navigation system

nicopap:
[RENDERED](https://github.com/nicopap/rfcs/blob/ui-navigation/rfcs/41-ui-navigation.md)

Two new components and a new system to define navigation nodes for the ECS UI.

On top of making it easy to implement controller navigation of in-game UIs, it enables keyboard navigation and answers accessibility concerns such as the ones [raised in this issue](https://github.com/bevyengine/bevy/issues/254).

# Roadmap

- [X] [Create a MVP that acts as a bevy plugin](https://github.com/nicopap/ui-navigation)
- [X] Design doc of latest implementation
- [ ] Solve open questions
    - [X] Naming: `NavFence` ⇒ `NavMenu`?
    - [x] Game-orientation and potential conflict with editor design
    - [X] **`NavRequest` and input handling**
    - [ ] Any other that were not asked in this RFC?
- [X] Implement downward navigation optimization & "deep focus memory"
- [X] Mouse support
- [ ] Find a name for the bevy crate, since `bevy_ui_navigation` is already taken (my bad :cry:)
- [ ] https://github.com/bevyengine/bevy/pull/5378
- [ ] After merge, open issues for identified shortcomings and possible improvements:
    - [ ] Multicursor
    - [ ] tab-navigation
    - [ ] make public the `TreeMenu` systems (`insert_tree_menus` and `set_first_focused`)
    - [ ] Higher level wrapper for menu switching
    - [ ] Handling of `Display::None` (if it doesn't work already)
    - [ ] Robustness from deleting parent menus while inside of them
    - [ ] Optimizations, reduce number of iterated focusables in various places


nicopap:
Implementation going on here: https://github.com/nicopap/ui-navigation

TheRawMeatball:
This is super exciting, though I have concerns about having people manually build the navigation tree. I think all navigation info should be derived from the existing widget tree, at no extra cost to the developer. The way webpages handle this stuff should probably also be considered under prior art.

Also btw, if you have a discord account it'd be great if came in and said hi at #ui-dev. I don't think Github is ideal for the fast-paced brainstorming etc. that happens early on in design.

nicopap:
The user only has to add `Navigable` components to relevant nodes. With this, the system should be able to construct a flat navigation graph based solely on the `Transform` (or `GlobalTransform` maybe?) of the nodes.

They only need to add `Container` components if they want special behaviors for different menus (like the tabbed menu in the example). Or hierarchical menu navigation (like going option > graphical options > Resolution)

I think it's necessary to specify the navigable nodes, because there is a lot of irrelevant-to-navigation nodes in the UI tree. Some nodes only exist to specify the layout, others are just pictures etc.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,394 @@
...
+
+On top of making it easy to implement controller navigation of
+in-game UIs, it enables keyboard navigations and answers accessibility concerns
+such as the ones
+[raised in this issue](https://github.com/bevyengine/bevy/issues/254).
+
+## User-facing explanation
+
+We add a _navigation tree_ to the UI system, the tree will be used to enable
+key-based navigation of the specified UIs.
```
alice-i-cecile, line 17:
```suggestion
key-based (and joystick controllable) navigation of UI, by allowing users to rotate between different UI elements.
```

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,394 @@
...
+various buttons in the menus and specify which button press leads to other menu
+items, etc. Or more simply, limit themselves to pointer-based navigation to
+avoid having to think about menu navigation.
+
+By including navigation-related components such as `Navigable` and `Container`,
+the developer can quickly get a controller-navigable menu up and running. The
+`Container` accepts various `Plane`s to fine-tune more precisely which keys
+trigger which kind of navigations.
+
+To interact in real-time with the UI, the developer emits `NavCommand`s events.
```
alice-i-cecile, line 29: We should avoid reusing `Command` in a similar but distinct context :)

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,394 @@
...
+key-based navigation of the specified UIs.
+
+Currently, the developer must themselves write the relationship between
+various buttons in the menus and specify which button press leads to other menu
+items, etc. Or more simply, limit themselves to pointer-based navigation to
+avoid having to think about menu navigation.
+
+By including navigation-related components such as `Navigable` and `Container`,
+the developer can quickly get a controller-navigable menu up and running. The
+`Container` accepts various `Plane`s to fine-tune more precisely which keys
```
alice-i-cecile, line 26: Users have no idea what either a container or a plane is here, unless they've already seen them in other UI frameworks.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,394 @@
...
+trigger which kind of navigations.
+
+To interact in real-time with the UI, the developer emits `NavCommand`s events.
+The navigation system responds with `NavEvent` events, such as
+`NavEvent::FocusChanged` or `NavEvent::Caught`.
+
+The navigation system doesn't handle itself the state of the UI, but instead
+emits the `FocusChanged` etc. events that the developer will need to read to,
```
alice-i-cecile, line 34: IMO, this shouldn't be an event. Instead, toggle the value of `Focused` enum (like we have for buttons), and then let the developers use change detection for this.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,394 @@
...
+Therefore, with the little information we provided through this hand wavy
+specification, we should already have everything setup for the navigation to
+just work™.
+
+
+### Dimensionality
+
+Let's go back to our ambition of using LT/RT for navigation now. 
+
+UI navigation is not just 2D, it's N-dimensional. The
```
alice-i-cecile, line 135: Yes, firmly agree.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,394 @@
...
+LT/RT to change tabs is like navigating in a 3rd orthogonal dimension to the
+dimensions you navigate with UP DOWN LEFT RIGHT inside the menus.
+
+I'll call those dimensions `Plane`s because it's easier to type than
```
alice-i-cecile, line 139: Oh I see. `Plane` is not a great word for this: it implies multidimensionality, and doesn't work for looping behavior, which is both common and important.

`Axis` is probably the most common word that could be used here.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,394 @@
...
+
+We posited we could easily infer the physical layout and bake a navigation map
+automatically based on the `Transform` positions of the elements. This is not
+the case for the dimensions of our navigation. We can't magically infer the
+intent of the developer: They need to specify the dimensionality of their menus.
+
+Let's go back to the menu. In the example code, I refer to `:container` and
+`:navigable`. I've not explained yet what those are. Let's clear things up.
+
+A _navigable_ is an element that can be focused. A _container_ is a node entity
```
alice-i-cecile, line 167: I suspect we can just reuse the existing `Interaction` component, probably reflavoring to `Focus`. The venn diagram is a circle. so we can use the same component for both marking and storing the state.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,394 @@
...
+```rust
+struct Container {
+  inner: Option<Box<Container>>,
+  siblings: NonEmptyVec<Navigable>,
+  active: SiblingIndex,
+  plane: Plane,
+}
+```
+Note: this is the data structure for the navigation algorithm, the ECS
+componenet called `Container` will probably look like this:
```
alice-i-cecile, line 185:
```suggestion
component called `Container` will probably look like this:
```

rfcs/41-ui-navigation.md
TheRawMeatball: While for the current system this'd be the best option, in the context of a possibly larger reworking of UI adding a separate concept of "UI events" with propagation and using said events for this would probably be better. Out of this RFC s scope though.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,394 @@
...
+```rust
+#[derive(Component)]
+struct Container {
+  plane: Plane,
```
alice-i-cecile, line 189: So, given that we're using separate entities for these containers already, and you could have multiple of them, I think the planes (or whatever bikeshed names) should be components of the `Plane<P>` type.

Then, you can have separate systems for each plane type (using a generic helper method on `App`, stealing tech from events) and navigate very quickly using query filters. This will also make the meaning of each container much clearer in various tools like `bevy_egui_inspector`, since we don't need to go digging into the values.

Then, since looping is defined on a per-plane basis, this should be defined in the helper method on `App`.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,394 @@
...
+### Unique hierarchy
+
+If we implementation this with a cached navigation tree, it's going to be
+easy to assume accidentally that we have a single fully reachable navigation
+tree. This is not obvious, it's easy to just add the `Container` or `Navigable`
+`Component` to an entity that doesn't have itself a `Container` parent more
+than once.
+
+This needs to be solved, I'm thinking that an error message would be enough in
+this case. But it's worth considering design changes that makes this situation
```
alice-i-cecile, line 344: IMO this is out of scope for this RFC. This flavor of "graph invariant" is one of the central (and most challenging) problems in relations (a massive ECS-focused effort to better represent entity-entity connections).

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,394 @@
...
+I'm implementing first to see how cumbersome this is, and maybe revise the
+design with learning from that.
+
+Maybe it's possible to integrate more than one `Container` within other
+containers. The only motivation for this currently is that it makes it easier to
+implement and reason about the system.
+
+### Inflexible navigation
+
+The developer might want to add arbitrary "jump" relations between various
```
alice-i-cecile, line 364: IMO this should be handled by overriding the existing logic in some form, rather than something we build in an escape hatch for.

Building something in makes it much harder to teach, write and maintain and doesn't actually increase flexibility. The ECS is already an extremely powerful tool for this sort of customization.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,394 @@
...
+within the menu. I think this might enable cycling references within the
+hierarchy and break useful assumptions.
+
+## Unresolved questions
+
+Building the physical navigation tree seems not-so trivial, I'm not sure how
+feasible it is, although I really think it's not that difficult.
+
+
+## Isn't this overkill?
```
alice-i-cecile, line 379: Absolutely not. I think this is about as simple as we can get while still allowing something properly robust to complex user needs.

And using two seperate frameworks (one for simple cases, and then an incompatible one for complex ones) is a terrible idea due to the fragmentation and migration pain it induces.

TheRawMeatball:
> The user only has to add Navigable components to relevant nodes. ... They only need to add Container components if they want special behaviors for different menus. ...

While I agree with `Navigable`, I'm not quite as sure on what `Container` gets us. Wouldn't a simple depth-first iteration over the ui tree where we only look at `Navigable`s get us something basically equivalent?

rfcs/41-ui-navigation.md
nicopap: There is two reasons behind this:
1. It's a way for me to completely hand-wave away everything but exactly the navigation algorithm
2. It responds to that [specific request from ndarilek](https://github.com/bevyengine/bevy/issues/254#issuecomment-679359772)

rfcs/41-ui-navigation.md
nicopap: I really like this suggestion! I'll keep in mind for the implementation.

nicopap:
I thought a bit more about the design. A few things:
1. I want to get rid of the concept of `Plane`. Beside 2D movements with arrows, the only real use case for other "dimensions" is just going through a tabbed menu and going up/down the hierarchy. The idea of a "select" plane doesn't make sense and doesn't really work.
2. The "one container within other containers" is too limiting and I feel it's not too much of a burden to have a tree with multiple branches.
3. `Navigable` ⇒ `Focusable`. Hopefully this clarifies the difference with `Container`s.

Tomorrow I'll add a roadmap for the implementation in the [ui-navigation repo](https://github.com/nicopap/ui-navigation)

nicopap:
> > The user only has to add Navigable components to relevant nodes. ... They only need to add Container components if they want special behaviors for different menus. ...
> 
> While I agree with `Navigable`, I'm not quite as sure on what `Container` gets us. Wouldn't a simple depth-first iteration over the ui tree where we only look at `Navigable`s get us something basically equivalent?

The difference between `Container` and `Navigable` is that you can *focus* on a navigable, while the container cannot be focused. In the example I gave, it wouldn't make sense to be able to focus on the entire submenu, including the panels and the "ABC" menu. Only the specific panels and the individual circles should be able to be focused. The `Container` is still necessary because it separates the "ABC" menu from the "soul" menu. You would have to press A and B to go from one to another. The other thing is that the "ABC" menu is not the same kind of thing as the panels on the left side, and when cycling through the panels, you shouldn't be able to end up in the large menu.

A case can be made that you should. This was part of the initial design, but as I came up with a more concrete proposal, it became less practical.

In the implementation, the `Container` also act as hard boundary when searching siblings in the UI hierarchy, reducing the depths of searches, and specifying exactly where to look for focusable elements, rather than going through dead ends.

nicopap:
Updated [plugin demo](https://github.com/nicopap/ui-navigation) with reference and bevy-specific implementations. There is an example usage at [flat_2d_nav.rs](https://github.com/nicopap/ui-navigation/blob/master/bevy/examples/flat_2d_nav.rs)

I'm surprised how the ECS version is basically just a series of function call. Although the flipside is that it's way more error-prone than the "generic" version which enforces the invariants at the API level.

Next step is clarify this RFC. Afterward, I'll add Action/Cancel to the implementation which seems trivial.

TheRawMeatball:
> The other thing is that the "ABC" menu is not the same kind of thing as the panels on the left side, and when cycling through the panels, you shouldn't be able to end up in the large menu. A case can be made that you should. This was part of the initial design, but as I came up with a more concrete proposal, it became less practical.

I'm very not sure whether we should block traversing out of a container node. A counter example might be a webpage or most other UI, where you can visit every focusable UI node by hitting tab. I don't think we should have hard boundaries.

nicopap:
> I'm very not sure whether we should block traversing out of a container node. A counter example might be a webpage or most other UI, where you can visit every focusable UI node by hitting tab. I don't think we should have hard boundaries.

Nothing to worry about. `Container` (that I renamed for now into `NavNode`) is entirely optional [in this example](https://github.com/nicopap/ui-navigation/blob/master/bevy/examples/flat_2d_nav.rs) `NavNode` is not even imported.

In [this example](https://github.com/nicopap/ui-navigation/blob/master/bevy/examples/flat_2d_across_nodes.rs) we explicitly use `NavNode` as a way to isolate some parts of the UI from other. You'll notice that we can still move from one container to the other, only the one explicitly marked with a `NavNode` is closed-off.

Maybe naming it `NavigationBarrier` would be more explicit?

alice-i-cecile:
> Maybe naming it `NavigationBarrier` would be more explicit?

Yes please, although probably `NavBarrier` for consistency.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,369 @@
...
+various buttons in the menus and specify which button press leads to other menu
+items, etc. Or more simply, limit themselves to pointer-based navigation to
+avoid having to think about menu navigation.
+
+By including navigation-related components such as `Focusable` and `NavFence`,
+the developer can quickly get a controller-navigable menu up and running.
+
+We could also add the `Focusable` component to standard UI widget bundles such
+as `ButtonBundle` and a potential `TextField`. This way, the developer _doesn't
+even need to specify the `Focusable` elements_.
```
nicopap, line 55: @TheRawMeatball 
> This is super exciting, though I have concerns about having people manually build the navigation tree. I think all navigation info should be derived from the existing widget tree, at no extra cost to the developer.

It is completely possible to make the widgets navigable without extra cost to the developers, simply by adding the `Focusable` component as a field to the various relevant component bundles. I didn't think of it earlier because I was locked in the "this is a plugin" mindset.

JonahPlusPlus:
@alice-i-cecile told me to check this out. I got to say, I really like this design, and especially appreciate how `NavCommands` are sent by the app and not the library. If you add mouse support, I think you should make it another `NavCommand` that stores a `Vec2` or something, and let people create the bindings themselves (unlike the current UI focus system). It requires more work on their side, but I think if bevy wants to work for bigger and more diverse titles, it needs flexibility.

nicopap:
I cooked up a new design! It solves basically all the unresolved questions. It's a relatively minor implementation change, but it's a major change in the way to think about the navigation graph, and it requires changing important parts of this RFC.

I'm quite fed up editing graphs in Inkscape, but I think this is the last stretch and having a clear documentation on top of the implementation would be golden. Hopefully a good formal source of inspiration for implementing ui navigation in other UI projects too.

nicopap:
This is the final version of the RFC. It is very different from the first and second versions. I guess I'm requesting for comments :)

Nothing urgent. I implemented everything I need for now and don't expect to start work on the bevy PR before a while. I just need confirmation for the `NavRequest` open question.

This might even need a new RFC! How bevy handles inputs is important to discuss and this feature quite obviously deals with inputs.

JonahPlusPlus:
So mouse input would be handled with a `NavRequest::FocusOn`, correct? All you would have to do is check which entity the mouse is over when you click and send the `Entity`. I can get behind that. Gives more options for plugins to work with. As for naming, maybe `NavRequest::ChangeMenu` and `NavEvent::Pass` or `NavEvent::Through` or just `NavEvent::NoChange`?

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,400 @@
...
+            _ => {}
+        }
+    }
+}
+fn button_system(
+    materials: Res<Materials>,
+    mut focusables: Query<(&Focusable, &mut Handle<ColorMaterial>), Changed<Focusable>>,
+) {
+    for (focus_state, mut material) in focusables.iter_mut() {
+        if focus_state.is_focused() {
```
alice-i-cecile, line 123: Does this make more sense as a `match` on an enum? I tend to find that to be clearer, and it helps me catch edge cases more easily.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,400 @@
...
+
+I am against an inflexible system where it is impossible to change the input
+mappings, any respectable game let you change the input mapping. And if not,
+your players start very quickly to ask for one! So it is necessary to be able
+to remap the buttons.
+
+How would that work? Is there already examples in bevy where we provide default
+button mappings that can be changed?
+
+I think we should think deeply about input handling in default plugins.
```
alice-i-cecile, line 315: Agreed; this should be subsumed under whatever input mapping standard we agree upon.

rfcs/41-ui-navigation.md
nicopap: It's implemented as an `enum` under the hood! Makes total sense to expose it then.

nicopap:
> So mouse input would be handled with a `NavRequest::FocusOn`, correct? All you would have to do is check which entity the mouse is over when you click and send the `Entity`. I can get behind that. Gives more options for plugins to work with. As for naming, maybe `NavRequest::ChangeMenu` and `NavEvent::Pass` or `NavEvent::Through` or just `NavEvent::NoChange`?

I still think your suggestion of having a `NavRequest::MousePosition` or similar is, if not the best, at least a good solution. I guess it's still an open question, how does mouse hover state and click interact with the navigation system? I completely skipped that part. Do you have a suggestion?
* Hover ≠ Focus: example: hovering over a tab menu shouldn't modify focus (and change the submenu)
* Left click = `NavRequest::Action`?

There might very well be two systems that simply must deal with each other. For example a `MouseFocusPlugin` that manages its own state independently from the current proposal and emits `NavRequest::FocusOn` when relevant. Ahah, so exactly what you described.

JonahPlusPlus:
Hmmm... A `MouseFocusPlugin` could always just send both events on a click: `FocusOn` and `Action` and then you could add a `HoverOn` request that could be sent for movement that takes an entity as well. By handling `FocusOn` first, then `Action`, it would effectively click on whatever it is hovering over. 

Something like `MousePosition` would make it simple, but is inflexible the more I think about it. For instance, mapping space. Users may want to do complex stuff with ring menus, and perhaps they'll want to switch menu items even when not hovering exactly over them (sort of snapping the mouse to an item). Or perhaps they are making a sort of trolling puzzle game that makes you click on something else to click the button. 

Who knows, but it would be redundant to map the `Vec2` only for it to calculate which one needs to be clicked when you know which one needs to be clicked. I think it's better to keep it complex and let plugins hide that complexity.

JonahPlusPlus:
However, if the `FocusOn` technique is what gets implemented, in order to make it easier to work with, there should be a method in the UI to find what `Entity` is at a location. It wouldn't be necessary for people who store the `Entity` in some `Component` but for a lot of use cases, it would be annoying to implement something so basic.

nicopap:
While dogfooding the ui-navigation library with a large and straining usecase, I could identify limitation and improvement leads for the library. Developments in the design of the ui-navigation library:

## Cancel focusables

Adding a `cancel` mod to focusables gives a natural way to specify going backward in the menu hierarchy, especially with a mouse.

https://github.com/nicopap/ui-navigation/issues/2

## Lock focusables

Adding a `lock` mod to focusables could facilitate implementation of widgets with complex encapsulated controls. It could also simplify transition between gameplay and inventory.

https://github.com/nicopap/ui-navigation/issues/7

## Improvement to menu management

I often needed a way to check in which menu I'm currently in, there is no way to do that with the current implementation.

The way I "worked around" the limitation is by adding a component that specify the current menu to each `focusable` and query that component in the system responsible to manage `NavEvent` responses.

I explore designs in https://github.com/nicopap/ui-navigation/issues/4

This is still an open question.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,403 @@
...
+but loaded into the ECS. In this case, the navigation tree might look like
+this:
+
+![RPG menu graph without focusables, only
+menus](https://user-images.githubusercontent.com/26321040/141672033-8bd43660-78af-4521-b367-03c7270c58c5.png)
+
+### Algorithm
+
+The algorithm is the `resolve` function in [ui-navigation](https://github.com/nicopap/ui-navigation/blob/master/src/lib.rs#L414).
+The meat of it is 50 lines of code.
```
nicopap, line 242: This is now false. The `resolve` function is 111 lines of code now

aevyrie:
This looks great. My only note is that mouse input will need to be abstracted to any type of `Pointer`, including gamepad controlled software cursors. This is something I'm working on, and can integrate on top of this when I upstream. Glad this adds focus, I was missing this with pointer interactions.

nicopap:
This is very out of date. The design changed in key parts since the last draft (remember, this is basically a specification of https://github.com/nicopap/ui-navigation). I actually started drafting an update last week, but  got caught up in things. I'll prioritize a little bit.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+The only way to interact with the UI currently
+is through the `Interaction` and `FocusPolicy` components.
+`Interaction` only supports mouse or touch-based input.
+We want gamepad supports and more game-oriented functionalities,
+such as menu trees and movement between menus.
+
+The current UI focus system is limited to `bevy_ui` and
+can't be swapped out for something the user deems more appropriate.
+
+This is in contradiction with the general philosophy of bevy,
```
alice-i-cecile, line 34: IMO we can just cut this paragraph.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+which is of providing to the user all the tools necessary to create something
+that fits their needs optimally.
+
+This RFC proposes to resolve all those questions.
```
alice-i-cecile, line 38: Filler paragraph; cut.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+[`bevy-ui-navigation`] provides a highly customizable ECS-based navigation engine.
+
+## User-facing explanation
```
alice-i-cecile, line 41: I think this will probably want a rewrite, but for now the bullet points are fine.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+  to go from the last focus position to the new one.
+- ui-navigation provides default input systems, 
+  but it's possible to disable them and instead use custom ones.
+  This is why we restrict interactions with navigation to events.
+- ui-navigation provides a system to declare buttons as "leading to menu X"
+  in order to create menu trees.
+- ui-navigation defines [custom `SystemParam`][event_helpers]s
+  to ease reaction to events (such a button press or menu change).
+  The most naive usage results into [elm architecture][elm_architecture]-like code.
+  Systems updating the model (ECS hierarchy) and systems reacting to events.
```
alice-i-cecile, line 57: Grammar is a bit weird here.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+  You can specify a focusable in any other menu that will lead to this
+  one once activated.
+
+## Implementation
+
+Since `bevy-ui-navigation` already exists,
+we will discuss its design, it's functionalities and their uses.
+
+Implementation details are already covered by the implementation,
+which btw is fully documented (including internal architecture).
```
alice-i-cecile, line 71:
```suggestion
which is fully documented (including internal architecture).
```

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+- An optional menu tree management system
+
+## Terminology
+
+- _menu tree_: the access hierarchy of menus.
+- _activating_ a focusable: Sending the `Action` event while this focusable is focused.
+- _focused menu_: The menu in which the currently focused `Focusable` is.
+- _child menu_ (of a focusable): The menu accessed from this focusable.
+
+## Motivation
```
alice-i-cecile, line 22: The other big missing piece of motivation here is screenreader integration and accessibility.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+we will discuss its design, it's functionalities and their uses.
+
+Implementation details are already covered by the implementation,
+which btw is fully documented (including internal architecture).
+
+### Exposed API
+
+The crux of `bevy-ui-navigation` are the following types:
+- `Focusable` component
+  - `lock`: activating it will lock the navigation engine and discard all `NavRequest`.
```
alice-i-cecile, line 77:
```suggestion
  - `lock`: locks the navigation engine and discards all `NavRequest`.
```

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+  - `cancel`: activating it will do the equivalent of receiving a `Cancel` request.
```
alice-i-cecile, line 78: This feels circular and is very unclear to me.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+  - `dormant`: this will be the preferred entity to focus when entering a menu.
```
alice-i-cecile, line 79: Definitely feels like there's a missing sentence explaining what this actually does.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+Since `bevy-ui-navigation` already exists,
+we will discuss its design, it's functionalities and their uses.
+
+Implementation details are already covered by the implementation,
+which btw is fully documented (including internal architecture).
+
+### Exposed API
+
+The crux of `bevy-ui-navigation` are the following types:
+- `Focusable` component
```
alice-i-cecile, line 76: Can we model this as an enum?

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+  - `lock`: activating it will lock the navigation engine and discard all `NavRequest`.
+  - `cancel`: activating it will do the equivalent of receiving a `Cancel` request.
+  - `dormant`: this will be the preferred entity to focus when entering a menu.
+- `NavMenu` enum
```
alice-i-cecile, line 80: The name isn't very descriptive here; I'm not sure how this is used.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+  - `ScopeMove(ScopeDirection)`: tab menu hotkey
+- `NavEvent` event
+  - `InitiallyFocused`: We went from 0 focused to 1 focused element
+  - `NoChanges`: The `NavRequest` did not change the focus state.
+  - `FocusChanged`: The `NavRequest` changed the focus state,
+    has the list of items that were made inactive
+    and items made active.
+  - `Locked`: The `NavRequest` caused the focus system to lock.
+  - `Unlocked`: Something unlocked the focus system.
+- `MoveParam` trait
```
alice-i-cecile, line 98: This is a odd name. Maybe something like `MenuNavigationStrategy`? It's a niche trait so we can go with something verbose and descriptive.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+  The most naive usage results into [elm architecture][elm_architecture]-like code.
+  Systems updating the model (ECS hierarchy) and systems reacting to events. 
+- It is possible to create isolated menus by using the `NavMenu` component.
+- All `Focusable` children in the `NavMenu` entity's tree will be part
+  of this menu, and moving using a gamepad from one focusable in a menu
+  can only lead to another focusable within the same menu.
+  You can specify a focusable in any other menu that will lead to this
+  one once activated.
+
+## Implementation
```
alice-i-cecile, line 65: Like in the user-facing explanation, this would benefit from more prose laying out how all the bits are connected.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+this requires a way to pick a focusable in a provided direction.
+This is delegated to the UI library implementation
+through the `MoveParam` trait.
+In the case of bevy, `bevy_ui` is the UI library implementation,
+and the `GlobalTransform` is used to locate focusables in regard to each-other.
+
+The `Focusable` component holds state about what happens when it is activated
+and what focus state it is in. (focused, active, inactive, etc.)
+
+#### Note on hovering state
```
alice-i-cecile, line 116: Probably belongs in Drawbacks?

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+    // Any dormant focusable
+    .or_else(any_dormant)
+    // Any focusable
+    .or_else(fallback)
+```
+
+
+### Menu navigation handling
+
+The public API has `NavMenu`, but we use internally `TreeMenu`,
```
alice-i-cecile, line 164: This choice isn't justified well enough. There's a fair bit of complexity here; why not just expose `NavMenu`?

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+this prevents end users from breaking assumptions about the menu trees.
+
+We define the `TreeMenu` component.
```
alice-i-cecile, line 167: This should try to make clear that it ties into the existing `Parent` and `Child` components.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+All `Focusable` direct or indirect children of a `TreeMenu` are part of that menu.
+A `TreeMenu` has a `Focusable` focus parent,
+when that `Focusable` is focused and the `Action` request is received,
+the focus changes to the `TreeMenu`
+
+In short, the navigation tree is built on two elements:
+- The `reachable_from` that a `NavMenu` was spawned with
+- The `Focusable` within the hierarchy of a `NavMenu`
+
+This is a tree in the ECS.
```
alice-i-cecile, line 177: Cut.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+A `NonEmpty<T>` has a `first` method that returns a `T`,
+which in our case would be the last and new focused elements.
+I chose this crate over similar ones,
+because the code is dead simple and easy to vet.
+
+#### Going back to a previous menus
+
+`Focusable`s have a `dormant` state that is set when they go from
+active/focused to not active.
+This is a form of memory that allows focus to go back to the last focused
```
alice-i-cecile, line 210: This is a good feature, but I'm not sure this is the clearest implementation? I think most of my problem here is that the "dormant" name isn't intuitive to me.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+This introduces a single frame lag on menu spawn,
+but it improves ergonomics to users.
+I thought it was a pretty good trade-off,
+since spawning menus is not time critical, unlike input.
+
+
+#### cancel events
+
+When entering a menu, you want a way to leave it and return to the previous menu.
+To do this, we have two options:
```
alice-i-cecile, line 236: Why do we need both options? What's the difference?

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+### Game-oriented assumptions
+
+This was not intended during implementation, but in retrospect, the `NavMenu`
+system is relatively opinionated.
+
+Indeed, we assume we are trying to build a game UI with a series of "menus" and
+that it is possible to keep track of a "trail" of buttons from a "root menu" to
+the deepest submenu we are currently browsing.
+
+The example of a fancy editor with a bunch of dockered widgets seems to break
```
alice-i-cecile, line 294:
```suggestion
The example of a fancy editor with a bunch of docked widgets seems to break
```

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+This is to support re-focusing the last focused element when moving between
+menus.
+The design of this RFC _hinges on_ the whole tree been loaded in ECS while
+executing navigation requests, and would work poorly if menus were spawned and
+despawned dynamically.
+From what I gather from questions asked in the `#help` discord channel,
+I think the design most people come up with naturally is to spawn and despawn
+dynamically the menus as they are traversed, which is incompatible with
+this design.
+However, in practice, in my own games, I worked around it by expanding the
```
alice-i-cecile, line 333: IMO we should use RenderLayers for this instead, rather than fussing with moving the UI camera.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,364 @@
...
+whole UI tree beyond the UI camera view and moving around the camera.
+An alternative would simply to set the `style.display` of non-active menus to
+`None` and change the style when focus enters them.
+This also fixes the already-existing 1 frame latency issue with the
+spawn/despawn design.
+
+Something that automates that could be a natural extension of the exposed API.
+It may also help end users go with the best design first.
+
+### User upheld invariants
```
alice-i-cecile, line 343: Definitely nervous about this.

rfcs/41-ui-navigation.md
alice-i-cecile, line 65: The other edge case I want examined is "multiple cursors", like in Smash Bros character select. This was raised in Discord and it seemed like you had a good fix, but it's not in this RFC.

rfcs/41-ui-navigation.md
nicopap: Interesting ideas. The current design has `Focusable` be two different enums (one enum for the state `FocusState`, and one enum for the "kind" of focusable (lock, cancel, normal)). It might be possible to separate them in two components. I've never considered this.

However, we need some sort of strategy to prevent users from directly mutating the state, because the navigation algorithm relies on having full control over it. It's read only (and for very good reason!)

rfcs/41-ui-navigation.md
nicopap: I love this suggestion. This is a very good name

rfcs/41-ui-navigation.md
nicopap: I spent a lot of time trying to come up with a good name here.

The difficulty I had was to **not** make it sound like it is the opposite of active.
[Here is the whole enum list for reference](https://docs.rs/bevy-ui-navigation/latest/bevy_ui_navigation/enum.FocusState.html)

alice-i-cecile: `PreviouslyActive`?

rfcs/41-ui-navigation.md
nicopap: A "go back to previous menu" button is a very common pattern, and accommodating it is fairly neat.

It's not too hard for users to implement their own back button, but then it requires them to actual understand the low level navigation API we are exposing here.

alice-i-cecile: Seems reasonable :) Explain that context here!

rfcs/41-ui-navigation.md
nicopap: Honestly, not that bad. The two conditions that are absolutely necessary are checked and cause panics with a detailed error message in case the invariants are not held. In any case, if there were not enforced, it would not cause UB, just buggy behavior.

But maybe I can take a closer look at them and see if they are that necessary.

rfcs/41-ui-navigation.md
nicopap: I've now copied most of the API docs from bevy-ui-navigation into this RFC. I'm not a fan (just click the "see doc" link :P) but it has the quality of explaining very well each item of the library.

rfcs/41-ui-navigation.md
nicopap: I've replaced this whole section with links to their corresponding rust doc items, since it's pointless to try to summarize the whole library into a few lines.

rfcs/41-ui-navigation.md
nicopap: I've added a more descriptive "Internal representation" section.

rfcs/41-ui-navigation.md
nicopap: There are now 7 paragraphs of justification in the "Spawning menus" section.

rfcs/41-ui-navigation.md
nicopap: Can be replace with `MenuConfig` or `MenuSettings`. I recall bevy is not very consistent when it comes to `Config`, `Configuration` or `Settings`. I actually alias it to `MenuSettings` in some places in the current implementation.

rfcs/41-ui-navigation.md
nicopap: I went with `Prioritized`. It doesn't tie much with the fact that it was previously active, but in practice, the variant is only used to chose which focusable to focus first in a menu, so the name makes sense.

nicopap:
I decided to copy parts of the bevy-ui-navigation API docs for this. Since those are well thought out.

rfcs/41-ui-navigation.md
nicopap: I rather think `display: Display::None` would be the correct solution (or the simplest to implement for end users). Due to layouting and how it (does not) interact with `Visibility` and `RenderLayers`.

rfcs/41-ui-navigation.md
nicopap: This might be better reserved for a "future possibilities" section. This RFC already skips a few non-necessary/nice-to-have features from `bevy-ui-navigation`, and it's still massive.

alice-i-cecile: Agreed, I'm on board with that.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,848 @@
...
+  \
+  If the player moves to a direction where there aren't any focusables,
+  the focus will "wrap" to the other direction of the screen.
+* **BoundScope**: Non-wrapping scope menu
+  Controlled with `NavRequest::ScopeMove`
+  even when the focused element is not in this menu, but in a submenu
+  reachable from this one.
+* **WrappingScope**: Wrapping scope menu
+  Controlled with `NavRequest::ScopeMove` even
+  when the focused element is not in this menu, but in a submenu reachable from this one.
```
nicopap, line 498: I'm aware this could be replace by two booleans. And I should probably do it.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,842 @@
...
+The parent is just a button in another menu.
+
+It is inconvenient to have to pre-spawn each button to acquire their `Entity` id
+just to be able to spawn the menu you'll get to from that button.
+[`bevy-ui-navigation`] uses a proxy holding both the parent
+and the menu state info.
+
+That proxy component is `MenuSeed`,
+you can initialize it either with the `Name` or `Entity` of the parent focusable of the menu.
+A system will take all `MenuSeed` components and replace them with a `TreeMenu`.
```
nicopap, line 516: Note that the current implementation has simplified this. `MenuSetting` is now a component, and we have another `MenuBuilder` component that still gets converted by the navigation plugin into a `TreeMenu`.

Issue is I don't like "`MenuBuilder`" and wait to have a better name before updating the RFC.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,842 @@
...
+## Terminology
+
+- _menu tree_: the access hierarchy of menus.
+- _activating_ a focusable: Sending the `Action` event while this focusable is focused.
+- _focused menu_: The menu in which the currently focused `Focusable` is.
+- _child menu_ (of a focusable): The menu accessed from this focusable.
+- _transitive_: If `A` is _foo_ of `B` (swap _foo_ with anything such as
+  _child_, _left_ etc.)
+  Then `A` is _transitive foo_ of `B` if there is any chain of `C`, `D`,
+  `E` etc. such as `A` is _foo_ `C`, `C` is _foo_ `D`, ... , `E` is _foo_ `B`.
```
nicopap, line 24: I'm so sorry. I hope this is a good enough explanation for everyone :D

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,859 @@
...
+    /// reachable from this one.
+    pub scope: bool,
+}
+```
+
+#### Going back to a previous menus
+
+`Focusable`s have a `prioritized` state that is set when they go from
+active/focused to not active.
+This is a form of memory that allows focus to go back to the last focused
```
Weibye, line 500: I think we need the ability for the developer to define: 

"When coming back to this menu, should we focus on the last-focused element or should we focus on the initial default element"

There are use cases in apps (filling in a new form) or games where we want to treat a menu as "first-time-user-visits" even though the user has visited before. The FFX battle system also springs to mind where you can choose between these two modes, which has a profound effect on muscle-memorisation.

rfcs/41-ui-navigation.md
Weibye: IMO there are much move severe invariants already in the engine, which to a some degree is needing to be upheld by the user (GlobalTransform + Transform for hierarchy to work comes to mind). 

That in itself is not a good argument for adding more, but we are going to need a good solution for Archetype Invariants engine-wide anyway which this design would benefit from so I'm less nervous about this. 

That being said, there is plenty of value is exploring alternatives but I don't think this should be that hard of a blocker in the name of getting some velocity on core UI features.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,859 @@
...
+Such a move would generate the following `NavEvent`:
+```rust
+NavEvent::FocusChanged {
+  to: [<abc entity>],
+  from: [<B entity>, <abc entity>],
+}
+```
+The "tabs menu" is defined as a "scope menu", which means that
+by default, the `LT` and `RT` gamepad buttons will navigate
+the "tabs menu" regardless of the current focus position.
```
Weibye, line 335: I'm guessing this is just for illustrative purposes, but we should aim to separate button inputs from action events, which would be out of scope for this RFC (and I suspect @alice-i-cecile's input manager is making a good job of that) but could this be example be rewritten to use something like `NavRequest::Next` and `NavRequest::Previous` (as suggested above)?

Weibye (commented):
This is a good and well thought out design! I like the decoupling of everything and I think that is important for a robust and flexible design.

I do think there should be a `NavRequest::Next` and `NavRequest::Previous` (or something similar) to facilitate behaviours like tab-cycling through a menu
  - A user should be able to press next (tab) or previous (shift + tab) to bring focus to the different focusable elements within the menu
  - A developer should be able to define the order the elements should be cycled through
  - This _could_ be completely unrelated to the geometric layout of the elements
  - I'm pretty sure this is an important feature for general accessibility and screen-reader support

Weibye:
Imagine this skill tree. I'm suspecting that just using `NavEvent::Move(direction)` would allow the user to freely navigate between nodes as if they were not connected by this hierarchy. Would this design allow me to define navigation so that the user is forced to navigate along the white path?

image

Source: Star Wars Jedi: Fallen Order

rfcs/41-ui-navigation.md
nicopap: That's something I overlooked. Current design doesn't allow the user to modify the `prioritized` focusable. I think this can be enabled by providing a custom `SystemParam`.

nicopap:
> I do think there should be a NavRequest::Next and NavRequest::Previous (or something similar) to facilitate behaviours like tab-cycling through a menu

Strongly agree, will add that to the design. Though I need to think what's the best way to let users specify ordering without being a burden.

> Imagine this skill tree. I'm suspecting that just using NavEvent::Move(direction) would allow the user to freely navigate between nodes as if they were not connected by this hierarchy. Would this design allow me to define navigation so that the user is forced to navigate along the white path?

Yes. The default navigation option is very rudimentary and only cares about position, but you can disable it to replace with your own strategy (`MenuNavigationStrategy` trait)

nicopap:
@Weibye On second though, I think the tab cycling feature should wait. I'll add it to the "future" section. The PR is massive as it is, and I don't want to pile on functionalities. It's already going to be difficult to review as-is.

There is a few features that I decided to delegate to future endeavor. What I will do is that if the ui navigation PR is merged, I'll immediately open an issue for each functionalities I left behind (and a tracking issue on top). This also allows more piecemeal PRs that are better for people to familiarize themselves with the ui-navigation code (I especially want the bus factor on ui-navigation to increase over 1 if possible) Is this fine for you?

Weibye:
> On second though, I think the tab cycling feature should wait. I'll add it to the "future" section. The PR is massive as it is, and I don't want to pile on functionalities. It's already going to be difficult to review as-is.
> 
> There is a few features that I decided to delegate to future endeavor. What I will do is that if the ui navigation PR is merged, I'll immediately open an issue for each functionalities I left behind (and a tracking issue on top). This also allows more piecemeal PRs that are better for people to familiarize themselves with the ui-navigation code (I especially want the bus factor on ui-navigation to increase over 1 if possible) Is this fine for you?

Yup, perfectly fine :) And this should probably be our preferred approach when dealing with large features like this. 

> (I especially want the bus factor on ui-navigation to increase over 1 if possible)

Could you elaborate what you mean by this?

nicopap:
I currently am the only person who really understands the code for ui navigation. I hope I wrote enough (internal) docs + this RFC for other people to be able to understand and contribute, but at lest for a short period I'll probably the only one confident enough to modify the navigation code. I want this to be as short a period as possible.

rfcs/41-ui-navigation.md
```diff
@@ -0,0 +1,872 @@
...
+
+### How does this work with the spawn/despawn workflow?
+
+The current design imposes to the user that all UI nodes such a buttons and
+menus must be already loaded, even if not visible on screen.
+This is to support re-focusing the last focused element when moving between
+menus.
+The design of this RFC _hinges on_ the whole tree been loaded in ECS while
+executing navigation requests, and would work poorly if menus were spawned and
+despawned dynamically.
```
Weibye, line 827: My mind keeps coming back to this statement. It is an extremely common use-case to not know the data ahead of time: I want to select a villager from a list. The game queries the list of villagers currently available, and spawns buttons for them so that I can click on one. Or just a search bar in an application.

Does this mean that we cannot support those use cases at all?

alice-i-cecile: Agreed, now that you highlight this I share your concerns.

nicopap: So. To  be precise, this section is really meant to contrast with the pattern of spawning and despawning each menu as they are created.

The limitation here (I've not tested it) I think is only when despawning the parent menu (for example despawning the main menu when entering the options menu), not when creating dynamically new menus you can enter, or despawning menus that do not have focus.

When you go back one level of menus (pressing backspace or clicking a "back" button), the navigation algo reads the `Entity` of the button the menu is "reachable from" (aka the button that activated the menu), and if that `Entity` is despawned, probably bad things happen (panic! most likely), as we now sort of "dereference a dangling `Entity`".

However, the reverse case of creating new menus at runtime is widely supported, the `MenuBuilder::ParentEntity` enum variant should help with runtime spawning of new menus, since you can just give it whatever `Entity` with a `Focusable` component you have available. It does mean however that the submenus must be created before the button is activated (for example when it gets focused/hovered).

It is not necessary for the whole tree to be loaded at once, but rather than the path back to the root menu is. For support for giving back focus to the last focused element (prioritized) you would also need to keep the `Focusable`s and menu tree of already-explored menus loaded. But typically, for a dropdown, this isn't something people expect.

This is why I mentioned an "infinitely deep recursive menu" as an edge case ui-navigation couldn't support.

I'm currently making a sort of "infinite weapon upgrade menu" as a cool example to demonstrate how easy it is to use your own navigation methods (using `Sprite` and `Text2d`) and generating new menus on the fly.

nicopap: When it comes to creating new `Focusable`s in an existing menu, it should be alright. But it's always going to be difficult when removing them (beside the mandatory minimum 1 `Focusable` in a menu). `TreeMenu` has a `active_child` field for avoiding iterating all children, but it completely violates the single-source-of-truth principle which makes despawning `Focusable`s dangerous (you must not despawn a focused or active focusable)

I'll probably have to rework access to the `active_child` to fallback to the ECS hierarchy when it doesn't exist anymore.

Weibye: That is good to know, and still impressive work with all of this. Is the last remaining hurdle then to find a good way to remove _leaf_ nodes in the navigation tree dynamically?

The requirement that the path back to root must exist I think is acceptable as long as we can dynamically remove leaf nodes

nicopap:
This needs to be split into three different changes

1. Event-based interactions
2. simple gamepad-based  UI controls
3. menu navigation.

It's simply not feasible to integrate such a huge chunk of changes in bevy without more fine grained reviews.

I'll close this RFC. I've no short term plans on opening the 3 RFCs the split would require.

PR #43, opened 2021-11-15
Many Worlds

alice-i-cecile:
[**RENDERED](https://github.com/alice-i-cecile/rfcs/blob/many-worlds/rfcs/43-many-worlds.md)

Building on the ideas introduced in #16 by @NathanSWard (and the associated thread), this RFC presents a general-purpose multiple worlds API.

The TL;DR:

1. The app contains many worlds, each with their own `WorldLabel`.
2. Each schedule is associated with a `WorldLabel`.
3. By default, all schedules run in parallel, then a special global schedule is run, then `AppCommands` are applied.
4. `AppCommands` can be used to send data between worlds, move entities, make new worlds, clone schedules and so on.

This should be useful for:
- cleaning up pipelined-rendering logic
- rollback networking
- simulation with varying hyperparameters
- sharding / chunking the game world
- much, much more!


rfcs/43-many-worlds.md
```diff
@@ -0,0 +1,508 @@
...
+
+## Rationale and alternatives
+
+### Why not sub-worlds?
+
+Ownership semantics are substantially harder to code and reason about in a hierarchical structure than in a flat one.
+
+They are significantly less elegant for several use cases (e.g. scientific simulation) where no clear hierarchy of worlds exists.
+
+Things will also get [very weird](https://gatherer.wizards.com/pages/card/details.aspx?name=Shahrazad) if your sub-worlds are creating sub-weorlds (as is plausible in an AI design), compared to a flat structure.
```
Weibye, line 439:
```suggestion
Things will also get [very weird](https://gatherer.wizards.com/pages/card/details.aspx?name=Shahrazad) if your sub-worlds are creating sub-worlds (as is plausible in an AI design), compared to a flat structure.
```
Typo?

rfcs/43-many-worlds.md
```diff
@@ -0,0 +1,508 @@
...
+
+### Why are schedules associated with worlds?
+
+Each system must be initialized from the world it belongs to, in a reasonably expensive fashion.
+Free-floating schedules,
```
Weibye, line 444: Seems you have half a sentence here 😄

Weibye (commented):
Did a pass on typos :)

Weibye:
Would this help making a future bevy editor a reality? Having an `editorworld` and a `gameworld` and having the editor interact between the two?

alice-i-cecile:
> Would this help making a future bevy editor a reality? Having an editorworld and a gameworld and having the editor interact between the two?

Oh, that's a neat architecture. I haven't dug deep enough into that problem space to say for sure, but I suspect it's worth exploring.

rfcs/43-many-worlds.md
```diff
@@ -0,0 +1,522 @@
...
+   for i in 0..N_WORLDS {
+      // Create a new world
+      app.spawn_world(SimulationWorld(i))
+         // Sets the currently active world for the App,
+         // which controls the schedule to which systems are added
+         // and where resources are inserted
+         .set_world(SimulationWorld(i))
+         // This ensures basic functionality like `Time` functions properly
+         // for each of our worlds
+         .add_plugins(MinimalPlugins)
```
TheRawMeatball, line 112: Would this even be necessary? I'm assuming something like `Time` would be in a global in such an architecture.

TheRawMeatball (commented):
I'm a fan of the multiple-schedule parts, but I feel like a combination of better control over task scheduling and dedicated thin api over spawning tasks for this use case could work better.

rfcs/43-many-worlds.md
alice-i-cecile: Hmm. I think you're right.

alice-i-cecile:
I have a basic proof of concept for 0.5 for the scientific simulation case [here](https://github.com/alice-i-cecile/janky-multiple-worlds). It's very limited in features, but should work well for that narrow use case.

bonsairobo:
I'm not sure how useful the `App` abstraction is for the kinds of use cases that benefit from many-worlds. I think the scheme described here is overly prescriptive:

> During each pass of the game loop, each schedule runs in parallel, modifying the world it is assigned to. These worlds are fully isolated: they cannot view or modify any other world, and stages progress independently across schedules. Then, the global world (which stores common, read-only data) is processed, and any systems in its schedule are run. Finally, at the end of each game loop, all of the worlds wait for a synchronization step, and apply any AppCommands that were sent.

Why require that the schedules run in parallel? Could we instead allow users to decide when and how schedules are run?

What if instead we just had a low-level API for running a single frame on a given `(Schedule, World, TaskPool)`? Then users could manage their own orchestration of schedules and worlds. And if the worlds are truly isolated, then we can just rely on the compiler to ensure that there aren't multiple schedules mutating the same world in parallel.

TheRawMeatball:
Here's a task-pool based design for multi-world I had, which can be implemented fully user-side
```rs
fn add_world_to_app<P, F>(app: &mut App, mut schedule: Schedule, mut sync_fn: F)
where
    P: SystemParam,
    F: for<'w, 's> FnMut(&mut World, <P::Fetch as SystemParamFetch<'w, 's>>::Item),
    for<'w, 's> <P::Fetch as SystemParamFetch<'w, 's>>::Item: SystemParam<Fetch = P::Fetch>,
    F: Send + Sync + 'static,
{
    let (task_to_world_tx, task_to_world_rx) = async_channel::unbounded();
    let (world_to_task_tx, world_to_task_rx) = async_channel::unbounded();

    let task = async move {
        let mut world = Some(World::new());
        let tx = task_to_world_tx;
        let rx = world_to_task_rx;

        loop {
            schedule.run(world.as_mut().unwrap());
            tx.send(world.take().unwrap()).await.unwrap();
            world = Some(rx.recv().await.unwrap());
        }
    };

    let system = move |system_param: <P::Fetch as SystemParamFetch<'_, '_>>::Item| {
        let tx = &world_to_task_tx;
        let rx = &task_to_world_rx;
        if let Ok(mut world) = rx.try_recv() {
            sync_fn(&mut world, system_param);
            tx.try_send(world).unwrap();
        }
    };

    app.add_system(system);
    app.world
        .get_resource::<AsyncComputeTaskPool>()
        .unwrap()
        .spawn(task)
        .detach();
}

fn main() {
    let mut app = App::new();
    // the turbofish shouldn't be necessary, but it is. blame rustc
    add_world_to_app::<Res<u32>, _>(&mut app, Schedule::default(), |world, res| {
        // this closure is ran in a system in the main world.
        // and here you have access to the entire subworld and any resources / queries you want from the main world.
        // this function runs once per subworld tick for synchronization
        // and a subworld tick does not block main world progress
    })
}
```

bonsairobo:
Hey sorry I left a comment and then deleted it because I realized that I hadn't fully grokked the RFC when I wrote it, and I want to leave more coherent feedback. I like the general direction of the RFC.

I'm thinking about how it will address my particular use case of running a discrete time simulation where I can manually call `Schedule::run_once`. One obvious blocker is that the executors assume there is a `ComputeTaskPool` on whatever `World` we use to run the schedule, but we should prefer to use a global pool; I believe this is properly addressed by the RFC.

Then, for example, let's say I want to run my schedule every time a specific key is pressed. Doing this from a custom runner might work, although I don't think the RFC proposed a specific interface for custom runners. If we just have exclusive access to the `App`, then I could manually inspect the global input events and then run the schedule.

But I also think it might be nicer if I could run my schedule within another system via `ResMut<(Schedule, World)>`. Then so long as my system can access the global task pool, I have a more ergonomic interface for accessing resources.

> Finally, the ergonomics, discoverability and potential for optimization of a custom-built API for this very expressive feature is much better than Res<World>-based designs.

Could you expand on this?

rfcs/43-many-worlds.md
```diff
@@ -0,0 +1,522 @@
...
+
+`AppCommands` directly parallel `Commands` API and design where possible, likely abstracting out a trait.
+
+They operate on `&mut App`.
+
+The initial set of `AppCommands` needed for this RFC are:
+
+- `spawn_world(label: impl WorldLabel)`
+- `despawn_world(label: impl WorldLabel)`
+- `set_world(label: impl WorldLabel)`
```
m0nkfish, line 376: i am wary of builder APIs which change the meaning of downstream methods in this way. usually i would go with something like

```rust
app.with_world(label, |&mut app| app.add_plugin(MyPlugin))
```

but i maybe there are backwards compat concerns?

SamPruden:
There are two things that jump out to me as being desirable here, although I'm not sure whether either is feasible.

1. It would be cool to be able to create a world in which all components and resources are statically typechecked to implement some trait (or collection of traits), including the ability to safely enumerate them as `dyn Trait`. Hiding worlds behind wrappers that enforce this is one option, but we would somehow need to deal with the fact that `.set_world()` ... `insert_resource()` would bypass that. This is motivated by https://github.com/bevyengine/bevy/issues/3877, but it's also conceivable that a user may want to do this with a custom trait on a custom world. It's not obvious to me that Rust's typesystem can support this.
2. If people are dealing with multiple worlds, I predict a common class of bug where an `Entity` is used in the wrong world. We could maybe do something with `'world` lifetime magic here, but maybe that would add undesirable complexity to `Entity` in other places. Would this even be reliably caught at runtime in the current proposal?

ottworks:
Does this RFC support simulating different worlds at different update rates?

mockersf:
is this still relevant now that we have sub apps with their own world?

alice-i-cecile:
> Does this RFC support simulating different worlds at different update rates?

Currently no. I'm not fully happy with the current state of this RFC; moving it to draft until I have some cycles to devote to it.

> is this still relevant now that we have sub apps with their own world?

Very. This is / was an attempt to try and wrangle the API and complexities of the subapps to create a coherent design.

rfcs/43-many-worlds.md
```diff
@@ -0,0 +1,522 @@
...
+
+      // Making a simulation world that's identical to the main world in all relevant ways
+      app_commands
+         .spawn_world(sim_world)
+         .clone_schedule(GameRulesWorld, sim_world)
+         // As above, blocked by bevy #1515
+         .clone_entities_to_world(sim_query.entities(), sim_world)
+         // This system will evaluate the world state, send it back with an event reporting the values
+         .add_system(evaluate_action_system.to_world(sim_world))
+         .send_event(proposed_action, sim_world);
```
LogicFan, line 269: Currently, we use `EventWriter` to write events, here we can also you app_commands to send events, what will be the difference?

Also, it's possible that one system in the world wants to broadcast the event to all world since it does not know which world should respond to it. How this situation should be handled?

MDeiml:
This seems to be a quite hard problem even without sharing entity namespaces. See https://github.com/bevyengine/bevy/issues/3096 for why `AppCommands::spawn` basically can't return an `Entity`. As I see it there are 3 solutions to this:
1. Make `Entities` thread safe, e.g. put it behind a `RwLock`. This should be ok as there should only be one `write` per cycle to flush reserved entities, so most of the time aquiring the lock shouldn't block.
2. Don't run schedules in parallel with worlds they need to access the entity namespace of. Also works fine, but the implementation might get pretty ugly as the execution model suddenly becomes very complicated. This is the approach that `RenderStage::Extract` is taking right now
3. Have some intrinsic entity mapping in `AppCommands`. E.g. have an extra namespace for entities spawned in other worlds and keep a `HashMap` of relations that is updated when `AppCommands` are applied. This would probably be the easiest to implement, but it doesn't solve the problem that `AppCommands` can't know about entity references stored in component, so developers will have to write some manual mapping for them

Personally I like 1. the most. With some effort we might even get around using `RwLock`s or similar if ownership of `Entities` is given to the `App` or global world. `Entities` already supports this "mutate in parallel, then flush synchronously" flow.

MDeiml:
Proposal what to add to the RFC:

`Entities` will no longer be stored in `World`. Methods in `World` that need access to `&Entities` will take it as a parameter.

For convenience a `Universe` (I like that name :sweat_smile:) struct is added, which is a wrapper around `World` and `Entities`. It will support the same interface that `World` supports now (no `Entities` as method parameters). The global `World` is instead a global `Universe`.

Alongside a list of worlds `App` will also contain `namespaces: Vec<Entities>` as well as `namespace_map: HashMap<WorldId, usize>`, which determines which namespace to use for each world.

All in all execution of worlds will not be impacted. During the "global step" entities are flushed.

Pros:
* No performance impact
* Developers can't mess up when copying / moving components to other worlds

Cons:
* More difficult API, almost duplicate interfaces of `World` and `Universe`
* Code gets ugly as `&Entities` is passed around
* Developers can still mess up if worlds don't share a namespace

SamPruden:
> Exclusive systems for example will get a &mut Universe instead of &mut World

Is this overly restrictive in terms of blocking all worlds at once? I can imagine scenarios where one small world may want to exist with an exclusive system that doesn't block all of the other worlds. Would it still be feasible to have world-exclusive as well as universe-exclusive systems?

It's been a while since I looked at this part of the code, so I don't have any technical feedback at this time.

MDeiml:
`Universe` would also only have a `&Entity` not `&mut Entity` or `Entity`. So it's non-blocking with regard to other worlds. This should pretty much be ok since the only thing you *can't* do without a mutable reference to `Entities` is flush and delete entities, both of which would be done during the global sync step.

rfcs/43-many-worlds.md
```diff
@@ -0,0 +1,522 @@
...
+## User-facing explanation
+
+When you need to fully isolate entities from each other (or make sure logic doesn't actually run on some chunk of them), you can create **multiple worlds** in your Bevy app.
+
+At this point, you should be familiar with the basics of both `Worlds` and `Schedules`.
+Your `App` can store any number of these, each corresponding to a particular `WorldLabel`.
+
+During each pass of the game loop, each schedule runs in parallel, modifying the world it is assigned to.
+These worlds are fully isolated: they cannot view or modify any other world, and stages progress independently across schedules.
+Then, the **global world** (which stores common, read-only data) is processed, and any systems in its schedule are run.
```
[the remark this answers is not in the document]
MDeiml: I think global resources are needed. Something like `AssetServer` or `WgpuSettings` would belong there. Global entities are not a good idea I think, so maybe it doesn't need to be a `World`. A global schedule would be good tho. This would be the place to put systems that are exclusive over all worlds.

rfcs/43-many-worlds.md
```diff
@@ -0,0 +1,522 @@
...
+2. Entity staging grounds, where complete or nearly-complete entities can be stored for quick later use (see [bevy #1446](https://github.com/bevyengine/bevy/issues/1446)).
+3. Entity purgatories, where despawned entities can be kept around while their components are inspected (see [bevy #1655](https://github.com/bevyengine/bevy/issues/1655)).
+4. A trivially correct structure for parallel world simulation for scientific simulation use cases.
+5. Look-ahead simulation for AI, where an agent simulates a copy of the world using the systems without actually affecting game state.
+6. Isolation of map chunks.
+7. Temporarily disabling entities in a system-agnostic fashion.
+
+Currently, this flavor of design requires abandoning `bevy_app` completely, carefully limiting *every* system used (including base / external systems) or possibly some very advanced shenanigans with exclusive systems.
+
+A more limited "sub-world" design (see #16 for further discussion) is more feasible, by storing a `World` struct inside a resource, but this design is harder to reason about, unnecessarily hierarchical and inherently limited.
```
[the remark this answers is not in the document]
MDeiml: I think the main point here is that the hierarchic structure of `App -> Subapp -> World -> Subworld` is overly complicated. Also this expands a lot more than #16 and standardizes communication between worlds.

lwansbrough:
@MDeiml 

> But it would be nice if worlds could just share the same "entity namespace", i.e. related entities just have the same id. This is how the relation between the main and render world works right now, which I think is a very valuable example for anything that works with multiple worlds. I'm not sure though if this implementable without changing very fundamental ECS code, the `Entities` struct basically needs to be thread safe I think.

I think this is an implementation detail per application. For instance, we might be running a "network replication" world where entities are spawned on the server. We can't assume that the client has total authority over the generation of entity IDs.

rfcs/43-many-worlds.md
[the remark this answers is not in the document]
alice-i-cecile: Yep, exactly!

alice-i-cecile:
Closing for now: this needs more thought in the context of what we've learned about pipelined rendering and I want to let others try their hand at a design for this!

PR #44, opened 2021-11-17
Removed query filter

alice-i-cecile:
[**RENDERED**](https://github.com/alice-i-cecile/rfcs/blob/removed-filter/rfcs/44-removed-filter.md)

Addresses https://github.com/bevyengine/bevy/issues/2148 by proposing a `Removed<C>` query filter.

This works exactly as you might hope, using reliable change detection semantics and mechanisms.
Full credit to @maniwani (Joy) for the core idea of just storing another component with the change tick. 

Some notes:
1. We can't fully remove `RemovedComponents`, since this doesn't work for despawned entities, since they no longer exist to be queried for.
2. The current design is opt-in, using the machinery from #27, to avoid pointless performance overhead.

### DIFF
--- a/rfcs/44-removed-filter.md
+++ b/rfcs/44-removed-filter.md
@@ -1,0 +1,87 @@
+# Feature Name: `removed-filter`
+
+## Summary
+
+`Removed<C>` can be used as a query filter, with reliable change detection behavior.
+Users must opt-in to enabling this by configuring the `Component` trait for `C`.
+
+## Motivation
+
+Removal detection is a powerful tool for ensuring the validity of game state, but the existing `RemovedComponents` system parameter is unwieldy (see the motivating [bevy #2148](https://github.com/bevyengine/bevy/issues/2148)).
+
+It cannot be easily composed with other queries, has an unexpected API, and only persists for two frames, unlike reliable change detection.
+
+## User-facing explanation
+
+In addition to `Changed` and `Added` query filters, you can also use `Removed` query filters.
+As you may expect, a query filter for `Removed<Asleep>` will filter the list of entities returned by your query to only include entities which have had a component of the type `Asleep` removed from them since the last time this system ran.
+
+For performance reasons, this behavior is opt-in and attempts to use removal detection without having opted-in will result in a run-time panic.
+You can enable it by configuring the `Component` trait.
+In practice, this will generally be done like so:
+
+```rust
+#[derive(Component)]
+#[component(removal_detection = "true")]
+struct Asleep;
+```
+
+However, be mindful that queries can only return entities which currently exist.
+If you need to respond to component removal of any kind, including that caused by entity despawning (as you might if you were attempting to maintain a valid entity graph), use the `RemovedComponents` system parameter instead.
+
+Note that this system parameter returns a list of entities which must be manually handled, and only tracks removals that occur over the past two game loops.
+As a result, you should prefer the `Removed` query filter where possible.
+
+## Implementation strategy
+
+Steps:
+
+- add a `LastRemoved<C: Component>` marker component, which stores the `change_tick` of the world at the time of entity removal
+- add an associated `RemovalDetection` bool to the component trait
+- if `RemovalDetection == true`, the `remove` method on `World` also inserts an appropriate `LastRemoved<C>` component to that entity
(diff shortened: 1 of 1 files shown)


rfcs/44-removed-filter.md
```diff
@@ -0,0 +1,87 @@
+# Feature Name: `removed-filter`
+
+## Summary
+
+`Removed<C>` can be used as a query filter, with reliable change detection behavior.
+Users must opt-in to enabling this by configuring the `Component` trait for `C`.
```
djeedai, line 6: Mention not necessary anymore?

rfcs/44-removed-filter.md
```diff
@@ -0,0 +1,87 @@
...
+Removal detection is a powerful tool for ensuring the validity of game state, but the existing `RemovedComponents` system parameter is unwieldy (see the motivating [bevy #2148](https://github.com/bevyengine/bevy/issues/2148)).
+
+It cannot be easily composed with other queries, has an unexpected API, and only persists for two frames, unlike reliable change detection.
+
+## User-facing explanation
+
+In addition to `Changed` and `Added` query filters, you can also use `Removed` query filters.
+As you may expect, a query filter for `Removed<Asleep>` will filter the list of entities returned by your query to only include entities which have had a component of the type `Asleep` removed from them since the last time this system ran.
+
+For performance reasons, this behavior is opt-in and attempts to use removal detection without having opted-in will result in a run-time panic.
```
djeedai, line 19: Same here.

rfcs/44-removed-filter.md
```diff
@@ -0,0 +1,87 @@
...
+- add a `LastRemoved<C: Component>` marker component
+- add an associated `RemovalDetection` bool to the component trait
+- if `RemovalDetection == true`, the `remove` method on `World` also inserts an appropriate `LastRemoved<C>` component to that entity
+- modify `RemovedComponents` to only update when this method is called, unifying the opt-in
+- add a `Removed<C: Component>` query filter using the `WorldQuery` trait with `Fetch: FilterFetch`, which fetches the value of `LastRemoved` and appropriately filters entities using the same logic as `Changed` or `Added`
+  - under the hood, this uses the `Added` change ticks of the data-less `LastRemoved<C>` component, which saves us a `u32` per component of duplicated data
+
+## Drawbacks
+
+1. Users must opt-in to removal detection.
```
djeedai, line 48: Is this still true now that all components require `#[derive(Component)]`?

djeedai (changes requested):
That looks like a very useful addition considering the cheat book had to add [a dedicated page just for this](https://bevy-cheatbook.github.io/programming/removal-detection.html).

But the PR looks stale; could you please update @alice-i-cecile to take into account recent changes like the mandatory `#[derive(Component)]`? Thanks!

rfcs/44-removed-filter.md
alice-i-cecile: Yes: the design here is that the performance cost is too high to enable by default.

alice-i-cecile:
It was actually written with `derive(Component)` in mind, but yes, this is a little stale. Let me take a look :)

rfcs/44-removed-filter.md
```diff
@@ -0,0 +1,87 @@
...
+## Summary
+
+`Removed<C>` can be used as a query filter, with reliable change detection behavior.
+Users must opt-in to enabling this by configuring the `Component` trait for `C`.
+
+## Motivation
+
+Removal detection is a powerful tool for ensuring the validity of game state, but the existing `RemovedComponents` system parameter is unwieldy (see the motivating [bevy #2148](https://github.com/bevyengine/bevy/issues/2148)).
+
+It cannot be easily composed with other queries, has an unexpected API, and only persists for two frames, unlike reliable change detection.
```
Aceeri, line 12: Should note that it doesn't actually persist for two frames which makes it more unwieldy.

maniwani:
One more question this RFC should address is, "What happens to `LastRemoved<T>` if you insert `T` again?"

Also, I'm thinking there should be a way to rig this internally so that the options are:
- no removal detection
- ticks only 
- ticks + data

Also if the strategy stays as described, I think "no removal detection" should be default if you can't auto-detect it from queries, since `LastRemoved<T>` being a component will fragment archetypes.

PR #45, opened 2022-01-14
Stageless: a complete scheduling overhaul

alice-i-cecile:
[**RENDERED**](https://github.com/alice-i-cecile/rfcs/blob/stageless/rfcs/45-stageless.md)

This RFC is now ready for final review. Feedback is welcome from users at all levels! Do you think this design easy-to-understand, useful and correct? How could the proposed API be improved?

## TL; DR

- Systems: now stored in the `World`
- Exclusive systems: no longer special-cased
- System sets: now just pure metadata, define order and conditions of groups of systems
- Schedules: now the executable instances of system sets, also stored in the `World`
- Run criteria: replaced by combineable, non-mutating, `bool`-returning functions
- States: now powered by new exclusive system + schedule combo instead of run criteria
- Fixed timestep: now powered by new exclusive system + schedule combo instead of run criteria
- Commands: now applied with an exclusive system
- Stages: yeet
- All of the awful bugs with stages, states, and run criteria: replaced with new, yet-to-be-discovered bugs!

Plus, some shiny new toys: running systems in commands and modifying schedules at runtime!

## Prototype

[iyes_loopless](https://github.com/IyesGames/iyes_loopless) is a Bevy 0.6-compatible third-party implementation of the core ideas of how this RFC handles run criteria and states. Go try it out!

## Context

This is the culmination of many months of discussion, work, debate and mind-melting from @maniwani (Joy), @hymm (WrongShoe) and myself; they should be considered equal coauthors on this RFC. If you're curious about the background, check out https://github.com/bevyengine/bevy/discussions/2801 to get a sense of the tangled mess we're trying to clean up.

Special thanks to @Ratysz and the rest of the ECS crew for all of the feedback and guidance throughout the process.

## Meta
What is the best way to handle the migration process from an organizational perspective?
  - Get an RFC approved, and then merge a massive PR developed on a off-org branch?
    - Straightforward, but very hard to review
    - Risks divergence and merge conflicts
  - Use an official branch with delegated authority?
    - Easier to review, requires delegation, ultimately creates a large PR that needs to be reviewed
    - Risks divergence and merge conflicts
  - Develop `bevy_schedule3` on the `main` branch as a partial fork of `bevy_ecs`?
    - Some annoying machinery to set up
    - Requires more delegation and trust to ECS team
    - Avoids divergence and merge conflicts
    - Clutters the main branch

PR #47, opened 2022-01-20
Port explanations from Stageless RFC

alice-i-cecile:
[**RENDERED**](https://github.com/alice-i-cecile/rfcs/blob/if-needed-ordering/rfcs/47-if-needed-ordering.md)

Extracted from #45.

## Summary

If-needed ordering constraints allow for safer, non-blocking constraints betweeen groups of systems.
If they would have no effect, they're completely ignored.
If they have an effect, they behave identically to strict ordering constraints.

## Motivation

The stageless rework ([RFC #45](https://github.com/bevyengine/rfcs/pull/45)) allows users to configure entire labels, creating ordering dependencies between entire subgraphs.
However, in many cases, pairs of systems ordered in this way have no logical connection between them (and do not even access the same data).

The current ordering, called strict ordering, will still force these pairs of systems to run according to the specified ordering, pointlessly restricting scheduling flexibility and thus performance.

### DIFF
--- a/rfcs/47-if-needed-ordering.md
+++ b/rfcs/47-if-needed-ordering.md
@@ -1,0 +1,173 @@
+# Feature Name: `if-needed-ordering`
+
+## Summary
+
+If-needed ordering constraints allow for safer, non-blocking constraints betweeen groups of systems.
+If they would have no effect, they're completely ignored.
+If they have an effect, they behave identically to strict ordering constraints.
+
+## Motivation
+
+The stageless rework ([RFC #45](https://github.com/bevyengine/rfcs/pull/45)) allows users to configure entire labels, creating ordering dependencies between entire subgraphs.
+However, in many cases, pairs of systems ordered in this way have no logical connection between them (and do not even access the same data).
+
+The current ordering, called strict ordering, will still force these pairs of systems to run according to the specified ordering, pointlessly restricting scheduling flexibility and thus performance.
+
+## User-facing explanation
+
+There are two basic forms of system ordering constraints:
+
+1. **Strict ordering constraints:** `.strictly_before` and `.strictly_after`
+   1. A system cannot be scheduled until "strictly before" systems have been completed during this iteration of the schedule.
+   2. Simple and explicit.
+   3. Can cause unnecessary blocking, particularly when systems are configured at a high-level.
+2. **If-needed ordering constraints:** `.before` and `.after`
+This conversation was marked as resolved by alice-i-cecile
+   1. A system cannot be scheduled until any "before" systems that it is incompatible with have completed during this iteration of the schedule.
+   2. In the vast majority of cases, this is the desired behavior. Unless you are using interior mutability (or accessing something outside of the ECS), systems that are compatible will always be **commutative**: their ordering doesn't matter, and these constraints behave "as-if" they were strict.
+   3. Note that if-needed ordering constraints are transitive. `a.before(b)` and `b.before(c)` implies `a.before(c)`. This behavior, while intuitive, can have some unexpected effects: it occurs even if the chain is broken by a **spurious** constraint (in which the two systems are compatible).
+
+### Spurious ordering constraints
+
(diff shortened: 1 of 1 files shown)


rfcs/47-if-needed-ordering.md
```diff
@@ -0,0 +1,173 @@
...
+## User-facing explanation
+
+There are two basic forms of system ordering constraints:
+
+1. **Strict ordering constraints:** `.strictly_before` and `.strictly_after`
+   1. A system cannot be scheduled until "strictly before" systems have been completed during this iteration of the schedule.
+   2. Simple and explicit.
+   3. Can cause unnecessary blocking, particularly when systems are configured at a high-level.
+2. **If-needed ordering constraints:** `.before` and `.after`
+This conversation was marked as resolved by alice-i-cecile
```
MiniaczQ, line 25:
```suggestion
```

MiniaczQ:
Can there be a situation where 2 systems need a relative ordering if they share no common `Param`s?
Perhaps strict ordering can be omitted between systems of such nature.

alice-i-cecile:
Yes; interior mutability and transitive dependencies.

MiniaczQ:
I was going to propose something similar after discord discussion of similar topic, so I'm glad I stumbled here.
Definitely a great idea to make this the default way of adding systems (I only thought of warning about too restrictive ordering).
The only thing that I'd change is the naming, which didn't immediately click for me.
I propose renaming:
- `strict` to `forced`, since it'd be forcing an ordering between 'unrelated' systems.
- `if-needed` to `lax` or `relaxed`, since it's not a binary problem whether the constraint exists, transitivity can keep the constraint, but change the systems involved.

alice-i-cecile:
I like "relaxed" quite a bit: it's a good contrast to "strict", and much more approachable.

I don't think "forced" explains the concept any better: I personally think "strict" captures the essence a bit better: the idea is "yes, for real, I know better than you".

JoJoJet:
My bikeshed opinion: how about "fixed" and "relaxed"? If someone explicitly declares an ordering between two systems, I think that by default we should adhere to that, so we shouldn't be using negative language like "strict" or "forced". In cases when a label is only there to clear up potential ambiguities, you could explicitly mark it as "relaxed" to allow for faster scheduling.

alice-i-cecile:
Strict has a pretty standard (non-negative) meaning in math and computer science: it typically contrasts to relaxed, partial, or loose and has an implication of being the more maximal version of this. I'm not sure that it's immediately clear what a "fixed ordering constraint" does, while it feels like a "strict ordering constraint" is clearer.

I would be happy with "strong / weak ordering constraints" instead though. Is that clearer to you?

JoJoJet:
Seems reasonable, "strict/relaxed" works. Definitely prefer it over "strong/weak".

rfcs/47-relaxed-ordering-constraints.md
```diff
@@ -0,0 +1,172 @@
...
+Suppose our user specifies the following relaxed ordering constraints:
+
+1. `determine_player_movement` is before `collision_detection`
+   1. Spurious: these systems are compatible.
+2. `collision_detection` is before `collision_handling`
+   1. Real: these systems conflict on `Events<Collisions>`
+3. `collision_handling` is before `apply_velocity`
+   1. Real: these systems conflict on `Velocity`
+4. `apply_player_movement` is after `collision_handling`
+   1. Spurious: these systems are compatible.
```
greaka, line 153: Why are these systems compatible? Since you cannot reason about what `apply_player_movement` exactly does to `Velocity`, it should be treated as "reads and writes `Velocity`".
Imagine the following:
```rs
fn apply_player_movement(_: Res<&PlayerIntent>, velocity: Query<(&mut Velocity, With<Player>)>) {
  if velocity > 5 {
    // doesn't really matter what happens here,
    // it branches based on potentially stale data if this is considered spurious
  }
}
```

PR #48, opened 2022-01-24
Different plugin groups to support different use cases (diverse-defaults for short)

colepoirier:
There are many different categories of bevy applications. Each use case has a sufficiently different commonsense default configuration to warrant different use case specific default plugins. This RFC is a proposal to consider making such a change.

### DIFF
--- a/rfcs/diverse-defaults.md
+++ b/rfcs/diverse-defaults.md
@@ -2,7 +2,19 @@
 
 ## Summary
 
-There are many different catagories of bevy applications. Each use case has a sufficiently different commonsense default configuration to warrant different use case specific default plugins. The catagories of bevy apps are at least the following:
+There are many different catagories of bevy applications. Each use case has a sufficiently different commonsense default configuration to warrant different use case specific default plugins.
+
+## Motivation
+
+Bevy would be doing this to support easier and faster 'getting started' user story for non-minimal bevy applications. It will be particularly helpful to beginners but would also be a nice ergonomics/UX win for experienced bevy devs to not have to configure every minute detail manually when setting up a new project that is in one of the above outlined application domains.
+
+## User-facing explanation
+
+Where there used to just be `DefaultPlugins` and `MiniminalPlugins` there will now be many different `DefaultPlugins` a user can choose from that exist for many common catagories of bevy applications.
+
+## Implementation strategy
+
+The catagories of bevy apps are at least the following:
 
 - minimal (current bevy DefaultPlugins)
 - game
@@ -38,34 +50,14 @@ For example work-minimizing scheduling for low-power consumption web and mobile 
 
 Building for touch vs mouse vs gamepad or other midi/advanced input controllers different use cases with very different requirements.
 
-## Motivation
-
-Bevy would be doing this to support easier and faster 'getting started' user story for non-minimal bevy applications. It will be particularly helpful to beginners but would also be a nice ergonomics/UX win for experienced bevy devs to not have to configure every minute detail manually when setting up a new project that is in one of the above outlined application domains.
-
-## User-facing explanation
-
-TODO:
-
-## Implementation strategy
-
-TODO:
-
 ## Drawbacks
 
-TODO:
+The primary drawback I can see is the need for bevy to maintain the multiple `DefaultPlugins`, however, I forsee the mainenance burden of this to be exceedingly minimal.
 
 ## Rationale and alternatives
 
-TODO:
-
-## \[Optional\] Prior art
-
-TODO:
+The alternative to this proposal is for bevy to continue providing only `DefaultPlugins` and `MinimalPlugins` and for users to customize the plugins themselves through writitng their own or using 3rd-party plugins.
 
 ## Unresolved questions
-
-TODO:
-
-## \[Optional\] Future possibilities
 
 TODO:
--- a/rfcs/diverse-defaults.md
+++ b/rfcs/diverse-defaults.md
@@ -1,0 +1,136 @@
+# Feature Name: Different plugin groups to support different use cases (diverse-defaults for short)
+
+## Summary
+
+There are more than half a dozen different catagories of bevy 'Apps'. Each 'use case' has a sufficiently different 'common sense' default configuration to warrant different use case specific default plugins. The catagories I have been able to come up with @alice-i-cecile's help are:
+
+- minimal (current bevy DefaultPlugins)
+- game
+ - 2D
+ - 3D
+- TUI
+- application
+ - 2D
+ - 3D 
+- simulation
+
+Each of which may have some combination of the following divergent platform configuration needs:
+
+- desktop
+ - native
+ - web
+- tablet
+ - ios native
+ - android native
+ - web
+- phone
+ - ios native
+ - android native
+ - web
+
+
+Each of these use cases can be target in a custom manner by the user. This is fine for advanced users, but is less than an ideal user story for bevy beginners who just want to get 'something' working as fast as possible lest they decide to try a different easier to use engine/framework. Were bevy to provide sane defaults for each use case new users would be able to get soemthing working in non-standard (i.e. minimial DefaultPlugins) very easily. This can also be highlighted by different book chapters per use case.
+
+THis also involves not-yet-implemented but planned bevy features that will have different implementations per use case.
+
+For example work-minimizing scheduling for low-power consumption web and mobile applications, or desktop productivity apps are a very different configuration from high performance gaming and high performance scientific simulation.
+
+2D vs 3D will have different built-in bevy selection methods/mechanisms and different cameras... Anything else?
+
+Building for touch vs mouse vs gamepad or other midi/advanced input controller are yet more very different use cases with very different requirements.
+
+I have run out of ideas? Any suggestions?
+
+## Motivation
+
+Bevy would be doing this to support easier an faster getting started in the major catagories of different use cases for bevy. It is particularly helpful to beginniers but would be a nice ergonomics/UX win for experienced bevy devs to not have to configure every minute detail manually when working in multiple 'use case' domains on different projects. Anything else you can think of?
+
+## User-facing explanation
+
+Explain the proposal as if it was already included in the engine and you were teaching it to another Bevy user. That generally means:
+
+- Introducing new named concepts.
+    - DefaultPlugins can become an enum with each usecase as a different enumerant?
+- Explaining the feature, ideally through simple examples of solutions to concrete problems.
+    - Small book chapters would probably be the way to go?
+- Explaining how Bevy users should *think* about the feature, and how it should impact the way they use Bevy. It should explain the impact as concretely as possible.
+    - What I would put here is identical to what I have written in the motivation section, is that ok?
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+    - DefaultPlugins for the current minimal config would still work the same, IIUC this would be a purely additive feature.
+- If applicable, explain how this feature compares to similar existing features, and in what situations the user would use each one.
+    - Currently users can create these configs manually, the change would be bevy providing more standard getting started configs/DefaultPlugins out of the box.
+
+## Implementation strategy
+
+I need others help/input with this.
+
+This is the technical portion of the RFC.
+Try to capture the broad implementation strategy,
+and then focus in on the tricky details so that:
+
+- Its interaction with other features is clear.
+- It is reasonably clear how the feature would be implemented.
+- Corner cases are dissected by example.
+
+When necessary, this section should return to the examples given in the previous section and explain the implementation details that make them work.
+
+When writing this section be mindful of the following [repo guidelines](https://github.com/bevyengine/rfcs):
+
+- **RFCs should be scoped:** Try to avoid creating RFCs for huge design spaces that span many features. Try to pick a specific feature slice and describe it in as much detail as possible. Feel free to create multiple RFCs if you need multiple features.
+- **RFCs should avoid ambiguity:** Two developers implementing the same RFC should come up with nearly identical implementations.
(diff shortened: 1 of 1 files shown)


rfcs/48-diverse-defaults.md
```diff
@@ -0,0 +1,60 @@
+# Feature Name: Different plugin groups to support different use cases (diverse-defaults for short)
+
+## Summary
+
+There are many different categories of bevy applications. Each use case has a sufficiently different commonsense default configuration to warrant different use case specific default plugins.
```
alice-i-cecile, line 5:
```suggestion
There are many different categories of Bevy applications. Each use case has distinct needs, and so we should provide a collection of `PluginGroups` to give our users good defaults.
```

rfcs/48-diverse-defaults.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Motivation
+
+Bevy would be doing this to support easier and faster 'getting started' user story for non-minimal bevy applications. It will be particularly helpful to beginners but would also be a nice ergonomics/UX win for experienced bevy devs to not have to configure every minute detail manually when setting up a new project that is in one of the above outlined application domains.
```
alice-i-cecile, line 9: This needs a bit of clarity pass.

rfcs/48-diverse-defaults.md
```diff
@@ -0,0 +1,60 @@
...
+
+## User-facing explanation
+
+Where there used to just be `DefaultPlugins` and `MiniminalPlugins` there will now be many different `DefaultPlugins` a user can choose from that exist for many common categories of bevy applications.
```
alice-i-cecile, line 13: IMO this whole line can just be cut. This section should be "new docs", not "migration guide" :)

rfcs/48-diverse-defaults.md
```diff
@@ -0,0 +1,60 @@
...
+
+Bevy now provides you a number of convenient `PluginGroups` to help you get your application started with sensible defaults for your use case. The current list is:
+
+- `MinimalPlugins`: The very basic internal structure required to ensure things like `Time` function properly.
+- `GamePlugins`: Suitable for making both 2D and 3D real-time games with input handling, audio and rendering (note: previously `DefaultPlugins`).
+- `2dGamePlugins`: Like `GamePlugins`, but limited to the systems and resources needed in a 2D game.
+- `TuiPlugins`: Suitable for making TUI applications.
+- `ApplicationPlugins`: Suitable for making CAD applicaions or something like excel.
```
alice-i-cecile, line 21:
```suggestion
- `ApplicationPlugins`: Suitable for making relatively static non-game applications.
```

rfcs/48-diverse-defaults.md
```diff
@@ -0,0 +1,60 @@
...
+- `SimulationPlugins`: Plugins that make writing scientific simulations easier like enhanced determinism and ordering guarantees.
+
+## Implementation strategy
+
+Each of the above outlined `Plugins` may have some combination of the following divergent platform configuration needs:
```
alice-i-cecile, line 26: It's not clear how exactly this interacts with the "use case" plugins groups above 🤔 I agree that we want both, but they seem mostly orthogonal.

rfcs/48-diverse-defaults.md
```diff
@@ -0,0 +1,60 @@
...
+ - android native
+ - web
+- phone
+ - ios native
+ - android native
+ - web
+
+This feature also touches on not-yet-implemented but planned bevy features that will have different implementations per use case.
+
+For example work-minimizing scheduling for low-power consumption web and mobile applications, or desktop productivity apps are a very different configuration from high performance gaming and high performance scientific simulation.
```
alice-i-cecile, line 42:
```suggestion
For example, using making low-power consumption web and mobile applications,  desktop productivity apps and high performance games will all have very different scheduling needs.
```

rfcs/48-diverse-defaults.md
```diff
@@ -0,0 +1,60 @@
...
+Bevy now provides you a number of convenient `PluginGroups` to help you get your application started with sensible defaults for your use case. The current list is:
+
+- `MinimalPlugins`: The very basic internal structure required to ensure things like `Time` function properly.
+- `GamePlugins`: Suitable for making both 2D and 3D real-time games with input handling, audio and rendering (note: previously `DefaultPlugins`).
+- `2dGamePlugins`: Like `GamePlugins`, but limited to the systems and resources needed in a 2D game.
+- `TuiPlugins`: Suitable for making TUI applications.
+- `ApplicationPlugins`: Suitable for making CAD applicaions or something like excel.
+- `SimulationPlugins`: Plugins that make writing scientific simulations easier like enhanced determinism and ordering guarantees.
+
+## Implementation strategy
```
alice-i-cecile, line 24: I think it would be helpful hear to identify "axes of variation"; ways that each of the use cases commonly vary, so we can think about how to structure them.

rfcs/48-diverse-defaults.md
```diff
@@ -0,0 +1,60 @@
...
+## User-facing explanation
+
+Where there used to just be `DefaultPlugins` and `MiniminalPlugins` there will now be many different `DefaultPlugins` a user can choose from that exist for many common categories of bevy applications.
+
+Bevy now provides you a number of convenient `PluginGroups` to help you get your application started with sensible defaults for your use case. The current list is:
+
+- `MinimalPlugins`: The very basic internal structure required to ensure things like `Time` function properly.
+- `GamePlugins`: Suitable for making both 2D and 3D real-time games with input handling, audio and rendering (note: previously `DefaultPlugins`).
+- `2dGamePlugins`: Like `GamePlugins`, but limited to the systems and resources needed in a 2D game.
+- `TuiPlugins`: Suitable for making TUI applications.
```
alice-i-cecile, line 20: This needs a note in implementation details on what we should include to get this working.

rfcs/48-diverse-defaults.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Motivation
+
+Bevy would be doing this to support easier and faster 'getting started' user story for non-minimal bevy applications. It will be particularly helpful to beginners but would also be a nice ergonomics/UX win for experienced bevy devs to not have to configure every minute detail manually when setting up a new project that is in one of the above outlined application domains.
+
+## User-facing explanation
+
+Where there used to just be `DefaultPlugins` and `MiniminalPlugins` there will now be many different `DefaultPlugins` a user can choose from that exist for many common categories of bevy applications.
+
+Bevy now provides you a number of convenient `PluginGroups` to help you get your application started with sensible defaults for your use case. The current list is:
```
alice-i-cecile, line 15: I would really like "turn-based game" as an plugin group starter.

rfcs/48-diverse-defaults.md
```diff
@@ -0,0 +1,60 @@
...
+
+For example work-minimizing scheduling for low-power consumption web and mobile applications, or desktop productivity apps are a very different configuration from high performance gaming and high performance scientific simulation.
+
+2D vs 3D will have different built-in bevy selection methods/mechanisms and different cameras.
+
+Building for touch vs mouse vs gamepad or other midi/advanced input controllers different use cases with very different requirements.
+
+## Drawbacks
+
+The primary drawback I can see is the need for bevy to maintain the multiple `DefaultPlugins`, however, I foresee the maintenance burden of this to be exceedingly minimal.
```
alice-i-cecile, line 50:
```suggestion
The primary drawback I can see is the need for Bevy to maintain the multiple `PluginGroups`, however, I foresee the maintenance burden of this to be exceedingly minimal.
```

alice-i-cecile (changes requested):
First round of feedback and clarity suggestions :)

rfcs/48-diverse-defaults.md
speakk: How would that differ from a regular 2d starter? I think there's a risk of overscoping and complicating this.

mockersf:
Instead of having different `PluginGroups`, I think this would be better suited for compilation features:
- they reduce compile time
- they guarantee that the code won't be present
- they are the only way to have different `const` values
- they work well with compilation target, which is what you call "divergent platform"
- they are already present and mostly usable that way in Bevy

For example, for a 2d game that won't use UI but will still display text in the 2d world, I use only those features: `"bevy_core_pipeline", "bevy_sprite", "bevy_winit", "bevy_render", "bevy_text", "png"`.

Features are probably not documented enough and not really friendly, Bevy could add meta features to group them into logical groups like you describe here.

rfcs/48-diverse-defaults.md
colepoirier: I very much agree, though I don't know how to flesh this out.

rfcs/48-diverse-defaults.md
colepoirier: What exactly would we need to include to get this working? I don't understand the details of the bevy `DefaultPlugins` code I guess.

rfcs/48-diverse-defaults.md
colepoirier: I second @speakk's sentiment and question.

colepoirier:
> Instead of having different `PluginGroups`, I think this would be better suited for compilation features:
> 
>     * they reduce compile time
> 
>     * they guarantee that the code won't be present
> 
>     * they are the only way to have different `const` values
> 
>     * they work well with compilation target, which is what you call "divergent platform"
> 
>     * they are already present and mostly usable that way in Bevy
> 
> 
> For example, for a 2d game that won't use UI but will still display text in the 2d world, I use only those features: `"bevy_core_pipeline", "bevy_sprite", "bevy_winit", "bevy_render", "bevy_text", "png"`.
> 
> Features are probably not documented enough and not really friendly, Bevy could add meta features to group them into logical groups like you describe here.

That's a really nice idea @mockersf and I think this solves the problem I am seeking to address in an even nicer way with the additional benefits you outlined!! :) :100:

rfcs/48-diverse-defaults.md
colepoirier: Done!

rfcs/48-diverse-defaults.md
colepoirier: I think @mockersf proposal enables use to meld the two in a sensible way!

rfcs/48-diverse-defaults.md
colepoirier: Done!

rfcs/48-diverse-defaults.md
colepoirier: Also would probably prefer to just migrate my initial proposal to the one @mockersf suggested which I think addresses the problem area even better! :)

rfcs/48-diverse-defaults.md
alice-i-cecile: Yeah, I'm totally wrong here.

rfcs/48-diverse-defaults.md
alice-i-cecile: The bigger problem is "I don't know how to make a terminal game in Bevy 0.6" at all. There were a few ecosystem crates, but they're no longer up to date.

rfcs/48-diverse-defaults.md
```diff
@@ -0,0 +1,58 @@
...
+
+2D vs 3D will have different built-in bevy selection methods/mechanisms and different cameras.
+
+Building for touch vs mouse vs gamepad or other midi/advanced input controllers different use cases with very different requirements.
+
+## Drawbacks
+
+The primary drawback I can see is the need for Bevy to maintain the multiple `PluginGroups`, however, I foresee the maintenance burden of this to be exceedingly minimal.
+
+## Rationale and alternatives
```
alice-i-cecile, line 50: I think you should discuss the pros and cons of plugin groups vs. feature groups in this section.

colepoirier: How would you define `FeatureGroups`?

rfcs/48-diverse-defaults.md
```diff
@@ -0,0 +1,61 @@
...
+
+Bevy now provides you a number of convenient `PluginGroups` to help you get your application started with sensible defaults for your use case. The current list is:
+
+- `MinimalPlugins`: The very basic internal structure required to ensure things like `Time` function properly.
+- `GamePlugins`: Suitable for making both 2D and 3D real-time games with input handling, audio and rendering (note: previously `DefaultPlugins`).
+- `2dGamePlugins`: Like `GamePlugins`, but limited to the systems and resources needed in a 2D game.
+- `TuiPlugins`: Suitable for making TUI applications.
+- `ApplicationPlugins`: Suitable for making relatively static non-game applications.
+- `SimulationPlugins`: Plugins that make writing scientific simulations easier like enhanced determinism and ordering guarantees.
+- `HeadlessServerPlugins`: Like MinimalPlugins but with no rendering or windowing.
```
alice-i-cecile, line 21:
```suggestion
- `HeadlessServerPlugins`: Like `GamePlugins`, but without rendering, input handling, audio or windowing.
```

rfcs/48-diverse-defaults.md
```diff
@@ -0,0 +1,61 @@
...
+- `HeadlessIntegrationTestPlugins`: Suitable for running Bevy's integration tests as rapidly as possible.
```
alice-i-cecile, line 22:
```suggestion
- `HeadlessIntegrationTestPlugins`: Suitable for running Bevy integration tests as rapidly as possible.
```

rfcs/48-diverse-defaults.md
```diff
@@ -0,0 +1,61 @@
...
+- `RenderTestPlugins`: The minimal plugins necessary for testing the Bevy renderer.
```
alice-i-cecile, line 23:
```suggestion
- `RenderTestPlugins`: The minimal plugins necessary for testing that Bevy is rendering objects as expected.
```

rfcs/48-diverse-defaults.md
alice-i-cecile: Rust features can enable other features, allowing users to toggle them on and off as needed. We can also 

See: https://doc.rust-lang.org/cargo/reference/features.html for a basic introduction.
https://docs.rs/cfg_feature_groups/latest/cfg_feature_groups/ or the like may be helpful to improve ergonomics.

PR #49, opened 2022-02-11
Animation Primitives

james7132:
[RENDERED](https://github.com/james7132/rfcs/blob/animation-primitives/rfcs/49-animation-primitives.md)

This RFC details the absolute bare minimum viable API for setting up a flexible and extendable property based animation system in Bevy... or at least part of it. This gets probably about 2/3 of the way there.


alice-i-cecile:
Ping @lassade; I'm very curious on your opinions on this if you have the time and interest.

rfcs/49-animation-primitives.md
```diff
@@ -0,0 +1,154 @@
...
+## Summary
+
+Animation is a particularly complex, with many stateful and intersecting
+elements. This RFC aims to detail a set of lowest-level APIs for authoring and
+playing animations within Bevy.
+
+## Motivation
+
+Animation is at the heart of modern game development. A game engine without an
+animation system is generally considered not yet production ready.
```
alice-i-cecile, line 12:
```suggestion
animation system is generally not considered production-ready.
```

rfcs/49-animation-primitives.md
```diff
@@ -0,0 +1,154 @@
...
+
+This RFC aims to detail the absolute lowest level APIs to unblock ecosystem
```
alice-i-cecile, line 14:
```suggestion
This RFC aims to detail the absolute lowest-level APIs to unblock ecosystem
```

rfcs/49-animation-primitives.md
```diff
@@ -0,0 +1,154 @@
...
+    animated behaviors.
+ 4. **Authoring**: this is how we create the animation assets used by the engine.
+
+This RFC specifically aims to resolve only problems within the domains of storage
+and sampling. Application can be distinctly decoupled from these earlier two
+stages, treating the sampled values as a black box output, and composition and
+authoring can be built separately upon the primitives provided by this RFC and
+thus are explicit non-goals here.
+
+## User-facing explanation
```
alice-i-cecile, line 38: IMO this deserves an explanation of how "animation is anything that changes over time in a smooth and predictable way".

A lot of new users, especially those who haven't been exposed to animation before, have a conceptualization of animation that only includes modifying the `Transform` of entities.

rfcs/49-animation-primitives.md
```diff
@@ -0,0 +1,154 @@
...
+
+The core of this system is a trait called `Sample<T>` which allows sampling
+values of `T` from an underlying type at a provided time. `T` here can be
+anything considered animatable. A few examples of high priority types to be
+supported here are:
```
alice-i-cecile, line 43: `Transform` very much needs to be considered.

rfcs/49-animation-primitives.md
```diff
@@ -0,0 +1,154 @@
...
+
+ - `f32`/`f64`
+ - `Vec2`/`Vec3`/`Vec3A`/`Vec4` (and their `f64` variants)
+ - `Color`
+ - `Option<T>` where `T` can also be sampled
+ - `bool` for toggling on and off features.
```
alice-i-cecile, line 49: This will cause interesting problems for interpolation.

alice-i-cecile (changes requested):
Generally I'm very impressed. These are good abstractions that are powerful, flexible and build on each other.

rfcs/49-animation-primitives.md
james7132: This isn't entirely true, since there are discrete valued animations as well, so the wording here might need a bit more wordsmithing.

rfcs/49-animation-primitives.md
james7132: I originally considered `Transform` as a composition of `Vec3`/`Quat` fields, but I think we can add a special case for it too.

rfcs/49-animation-primitives.md
james7132: The implementation right now has step-valued interpolation for types with discrete values.

rfcs/49-animation-primitives.md
```diff
@@ -0,0 +1,154 @@
...
+
+The design chosen here was explicitly to allow for maximum flexibility for both
+engine and game developers alike. The main alternative is to completely remove
+`Sample<T>`, the animation graph, and it's nodes, and let developers directly
+hook up animation curves from clips to entities. However, this lacks flexibility
+and does not allow for users of the API to inject their own alterations to the
+animation stream.
+
+The main potential issue with the current implementation is the very heavy use of
+`Arc<dyn Sample<T>>` which has CPU cache, lifetime, and performance implications
```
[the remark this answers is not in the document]
james7132: Due to how complex the graph evaluation already is, passing in a `&Assets<T>` severely complicate every level of the graph API. There's also the issue of directly tying the graph base inputs to `Assets` with a concrete type, as a user may have a new sample-able data source that isn't easily added into the graph.

We could make a concrete enum that has a trait object variant, but that still involves the complexity of passing `&Assets<T>` through the entire graph.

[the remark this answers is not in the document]
alice-i-cecile: Sounds good. This should be explained in Rationale and Alternatives :)

lassade:
In the storage domain you should include how streamed data will be handled and at mention available animation formats namely ACL (https://github.com/nfrechette/acl) in the prior art.

james7132:
Almost completely revised the RFC end to end:
 - Removed the mention of animation graph and mixing entirely, going to have a separate RFC for this
 - Dropped the idea of `Sample<T>`, and instead opting for defining the base characteristics required for an animatable property instead, which will be the base for the composition API
 - Dove deeper into the implementation details and memory layout.
 - Addressed review comments from @alice-i-cecile and @lassade 

Still quite a few TODOs, namely discussing a strategy for implementing `AnimationClip` and it's serialization.

--- 

To respond to @lassade's comments

> 1. Not sure how curves and sampling could be decoupled into to separated traits, to me they are the same;

Good point, retracted that idea after discussing this with you on Discord and reading your `Animator` implementation. I've changed it now to instead have a trait on the sampled value named `Animatable`, which takes your `Lerp` and logically extends it one step further for allowing anyone to make their own blendable value. This will be used to blend sampled values from curves the next Animation Graph attempt I'm testing instead of relying on generic nodes as previously proposed. This should remove generics from everywhere except for the `Curve<T>` trait itself.

> 2. Sampling also involves how data should be interpolated: step, linear and catmul-row; I propose to have 3 different types of curves like in my original implementation;

I'm tentatively addressing this by trying to use a newtype and implement `Animatable` on it instead. Having `T` be a generic parameter to `Curve` instead of an associated type actually allows us to have multiple `Curve<T>` implementations for a given type, and we can unwrap the newtype in a blanket impl if we need to. Having multiple curve types was originally always in the plan, just didn't have time to write down the full details, but if the underlying structure of the curve does not change fundamentally change, but the interpolation behavior changes, we can just newtype the value type instead. There's an argument to be made here about SoA vs AoS in wrapping values like this, which is why the `CurveVariable` in your implementation is the way it is.

> 3. For Quaternions there is a difference between lerp vs slerp, other than that generally the type it self defines how interpolate between values. You will almost never will want to use slerp by the way;

This can be addressed by the newtype pattern that I mentioned above. Though I'm curious why we wouldn't want to use slerp other than performance.

> 4. Color interpolation will be a nightmare If the Colors is defined as a enum that supports sRGB, LinSRGB, HSL and others, because each color space gives a different interpolation result; If you let the user choose the interpolation color space it might have a heavy performance cost, probably many times greater than using slerp;

Right, and that's sort of an implementation detail that we need to address, and is sort of separate from the design detailed here. I'm well aware this is an issue, but it's something we need to resolve in general, not just for animation.

> 5. Sampling should include some form of cursor (cache locality, think a curve with many keyframes) and should and use binary search (faster keyframe search). But you should only use the latter to make a simpler initial implementation;

I added a section on this, though I'm not 100% sold on it. After reading your implementation, I get the distinct impression that this very heavily complicates the implementation as the cursor must be mutably saved somewhere, and you might have the same curve/clip being sampled in different locations by multiple entities (or even within the same entity). That's not to say I think we should definitely not add it in, but it does seem like an optimization we can add in after we deliver a MVP on this.

> In the storage domain you should include how streamed data will be handled and at mention available animation formats namely ACL (https://github.com/nfrechette/acl) in the prior art.

Great point! Added that a mention of ACL in. I need to rewrite that section to mention the other efforts in the Rust community and other public literature about the approaches to this.

Weibye (commented):
To me this looks like flexible and powerful primitives to build on top of. I particularly like how this would build animation in bevy from the ground up to be able to animate anything and not just skeletal animation.

- Obvious outstanding TODOs
- Don't feel I have enough experience to give feedback on the suggested technical implementation details
- It is clearly and concisely written, so even with limited technical knowledge it is easy to read and understand

Overall this is great work, thank you!

ValorZard:
Looking at the curve system as explained in the RFC, it reminds me of how Godot does animations, specifically how Godot lets you animate literally everything including UI all using the same UI graph. I'm wondering, will the Curve<T> syntax work for everything, or only for certain things like skeletal animation?

rfcs/49-animation-primitives.md
```diff
@@ -0,0 +1,398 @@
...
+these paths are commonly used as HashMap and BTreeMap keys. The raw component
+name is only used when converting back into a raw string.
+
+As a result of needing runtime type information, serialization and deserialzation
+cannot be implemented with serde derive macros, but will require a `TypeRegistry`
+to handle the TypeId lookup. For simpler file representations, upon serialization,
+PropertyPaths are written out as raw strings, and parsed on deserialization.
+
+### `AnimationClip`
+A `AnimationClip` is a serializable asset type that encompasses a mapping of
```
Aceeri, line 343:
```suggestion
An `AnimationClip` is a serializable asset type that encompasses a mapping of
```

rfcs/49-animation-primitives.md
```diff
@@ -0,0 +1,398 @@
...
+may be difficult to know which one to use for when and will require signifigant
+documentation effort. The core also centering around `Curve<T>` as a trait also
+encourages use of trait objects over concrete types, which may have runtime costs
+associated with its use.
+
+## Rationale and alternatives
+Bevy absolutely needs some form of a first-party animation system, no modern game
+engine can be called production ready without one. Having this exist solely as a
+third-party ecosystem crate is unacceptable as it would promote a
+facturing of the ecosystem with multiple incompatible baseline animation system
```
Aceeri, line 381:
```suggestion
fracturing of the ecosystem with multiple incompatible baseline animation system
```

rfcs/49-animation-primitives.md
```diff
@@ -0,0 +1,398 @@
...
+animation compression algorithms for general use in game engines. It is
+unfortunately written in C++, which may make it difficult to directly integrate
+with Bevy, but it's open sourced under MIT, which would make a reimplementation
+of its algorithms in Rust compatible with Bevy.
+
+[acl]: https://technology.riotgames.com/news/compressing-skeletal-animation-data
+
+## Unresolved questions
+ - Can we safely interleave multiple curves together so that we do not incur
+   mulitple cache misses when sampling from animation clips?
```
Aceeri, line 395:
```suggestion
   multiple cache misses when sampling from animation clips?
```

Aceeri (changes requested):
I like the concept, I'm going to think about it for a bit but here are a couple of nitpick-y grammar/etc.

mweatherley:
I might be a couple years late to the party here, but I have some thoughts on `Curve<T>` as associated with other curve-related matters that have recently been on my mind.

In investigating matters related to, e.g. mesh generation by curve extrusion, I have also come up against curve sampling and similar questions related to what appear here as `CurveFixed` etc. For instance, in order to perform acceptable mesh extrusion for curves in general, one really needs at least the following in addition to the points on the curve:
1. Tangents — ideally with regularity assumptions, so that, for example, `T = Dir3` is supported for many curves we construct in addition to `T = Vec3` (or equivalent)
2. Arclength (`T = f32` or similar)
3. Rotation-minimizing frames, given by `T = Mat3`, `T = Quat` (interpreted as orientation quaternions) or similar, with something like spherical interpolation papering over the gaps

Notably, for curves constructed by splines, (1) is available algebraically (hence for arbitrary parameter values with no interpolation necessary), while (2) and (3) must generally be constructed by sampling (although for some simple classes of curves they could also be algebraic). Here are what I see as the main differences:
- For these kinds of application in particular, sampling arbitrary time parameters is often less important, since the resolution can often be bound to the task at hand directly — if I want to generate a mesh with some given curve-space resolution, I can just sample the values I need at that frequency (albeit perhaps with some substeps to avoid numerical error) in the first place
- Insofar as interpolation is needed, it would generally consist only of `Animatable::interpolate`
- Concrete `Curve` implementations will often blend discrete data (possibly interpolated) with data that can be exactly computed from underlying algebraic structures

I think the good news is that these differences are mostly easy to accommodate. For the first, it is not really much of a rift, as long as the `Curve<T>` implementor has its resolution specified at the time of construction anyway. 

For the third, I see this mostly as creating difficulties when serializing and deserializing. For example, something like type erasing the underlying algebraic structure, if necessary, could require passage from exact information to discrete samples; luckily, if we come to that bridge, the other kinds of data that we need are still "compatible" in the sense that they could be recomputed from discrete data anyway (that is to say that RMF, for instance, really only needs a `Curve<(Vec3, Dir3)>` in its construction).

The second is where I wanted to raise a minor point. I am against the exact definition of `Curve<T>` in this proposal only because it requires `T: Animatable` rather than something more general (especially with a name as all-encompassing as `Curve`). Concretely, what I would prefer is to do the following:
- Introduce a weaker trait, `Interpolable` (or similar) in `bevy_math`, with `Animatable: Interpolable` in `bevy_animation`
- Introduce the trait `Curve<T: Interpolable>` in `bevy_math` and implement it for sensible values of `T` with our spline output curves
- Have animation depend on `Curve<T> where T: Animatable` explicitly

I think that this would provide a solid common baseline (at least at the level of API) for the distinct areas.

(By the way, thank you for writing this, and sorry if I am verbose.)

PR #50, opened 2022-02-17
Cloneable components

alice-i-cecile:
[**RENDERED**](https://github.com/alice-i-cecile/rfcs/blob/clonable-components/rfcs/50-clonable-components.md)

TL;DR: Cloning components and entire entities is fundamentally useful.  Requiring `Clone` is too strict (thanks for the counterexamples!), so we must use an opt-in strategy. Trait reflection is the best tool to do so within the context of the Bevy ecosystem.

### DIFF
--- a/rfcs/50-clonable-components.md
+++ b/rfcs/50-clonable-components.md
@@ -1,0 +1,210 @@
+# Feature Name: `clonable-components`
+
+## Summary
+
+The `Component` trait requires the `Clone` trait to be implemented.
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
Closing this out; all of the interesting bits of this are just a special-case of trait reflection so this no longer warrants its own RFC.

colepoirier:
@alice-i-cecile can you link the Trait Reflection rfc here for future readers if/when that rfc exists?

PR #52, opened 2022-03-05
Unstable Features for faster iterations

mockersf:
[Rendered](https://github.com/mockersf/rfcs/blob/unstable-features/rfcs/unstable_features.md)

### DIFF
--- a/rfcs/unstable_features.md
+++ b/rfcs/unstable_features.md
@@ -1,0 +1,65 @@
+# Feature Name: `unstable_features`
+
+## Summary
+
+Accepting PRs for new features that may not have reach the high bar we set for Bevy, and disable their changes with a feature gate.
+
+## Motivation
+
+Some features would benefit from iterative work in Bevy, but that may be hard to do when a PRs need a high level of quality to be merged in.
+
+In this RFC, "feature" means the new feature or change under development, and "feature gate" the conditional compilation flag. "unstable" means a feature that is subject to change, not something that will break Bevy.
+
+## User-facing explanation
+
+So you want to contribute a new feature to Bevy! You have a few paths available to you, depending on the scope of the feature:
+- A small feature with a clear implementation? Go ahead and open that PR!
+- A large feature with an open design space that could use some discution about how to implement it? A RFC would be good.
+- Something in between that would benefit from iterative work? Here come the unstable feature workflow for you!
+
+You will need to open a tracking issue to explain the new feature you want to add. If it looks good, you'll be ready to open that PR with a feature gate for your changes!
+
+## Implementation strategy
+
+The new workflow for this kind of feature would be:
+- Submit an issue explaining the feature you want to add under a feature gate. This will be the tracking issue for stabilization once approved. This issue gets a new label `S-Stabilization`.
+- If approved, submit a PR with the new feature and the feature gate
+- This PR can be merged by someone with the merge rights on the related part of Bevy as soon as it's `S-Ready-For-Final-Review`, without an approval by the main Bevy maintainers.
+- Subsequent PRs on the feature must be linked to the tracking issue and follow the same process.
+- Once the feature is finished, enter the stabilization period. Open a PR that will remove the feature gate. This PR need to be approved by the main Bevy maintainers.
+
+An existing PR can be retroffited in this process on the suggestion of Bevy community members, and if the PR author agrees.
+
+The RFCs and the unstable features are closely related. An RFC can be implemented as an unstable feature to help resolve issues that need more experimentation. An unstable feature could require an RFC to reach stabilization.
+
+Each PR will still need to be approved by Bevy community members, and would still need to have the same level of quality as other PRs merged.
+
+The PRs for the initial feature, various changes during its finalization and for the stabilization can be opened by different persons.
+
+The feature gate should not be exposed on the main `bevy` crate, but only on the subcrates where it is relevant. It can be enabled as a user by adding dependencies directly on the subcrates and enabling the feature.
+
+## Drawbacks
+
+This will increase complexity in Bevy code.
+
+It will also increase complexity of a contribution that would like to use this process, but it is opt in and not mandatory. Think of it as a fast track for unstable features.
+
+## Rationale and alternatives
+
+- Improve the process to get new features in Bevy that would benefir from iterative work.
+- Alternatives:
+    - Continue as it is now, with semi large PRs that bitrot and RFCs that don't often move forward
+    - Work harder to implement a new feature as a third party crate (this is not always possible)
+    - Fork Bevy to merge unnaproved PRs and check how they work / interact
+
+
+## Prior art
+
+[The Rust process](https://rustc-dev-guide.rust-lang.org/implementing_new_features.html)
+
+## Unresolved questions
+
+- Do we want to limit the number of feature gates in Bevy to reduce complexity?
+    - this could help control the increased complexity in Bevy
+- Do we want to limit the maximum age of a unstabilized feature?
+    - this could help control the increased complexity in Bevy, and requires being ready to cut a feature off if it's not getting the use / stabilization expected


rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,65 @@
...
+
+Some features would benefit from iterative work in Bevy, but that may be hard to do when a PRs need a high level of quality to be merged in.
+
+In this RFC, "feature" means the new feature or change under development, and "feature gate" the conditional compilation flag. "unstable" means a feature that is subject to change, not something that will break Bevy.
+
+## User-facing explanation
+
+So you want to contribute a new feature to Bevy! You have a few paths available to you, depending on the scope of the feature:
+- A small feature with a clear implementation? Go ahead and open that PR!
+- A large feature with an open design space that could use some discution about how to implement it? A RFC would be good.
```
james7132, line 17:
```suggestion
- A large feature with an open design space that could use some discussion about how to implement it? A RFC would be good.
```

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,65 @@
...
+- Something in between that would benefit from iterative work? Here come the unstable feature workflow for you!
```
james7132, line 18: I think large RFC'ed changes should also be subject to this workflow too. It'd also help isolate the effects of implementing individual parts of the larger design before it becomes stabilized.

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,65 @@
...
+
+An existing PR can be retroffited in this process on the suggestion of Bevy community members, and if the PR author agrees.
+
+The RFCs and the unstable features are closely related. An RFC can be implemented as an unstable feature to help resolve issues that need more experimentation. An unstable feature could require an RFC to reach stabilization.
+
+Each PR will still need to be approved by Bevy community members, and would still need to have the same level of quality as other PRs merged.
+
+The PRs for the initial feature, various changes during its finalization and for the stabilization can be opened by different persons.
+
+The feature gate should not be exposed on the main `bevy` crate, but only on the subcrates where it is relevant. It can be enabled as a user by adding dependencies directly on the subcrates and enabling the feature.
```
james7132, line 39: How would we tackle breaking new ground with new official crates then? Should these be added as unstable features to the main bevy crate?

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,65 @@
...
+## User-facing explanation
+
+So you want to contribute a new feature to Bevy! You have a few paths available to you, depending on the scope of the feature:
+- A small feature with a clear implementation? Go ahead and open that PR!
+- A large feature with an open design space that could use some discution about how to implement it? A RFC would be good.
+- Something in between that would benefit from iterative work? Here come the unstable feature workflow for you!
+
+You will need to open a tracking issue to explain the new feature you want to add. If it looks good, you'll be ready to open that PR with a feature gate for your changes!
+
+## Implementation strategy
```
james7132, line 22: I think a convention for feature flags naming should be kept so that users can quickly discern which features are unstable, and we should have some log or central location to track these features.

Also unfortunately feature flag doc generation are currently bound to nightly. Not sure how to get around this before that feature gets stabilized.

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Prior art
+
+[The Rust process](https://rustc-dev-guide.rust-lang.org/implementing_new_features.html)
+
+## Unresolved questions
+
+- Do we want to limit the number of feature gates in Bevy to reduce complexity?
+    - this could help control the increased complexity in Bevy
+- Do we want to limit the maximum age of a unstabilized feature?
```
james7132, line 64: With the Bevy train release schedule, it'd be nice to have a minimum time for this too, to allow for folks using the released versions of Bevy to have to test out and file bugs. This could be mitigated by telling users to use the `main` branch, but that might not always be possible with the growing ecosystem and dependency compatibility.

As an upper bound, if a unstable feature doesn't get stabilized within two train releases, even with external users testing it out, I'd say we either need to reconsider the design of it as a full RFC or just scrap the idea entirely.

rfcs/52-unstable_features.md
mockersf: Yes I agree, but in those case I think the workflow should be RFC first, then unstable feature once the RFC is more or less agreed on. I tried to mention that in the "implementation" part, would that work for you?

rfcs/52-unstable_features.md
mockersf: No, a new crate can be added directly as an optional under feature gate dependency to the existing crate that need it, or directly as an added dependency. It shouldn't be any different than an existing subcrate

rfcs/52-unstable_features.md
mockersf: With Rust as an example where stabilising a feature can take a few years, I think a good scope would be at least one version, at most 4 (so between 3 months and a year)

rfcs/52-unstable_features.md
mockersf: Agreed on the convention.

I think it's good that the doc won't be built for those unstable features by default. If someone wants it, the doc can be built on demand with the feature gate enabled.

rfcs/52-unstable_features.md
james7132: Not sure if that's a good model to follow since many new crates will be building on top of existing foundational crates, not as dependencies for them. Or is the intention to have it added to bevy_internal in those cases?

rfcs/52-unstable_features.md
james7132: On one hand, I'm inclined to agree that we shouldn't be putting this forth front and center, but that also hides these features from those who might both benefit from them in the short term and be a good testbed for stabilization. By showing them and explicitly labeling APIs as "unstable, expect bugs", I think that should deter any user expecting stability. The same goes for nightly-bound unstable features in Rust: it garners both testers and active feedback.

rfcs/52-unstable_features.md
james7132: I see it now, but it's not particularly focused upon. Could you expand on each potential workflow down below?

mockersf: I tried!

rfcs/52-unstable_features.md
mockersf: I think a GitHub project showing tracking the stabilisation issues should be good enough for discovery, what do you think?
I added that proposition to the RFC. Keeping the project up to date could be done by GitHub Actions, but I think that's out of scope of this RFC.

rfcs/52-unstable_features.md
mockersf: I think having as dependencies:
```
bevy = "0.6"
bevy_ecs = { version = "0.6", features = [ "unstable-4242-my-great-feature" ] }
bevy_new_crate = { version = "0.6" }
```

or targeting main
```
bevy = { git = "https://github.com/bevyengine/bevy" }
bevy_ecs = { git = "https://github.com/bevyengine/bevy", features = [ "unstable-4242-my-great-feature" ] }
bevy_new_crate = { git = "https://github.com/bevyengine/bevy" }
```

should work?

The other option would be indeed to raise all the unstable feature gates to `bevy_internal`, but I'm not sure that's worth the (very small) cost

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,82 @@
...
+
+Each PR will still need to be approved by Bevy community members, and would still need to have the same level of quality as other PRs merged.
+
+The PRs for the initial feature, various changes during its finalization and for the stabilization can be opened by different persons.
+
+The feature gate should not be exposed on the main `bevy` crate, but only on the subcrates where it is relevant. It can be enabled as a user by adding dependencies directly on the subcrates and enabling the feature. The tracking issue should list the subcrates impacted.
+
+For organization and discovery, a GitHub project can be used to track the stabilization issues.
+
+## Drawbacks
```
MrGVSV, line 56: I think another drawback could be that such unstable features may need constant upkeep based on their dependencies. And it might be difficult for the author(s) and other contributors to know that a stable change affects an unstable one.

Although this might just be the expectation.

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,82 @@
...
+    - Open a RFC, reaching a decision on the design
+    - Implement the RFC for the unstable feature
+
+
+
+Each PR will still need to be approved by Bevy community members, and would still need to have the same level of quality as other PRs merged.
+
+The PRs for the initial feature, various changes during its finalization and for the stabilization can be opened by different persons.
+
+The feature gate should not be exposed on the main `bevy` crate, but only on the subcrates where it is relevant. It can be enabled as a user by adding dependencies directly on the subcrates and enabling the feature. The tracking issue should list the subcrates impacted.
```
MrGVSV, line 52: Should we allow unstable features to opt-in to other unstable features? Or should they be self-contained (i.e. no unstable dependencies)?

rfcs/52-unstable_features.md
james7132: > I think a GitHub project showing tracking the stabilisation issues should be good enough for discovery

For comparison, Rust has the [RFC book](https://rust-lang.github.io/rfcs/). A GitHub Project might be good enough as a start, but IMO if we want this to be a publicly recognizable format, we need more than that in the long run.

rfcs/52-unstable_features.md
james7132: I think they should be both scoped and independent of each other. The combinatorial state space of dependent unstable features gets unwieldy real fast.

rfcs/52-unstable_features.md
mockersf: Having something like the RFC book requires quite a bit more work...

It is needed in Rust where they need to keep track of everything back to the pre 1.0 time for people still using an old version, they have feature opened 7 years ago not yet closed, and they track a lot of them.
This is not what I want to Bevy, where I would rather have around 10 unstable features open at any given time, and for no more than a year.

I think filtering issues on the label, and presenting them in a project would be enough for some time

rfcs/52-unstable_features.md
mockersf: good point, added!

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,84 @@
...
+- Something in between that would benefit from iterative work? Here come the unstable feature workflow for you!
+
+You will need to open a tracking issue to explain the new feature you want to add. If it looks good, you'll be ready to open that PR with a feature gate for your changes!
+
+## Implementation strategy
+
+The new workflow for this kind of feature would be:
+- Submit an issue explaining the feature you want to add under a feature gate. This will be the tracking issue for stabilization once approved. This issue gets a new label `S-Stabilization`.
+- If approved, submit a PR with the new feature and the feature gate
+- This PR can be merged by someone with the merge rights on the related part of Bevy as soon as it's `S-Ready-For-Final-Review`, without an approval by the main Bevy maintainers.
```
james7132, line 27: We should further expand this. Currently there are three people with merge rights, and only one of them is unscoped. This, at best, seems to still have a heavy bottleneck here. We already have engine teams for specific focus areas. Would it be possible to set it up such that areas of the engine could have these unstable features merged with significant focus area engine team buy-in? The only other way is to expand the number of org members with scoped merge rights, and that seems to be the slowest area of growth for the organization as a whole.

mockersf: I think that is out of scope of this RFC. I wrote it mentioning roles, who has those roles is not the subject. What you propose would still work in the context of this RFC, as it can be read as "giving the focus area teams the scope to merge approved unstable features PRs".

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,84 @@
...
+An unstable feature accepted in Bevy may stop working due to the merge of another PR. This is the same burden as keeping a PR up to date with Bevy while it's waiting to be merged, but more visible. On the positive side, it would be easier for someone else to jump in and fix it.
+
+## Rationale and alternatives
+
+- Improve the process to get new features in Bevy that would benefit from iterative work.
+- Alternatives:
+    - Continue as it is now, with semi large PRs that bitrot and RFCs that don't often move forward
+    - Work harder to implement a new feature as a third party crate (this is not always possible)
+    - Fork Bevy to merge unnaproved PRs and check how they work / interact
+
```
alice-i-cecile, line 73:
```suggestion
```

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Prior art
```
alice-i-cecile, line 75: Original article and the discussion here are worth analyzing too.

https://www.reddit.com/r/rust/comments/sy65f7/we_need_to_talk_about_rfcs/

rfcs/52-unstable_features.md
alice-i-cecile: I like both the upper and lower bounds here. The upper bound in particular results in a nice forcing function to actually ship things or cut our losses.

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,84 @@
...
+
+It will also increase complexity of a contribution that would like to use this process, but it is opt in and not mandatory. Think of it as a fast track for unstable features.
+
+Stabilizing a feature could be harder, as it would mean a review across several PRs, and may not easily work with GitHub review UI. But the stabilization review should be easier than a large PR as there would be actual usage feedback.
+
+An unstable feature accepted in Bevy may stop working due to the merge of another PR. This is the same burden as keeping a PR up to date with Bevy while it's waiting to be merged, but more visible. On the positive side, it would be easier for someone else to jump in and fix it.
+
+## Rationale and alternatives
+
+- Improve the process to get new features in Bevy that would benefit from iterative work.
```
alice-i-cecile, line 68:
```suggestion
```
I don't think this line adds anything.

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,84 @@
...
+- Alternatives:
+    - Continue as it is now, with semi large PRs that bitrot and RFCs that don't often move forward
+    - Work harder to implement a new feature as a third party crate (this is not always possible)
+    - Fork Bevy to merge unnaproved PRs and check how they work / interact
```
alice-i-cecile, line 72:
```suggestion
    - Fork Bevy to merge unapproved PRs and check how they work / interact
```

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,84 @@
...
+
+Stabilizing a feature could be harder, as it would mean a review across several PRs, and may not easily work with GitHub review UI. But the stabilization review should be easier than a large PR as there would be actual usage feedback.
+
+An unstable feature accepted in Bevy may stop working due to the merge of another PR. This is the same burden as keeping a PR up to date with Bevy while it's waiting to be merged, but more visible. On the positive side, it would be easier for someone else to jump in and fix it.
+
+## Rationale and alternatives
+
+- Improve the process to get new features in Bevy that would benefit from iterative work.
+- Alternatives:
+    - Continue as it is now, with semi large PRs that bitrot and RFCs that don't often move forward
```
alice-i-cecile, line 70:
```suggestion
    - Continue as it is now, with semi large PRs that bitrot and RFCs that don't often move forward
    - Grant scoped merge rights to teams / team leads for more areas of the Bevy engine, and tell users to deal with the instability
```

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,84 @@
...
+
+The feature gate should not be exposed on the main `bevy` crate, but only on the subcrates where it is relevant. It can be enabled as a user by adding dependencies directly on the subcrates and enabling the feature. The tracking issue should list the subcrates impacted.
+
+For organization and discovery, a GitHub project can be used to track the stabilization issues.
+
+## Drawbacks
+
+This will increase complexity in Bevy code.
+
+It will also increase complexity of a contribution that would like to use this process, but it is opt in and not mandatory. Think of it as a fast track for unstable features.
```
alice-i-cecile, line 60:
```suggestion
It will also increase complexity of a contribution that would like to use this process, but it is opt-in and not mandatory. Think of it as a fast track for unstable features.
```

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,84 @@
...
+
+The PRs for the initial feature, various changes during its finalization and for the stabilization can be opened by different persons.
+
+The feature gate should not be exposed on the main `bevy` crate, but only on the subcrates where it is relevant. It can be enabled as a user by adding dependencies directly on the subcrates and enabling the feature. The tracking issue should list the subcrates impacted.
+
+For organization and discovery, a GitHub project can be used to track the stabilization issues.
+
+## Drawbacks
+
+This will increase complexity in Bevy code.
```
alice-i-cecile, line 58: IMO we can significantly reduce this complexity by forcing feature flags to remain as localized as possible within the code base.

MiniaczQ:
How do you imagine CI/CD working with unstable features on the main bevy?

alice-i-cecile:
> How do you imagine CI/CD working with unstable features on the main bevy?

IMO one-experimental-feature at a time would be the strategy, for tests only. That'll increase load on our CI though, and may tip us over into "actually a priority to upgrade".

mockersf:
> How do you imagine CI/CD working with unstable features on the main bevy?

With only GitHub runners, no CI/CD for unstable features. This is actually already an issue in Bevy as we don't test non default configurations.

If we can get more capacity, we could run more feature gate combinations, but I don't think that should be blocking for merging a PR.
If a PR break an unstable feature:
- the PR author may fix it if they have the possibility
- or it can be reported on the unstable feature tracking issue, then someone with more knowledge on that feature can fix it

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,84 @@
...
+## User-facing explanation
+
+So you want to contribute a new feature to Bevy! You have a few paths available to you, depending on the scope of the feature:
+- A small feature with a clear implementation? Go ahead and open that PR!
+- A large feature with an open design space that could use some discussion about how to implement it? A RFC would be good.
+- Something in between that would benefit from iterative work? Here come the unstable feature workflow for you!
+
+You will need to open a tracking issue to explain the new feature you want to add. If it looks good, you'll be ready to open that PR with a feature gate for your changes!
+
+## Implementation strategy
```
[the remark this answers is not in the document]
mockersf: If the small engine tweaks are non controversial, they should be split in another PR, that's already what we try to promote

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,89 @@
...
+Each PR will still need to be approved by Bevy community members, and would still need to have the same level of quality as other PRs merged.
+
+The PRs for the initial feature, various changes during its finalization and for the stabilization can be opened by different persons.
+
+The feature gate should not be exposed on the main `bevy` crate, but only on the subcrates where it is relevant. It can be enabled as a user by adding dependencies directly on the subcrates and enabling the feature. The tracking issue should list the subcrates impacted.
+
+For organization and discovery, a GitHub project can be used to track the stabilization issues.
+
+While it should be avoided, if another PR breaks an unstable feature, it is not the PR author responsability to fix it. They may do it if they want, otherwise it must be reported on the unstable feature tracking issue.
+
```
cart, line 60: Not something to add to this RFC, but it would be useful to come up with some "current" PRs that would benefit from being unstable features.

alice-i-cecile:
Current PRs that would benefit from being unstable features:

- animation
- geometric primitives
- viewports
- raycasting
- Command error handling
- runtime system insertion
- window icons
- rumble support
- system graphs

I've tried to select things that I think are:
- large features
- things we probably want
-  complex APIs that should be experimented on, played with and iteratively developed

There are obviously other PRs that I think should just be merged outright (or merged after a bit of cleanup), but those are a distinct category.

james7132:
IMO anything that would introduce a new crate or engine capability (looking at animation here), should probably fit into the RFC category and go through a more rigorous design process.

Though that brings up question of ecosystem crates depending on unstable features. We should probably discourage ecosystem crates from developing stable features of their own on top of unstable bevy features.

cart:
I think it's worth calling out what our policy is on "competing implementations".  Are we going to allow merging 3 different competing unstable skeletal animation systems? What if they conflict with each other? If we only allow one, how do we pick a "winner" without going through a "full" review process (including pulling in "lead maintainers")?

I also think its worth explicitly calling out what should be a "3rd party crate" and what should be a "first party unstable feature". I personally think we should still push experimental / un-prioritized features to 3rd party crates when possible. 3rd party crates still allows for collaboration without bloating bevy internals (or bloating the bevy dev process / getting blocked on maintainer approval). It also more cleanly accommodates competing implementations, and in somes ways is a simpler user-facing conceptual model than "unstable bevy features". I think the decision tree should be "can this be expressed as a third party crate? if yes ... do that. If not (because it needs to make internal changes), use an unstable feature. Skeletal animation would probably need to be an "unstable feature" because the PBR pipeline can't (yet) easily accommodate 3rd party changes to mesh bindings (and im not convinced it should). On the other hand, "geometric primitives" and "system graphs" _can_ be (and have been) expressed as third party crates. Maybe to help incentivize and streamline the "experiment in 3rd party crates first" process, we could have a way to "boost" 3rd party crates that are currently being considered for inclusion in bevy?

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,89 @@
...
+    - Part of the unstable feature is larger than expected with several design possibilities
+    - Open a RFC, reaching a decision on the design
+    - Implement the RFC for the unstable feature
+
+
+Each PR will still need to be approved by Bevy community members, and would still need to have the same level of quality as other PRs merged.
+
+The PRs for the initial feature, various changes during its finalization and for the stabilization can be opened by different persons.
+
+The feature gate should not be exposed on the main `bevy` crate, but only on the subcrates where it is relevant. It can be enabled as a user by adding dependencies directly on the subcrates and enabling the feature. The tracking issue should list the subcrates impacted.
```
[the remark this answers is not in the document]
mockersf: I would prefer to not have different classes of features on the main crate. Currently we have a few to enable functionalities, feature gates for unstable features would be a different type

rfcs/52-unstable_features.md
mockersf: The PR that motivated me to write that RFC is https://github.com/bevyengine/bevy/pull/3884

As you said yourself
> But this feels like a "let people yell at us if we did something wrong" sort of situation.

This RFC generalises that possibility: to have an advanced feature that we want to test, and let the interested people manage until it's ready for common use

cart:
Another quick thought: as this is pitched as a way to resolve "scaling issues" and it does come with a number of tradeoffs (which are already called out), we should fully discuss alternatives first. I think @mockersf already called out the clearest alternative. Rather than creating a new class of "unstable" code, we could instead just horizontally scale merging "stable" code (ex: some type of "subject matter expert merge rights" approach). It would require a lot more care to maintain the integrity of the engine (especially from my perspective ... there have been a lot of things that many "active" community members have voted to merge that I wouldn't merge without tweaks). But I think its something we should be "growing into".

james7132:
> Another quick thought: as this is pitched as a way to resolve "scaling issues" and it does come with a number of tradeoffs (which are already called out), we should fully discuss alternatives first. I think @mockersf already called out the clearest alternative. Rather than creating a new class of "unstable" code, we could instead just horizontally scale merging "stable" code (ex: some type of "subject matter expert merge rights" approach). It would require a lot more care to maintain the integrity of the engine (especially from my perspective ... there have been a lot of things that many "active" community members have voted to merge that I wouldn't merge without tweaks). But I think its something we should be "growing into".

Just to call out in this thread that there has been a sizable discussion of these kinds of alternatives in https://github.com/bevyengine/bevy/discussions/4131

alice-i-cecile:
> there have been a lot of things that many "active" community members have voted to merge that I wouldn't merge without tweaks

One of the perspectives (that I'm not sure how I personally feel about) is that we take a merge-first, tweak-later approach, especially with more folks with merge rights.

The majority of PRs we receive are clear improvements over the existing situation, and while they aren't always perfect in terms of code quality / docs / API design, collaboration is generally much easier once the code is merged in (and we'll have initial user feedback). PRs-to-PRs are a thing, but there's a lot of steps and they're not obvious: without assurance that "yes this will get merged soon" it's often hard to prioritize work done to review or improve others' PRs.

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,89 @@
+# Feature Name: `unstable_features`
+
+## Summary
+
+Experimental new features are merged into main behind a feature gate before stabilization.
```
ickk, line 5: I would propose we call these experimental features something other than "unstable", since a "stable feature" implies that we won't change the API of that feature going forwards; we don't make that guarantee about basically anything in Bevy at this point.

Rust has this to say about their own "unstable/stable" feature process:
> Code that works and runs on stable should not break.

> Until [it has been stabilized], the feature is not set in stone: every part of the feature can be changed, or the feature might be completely rewritten or removed.

Even a "stable feature" under the text of this RFC, as with almost all features in Bevy right now, is subject to change and might be completely rewritten or removed.

Plenty of good alternatives: `experimental`, `prototype`, `canary`, `proto`, `future`, et c.

alice-i-cecile: prototype is my favorite of those. Definitely like it better than unstable in the context of Bevy.

ickk: I like it too - especially given the parallel to `bevy_prototype_<name>` community crates which, from other discussion here, it sounds like will still be encouraged where possible.

MrGVSV: This is a valid point and I’d like to cast my vote on `experimental`. I think the others are a bit more vague/formal/rigid whereas `experimental` feels more approachable. To me it says, "this is an _experiment_ 🧪 to gain feedback and toy around with implementations to see if we (as a community) want to pursue this feature or a feature like this."

rfcs/52-unstable_features.md
djeedai: > I think a GitHub project showing tracking the stabilisation issues should be good enough for discovery, what do you think?

This is hard to discover for Bevy users which are not contributors.

rfcs/52-unstable_features.md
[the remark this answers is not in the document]
djeedai: The small engine tweaks cannot always be purely additive, I can imagine sometimes they're breaking changes.

rfcs/52-unstable_features.md
djeedai: Can we mark this as resolved now that all maintainers have merge rights?

rfcs/52-unstable_features.md
```diff
@@ -0,0 +1,89 @@
...
+    - Implement part of the RFC as an unstable feature
+    - Feedback into the RFC with the gained knowledge of implementation and usage
+- Unstable Feature to RFC
+    - Open an tracking issue, start implementing it
+    - Part of the unstable feature is larger than expected with several design possibilities
+    - Open a RFC, reaching a decision on the design
+    - Implement the RFC for the unstable feature
+
+
+Each PR will still need to be approved by Bevy community members, and would still need to have the same level of quality as other PRs merged.
```
djeedai, line 51: Review this inline with more recent rules about maintainers and reviewers? The rules is 2 community reviews I think.

PR #53, opened 2022-03-09
Consistent Hierarchies

james7132:
[RENDERED](https://github.com/james7132/rfcs/blob/generalized-hierarchies/rfcs/53-consistent-hierarchy.md)

This is a summary of the discussion in bevyengine/bevy#4141 as a RFC. The primary goal here is to resolve the ergonomics and consistency issues of dealing with hierarchies in Bevy, with a secondary focus on providing performant access and mutation patterns for it.

### DIFF
--- a/rfcs/53-consistent-hierarchy.md
+++ b/rfcs/53-consistent-hierarchy.md
@@ -1,4 +1,4 @@
-# Feature Name: `generalized-hierarchies`
+# Feature Name: `consistent-hierarchy`
 
 ## Summary
 The current transform hierarchy operations currently allow for it to remain in a
@@ -90,6 +90,20 @@ follows:
    added or removed from a parent.
  - Create a `HiearchyQuery` custom SystemParam for more easily iterating over
    components in the hierarchy.
+
+This change will repurpose the following existing `EntityCommand` extension
+functions for this purpose:
+
+ - `add_child`
+ - `push_children`
+ - `insert_children`
+ - `remove_child` (this will be added as a wrapper around `remove_children`)
+ - `remove_children`
+
+An additional `set_parent(Option<Entity>)` command should also be added as a
+child-side way of altering the hierarchy, defaulting to adding the entity as the
+last child to the provided parent if provided, and detaching the child from it's
+parent if `None`.
 
 ### HierarchyQuery
 An extension to this design is to create dedicated `HierarchyQuery<Q, F>`, which
@@ -294,37 +308,6 @@ hierarchy related commands.
  - We lose cache locality for entities with few children. Adding a single child
    forces an allocation.
 
-### Type Markers in Components
-It may be possible to include type markers as a part of the hierarchy components
-to differentiate multiple hierachies from each other. This would allow an entity
-to be part of multiple hierarchies at the same time.
-
-```rust
-#[derive(Component)]
-pub struct Parent<T> {
-  parent: Entity,
-  _marker: PhantomData<T>,
-}
-
-#[derive(Component)]
-pub struct Children<T> {
-  children: SmallVec<[Entity, 8]>,
-  _marker: PhantomData<T>,
-}
-```
-
-#### Benefits
-This allows us to directly tie hierarchies to the components they're targetting.
-Instead of a general `Parent`, a `Parent<Transform>` declares it's meant
-for `bevy_transform`'s transform components.
-
-It also allows multiple overlapping hierarchies to be made that may differ from
-each other depending on the use case.
-
-#### Drawbacks
-This can be quite confusing to show in an editor and is reasonably complex to
-understand from a user perspective. A compelling use case might be needed here.
-
 ### Alternative: Hierarchical ECS Storage
 This involves directly adding a third ECS storage option. In this solution, a
 data-oriented SoA structure of split BlobVec columns is used, virtually identical
@@ -384,6 +367,42 @@ penalties for heavy use.
 
  - [GameObject.GetComponentsInChildren](https://docs.unity3d.com/ScriptReference/GameObject.GetComponentsInChildren.html)
  - [GameObject.GetComponentsInParent](https://docs.unity3d.com/ScriptReference/GameObject.GetComponentsInParent.html)
+
+## Future Work
+One of the immediate pieces of future work here is to patch up the potential
+footgun of using `EntityCommands::remove` on either `Parent` or `Children` via
+add/remove/changed hooks. This is currently mitigatable via documentation, but a
+complete solution prevents users from breaking the invariant in the first place.
+
+### Multiple Hierarchies via Type Markers
+It may be possible to include type markers as a part of the hierarchy components
+to differentiate multiple hierachies from each other. This would allow an entity
+to be part of multiple hierarchies at the same time.
+
+```rust
+#[derive(Component)]
+pub struct Parent<T> {
+  parent: Entity,
+  _marker: PhantomData<T>,
+}
+
+#[derive(Component)]
+pub struct Children<T> {
+  children: SmallVec<[Entity, 8]>,
+  _marker: PhantomData<T>,
+}
+```
+
+This allows us to directly tie hierarchies to the components they're targetting.
+Instead of a general `Parent`, a `Parent<Transform>` declares it's meant
+for `bevy_transform`'s transform components.
+
+It also allows multiple overlapping hierarchies to be made that may differ from
+each other depending on the use case.
+
+However, this may be quite confusing to show in an editor and is reasonably
+complex to understand from a user perspective. A compelling use case might be
+needed to justify this kind of change.
 
 ## Unresolved questions
 
--- a/rfcs/53-consistent-hierarchy.md
+++ b/rfcs/53-consistent-hierarchy.md
@@ -213,8 +213,9 @@ transforms for 2D and UI.
  - Updates are still not immediately visible within a stage and deferred until
    the next command buffer flush.
  - Hierarchy updates are now single threaded. Commands can be generated from
-   multiple systems at the same time, but they now require exclusive access to
-   apply.
+   multiple systems, [or from multiple threads in the same
+   system](https://github.com/bevyengine/bevy/pull/4749) at the same time, but
+   they now require exclusive access to apply.
  - Each of the changed commands do a lot more work before and are signifgantly
    more branch heavy. This could negatively impact the performance of
    command-heavy workloads (i.e.  scene loading)
--- a/rfcs/53-consistent-hierarchy.md
+++ b/rfcs/53-consistent-hierarchy.md
@@ -196,7 +196,12 @@ transforms for 2D and UI.
    issue, and all hiearchy traversals require heavy random access into memory.
  - Updates are still not immediately visible within a stage and deferred until
    the next command buffer flush.
- - Hierarchy updates are now single threaded.
+ - Hierarchy updates are now single threaded. Commands can be generated from
+   multiple systems at the same time, but they now require exclusive access to
+   apply.
+ - Each of the changed commands do a lot more work before and are signifgantly
+   more branch heavy. This could negatively impact the performance of
+   command-heavy workloads (i.e.  scene loading)
  - The commands applied for controlling the hierarchy are computationally more
    expensive, requiring more branching access and more `World::get` calls.
  - Some potential implementations of this hierarchy requires normal ECS commands
--- a/rfcs/53-consistent-hierarchy.md
+++ b/rfcs/53-consistent-hierarchy.md
@@ -131,30 +131,46 @@ any system that needs some form of hierachy traversal. These queries can be quit
 expensive to run, so the documentation for them should reflect this and
 discourage misuing it.
 
-### HierarchyEvents
-Three new events should be added here, each corresponding to a specific change
-in the hierarchy.
+### HierarchyEvent
+One new event should be added here, with enum variants corresponding to a
+specific change in the hierarchy:
+
+ - `ChildAdded` - Fired when a previous root entity has been added as a child to
+   a entity in the hierarchy.
+ - `ChildRemoved` - Fired when an entity is removed from the hierarchy and is now
+   a root entity.
+ - `ChildMoved` - Fired when an child is moved from one parent in the hierarchy
+   to another.
+
+These can be used to detect changes where `PreviousParent` is currently used.
+Each event is exclusive, meaning that any change only generates one event. A move
+will not fire additional separate `ChildAdded` and `ChildRemoved` events.
+
+Only one event type is used here to ensure that hierarchy alterations are
+properly sequenced.
+
+As hierarchy edits typically happen in bursts (i.e. scene loads). There may be
+large number of these events generated at once. To reclaim this memory, it may be
+necessary to add support for automatic garbage collection to events.
 
 ```rust
-pub enum ChildAdded {
-  pub parent: Entity,
-  pub child: Entity,
-}
-
-pub enum ChildRemoved {
-  pub parent: Entity,
-  pub child: Entity,
-}
-
-pub enum ChildMoved {
-  pub parent: Entity,
-  pub previous_parent: Entity,
-  pub new_parent: Entity,
+#[derive(Clone)]
+pub enum HierarchyEvent {
+  ChildAdded {
+    parent: Entity,
+    child: Entity,
+  },
+  ChildRemoved {
+    parent: Entity,
+    child: Entity,
+  },
+  ChildMoved {
+    parent: Entity,
+    previous_parent: Entity,
+    new_parent: Entity,
+  }
 }
 ```
-
-Both `ChildAdded` and `ChildRemoved` will be fired when a child moved occurs.
-These can be used to detect changes where `PreviousParent` is currently used.
 
 ### `bevy_hierarchy`-`bevy_transform` split.
 These hierarchy types should be moved to a new `bevy_hierarchy` crate to avoid
@@ -229,9 +245,10 @@ pub struct Hierarchy {
 ```
 
 This would be strikingly similar to the `SparseSet` used in `bevy_ecs`. The main
-difference here is that the parallel `Vec`s are kept in topological order. When
-iterating over the hierarchy, children that are found before their parent will
-swapped with their parents.
+difference here is that the parallel `Vec`s are roughly kept in topological order.
+Inserting should be `O(1)` as it just updates the new parent. Removing should be
+`O(n)` with the number of children of that `Entity`. At least once per frame, a
+system should run a topological sort over the resource if it's dirty.
 
 An alternative version of this hierarchy resource only stores parents. This may
 speed up iteration as individual relations are smaller during swaps, but finding
@@ -243,15 +260,29 @@ can be kept to O(n) in the next depth.
 
  - Iteration in hierarchy order (parents first) is entirely linear. This may make
    transform and other parent-reliant propagation signfigantly faster.
- - Can be used with the command-driven interface proposed in this design.
+ - Finding roots in the sorted hierarchy is both linear and compacted tightly in
+   memory.
+ - Removes `Parent` and `Children` ECS components entirely. Could further reduce
+   archetype fragmentation. The hierarchy is stored densely in a single location
+   and is not spread between multiple archetypes/tables.
+ - Can still be used with the command-driven interface proposed in this design
+   instead of using ECS storage.
 
 #### Drawbacks
 
  - Merging and splitting hierarchies loaded from scenes can be quite difficult to
    maintain.
- - Iteration requires mutative access to the resource, which can limit
-   parallelism of systems traversing the hierarchy. The alternative requires a
-   dedicated O(n) sweep over the hierarchy whenever it's mutated.
+ - The non-hierarchy components are still stored in an unordered way in ECS data.
+   Iterating over the hierarchy will still incur random access costs for anything
+   non-trivial.
+ - Requires a dedicated topological sort system. Mitigated by only sorting when
+   the hierarchy is dirty, or simply tracking which entities in the hierarchy are
+   dirty.
+ - Editing the hierarchy requires exclusive access to the resource, which can
+   will serialize all systems editing the hierarchy. This also prevents internal
+   parallelism (i.e. via `Query::par_for_each`) as `ResMut` cannot be cloned.
+   This can be partially mitigated by deferring edits via commands, as proposed
+   in this RFC.
 
 ### Alternative: Linked List Hierarchy
 Instead of using two separate components to manage the hierarchy, this
--- a/rfcs/53-consistent-hierarchy.md
+++ b/rfcs/53-consistent-hierarchy.md
@@ -378,6 +378,13 @@ footgun of using `EntityCommands::remove` on either `Parent` or `Children` via
 add/remove/changed hooks. This is currently mitigatable via documentation, but a
 complete solution prevents users from breaking the invariant in the first place.
 
+### Dedicated RectTransform and Transform2D systems
+Transform/GlobalTransform is not particularly well suited for encoding the
+hierarchy of UIs and 2D-only scenes. By splitting out the hierarchy from the
+exsisting transform system, dedicated transform types and systems for these two
+different types of hierarchies. This could also enable further parallelism by
+allowing each transform system to independently handle propagation.
+
 ### Multiple Hierarchies via Type Markers
 It may be possible to include type markers as a part of the hierarchy components
 to differentiate multiple hierachies from each other. This would allow an entity
--- a/rfcs/53-consistent-hierarchy.md
+++ b/rfcs/53-consistent-hierarchy.md
@@ -4,7 +4,7 @@
 The current transform hierarchy operations currently allow for it to remain in a
 inconsistent state. This RFC proposes a few changes to make it globally
 consistent at any given time, as well as make it easier to support non-Tranform
-uese cases.
+use cases.
 
 ## Motivation
 Hierarchies are present in all parts of game development. The one that
--- a/rfcs/53-consistent-hierarchy.md
+++ b/rfcs/53-consistent-hierarchy.md
@@ -114,8 +114,7 @@ querying for items in the hierarchy:
    query parameters that do not have a parent.
  - `parent(_mut)` - returns single result from the parent of an entity.
  - `iter_children(_mut)` - returns an iterator over `(parent, child)` results
-   from a query parameter. This does a depth-first traversal over the hierarchy
-   starting at some entity.
+   from a query parameter. This does not return Children of Children.
  - `iter_descendants(_mut)` - returns an iterator over `(parent, child)` results
    from a query parameter. This does a depth-first traversal over the hierarchy
    starting at some entity.
--- a/rfcs/53-generalized-hierarchy.md
+++ b/rfcs/53-generalized-hierarchy.md
@@ -137,7 +137,7 @@ querying for items in the hierarchy:
    from a query parameter. This bubbles up queries from the bottom to the top
    starting at some entity.
 
-As these operations are handled at any command buffer flush. The hierarchy
+As these operations are handled at any command buffer flush, the hierarchy
 maintenance system can be removed.
 
 ## Benefits
--- a/rfcs/53-generalized-hierarchy.md
+++ b/rfcs/53-generalized-hierarchy.md
@@ -105,7 +105,7 @@ This type has no public constructors and has no public APIs that allow for
 mutation of any of the internal data, only getters. The consistency between
 multiple instances of the component is maintained by a set of dedicated hierachy
 management commands. The following should be added as extensions to
-EntityCommands:
+`EntityCommands`:
 
   - `set_parent(Option<Entity>)` - sets the parent of a given entity.
       - If `Hierarchy` is on neither parent or child, it will be added and both
--- a/rfcs/53-generalized-hierarchy.md
+++ b/rfcs/53-generalized-hierarchy.md
@@ -1,8 +1,10 @@
 # Feature Name: `generalized-hierarchies`
 
 ## Summary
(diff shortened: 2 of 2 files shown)


rfcs/53-generalized-hierarchy.md
```diff
@@ -0,0 +1,273 @@
+# Feature Name: `generalized-hierarchies`
+
+## Summary
+Create a general hierarchy or hierarchies for Bevy worlds that is ergonomic to
```
alice-i-cecile, line 4: This really only covers the "objective". Good summaries should at least hint at how we got there.

rfcs/53-generalized-hierarchy.md
```diff
@@ -0,0 +1,273 @@
...
+   However, many of these use cases also share a similar pattern of detecting
+   changes and propagating them either down to their descendants or bubbling up a
+   message through their ancestors.
+
+All of these cases involve mutating, querying, and traversing these large trees
+or forests of linked entities. Throughout each of these operations, it's
+imperative that Bevy provides a globally consistent view of the
+hierarchy/hiearchies.
+
+Many of these operations are not cheap to do at runtime. Queries and traversals
```
alice-i-cecile, line 30:
```suggestion
Many of these operations are both abundantly common and moderately expensive. Queries and traversals
```

rfcs/53-generalized-hierarchy.md
```diff
@@ -0,0 +1,273 @@
...
+In this RFC, we will be using the following terms repeatedly:
+
+ - Parent: this refers to the upper level Entity relative to their children. Each
+   Entity can either have zero or one parents, but may have zero or more
+   children.
+ - Child: this refers to a lower level Entity relative to their parent. Any child
+   will have exactly one parent. Children can have children of their own.
+ - Sibling: a child that shares the same parent as another different entity.
+ - Roots: These are entities without parents. May or may not have children.
+ - Leaves: These are entities without children. May or may not have a parent.
```
alice-i-cecile, line 44:
```suggestion
 - **Parent:** this refers to the upper level Entity relative to their children. Each
   Entity can either have zero or one parents, but may have zero or more
   children.
 - **Child:** this refers to a lower level Entity relative to their parent. Any child
   will have exactly one parent. Children can have children of their own.
 - **Sibling:** a child that shares the same parent as another different entity.
 - **Roots:** These are entities without parents. May or may not have children.
 - **Leaves:** These are entities without children. May or may not have a parent.
```

rfcs/53-generalized-hierarchy.md
```diff
@@ -0,0 +1,273 @@
...
+aims to have only one: `Hierarchy`. This hierarchy component provides *read-only*
+access into an entity's immediate hierarchial relationships: its parent,
+previous sibling, next sibling, and it's first child. All of these may return
+`Option<Entity>`, signalling that such relationships do not exist. This creates
+a single component that creates a top-down forest of entity trees, where each
+entity forms a doubly-linked list of siblings. It may also cache infromation
+about the component's position within the hierarchy like it's depth.
+
+Mutations to the hierarchy: de-parenting, re-parenting, re-ordering of children,
+etc. can only be done via Commands. This enforces a globally consistent view at
```
alice-i-cecile, line 73:
```suggestion
etc. can only be done via commands. This enforces a globally consistent view at
```

rfcs/53-generalized-hierarchy.md
```diff
@@ -0,0 +1,273 @@
...
+  next: Option<Entity>
+  first_child: Option<Entity>,
+}
+```
+
+This type has no public constructors and has no public APIs that allow for
+mutation of any of the internal data, only getters. The consistency between
+multiple instances of the component is maintained by a set of dedicated hierachy
+management commands. The following should be added as extensions to
+EntityCommands:
```
alice-i-cecile, line 108:
```suggestion
`EntityCommands`:
```

rfcs/53-generalized-hierarchy.md
```diff
@@ -0,0 +1,273 @@
...
+        will be atomically updated to reflect this change.
+      - If the parent already had children, the new child will be prepended at
+        the start of the linked list, and the sibling pointers of both will be
+        updated.
+      - If the child already had a parent, that previous parent will be updated
+        too.
+      - A HierarchyEvent is added to the queue to reflect this update.
+  - `move_to_first(Entity)` - Moves one of the children to the front.
+  - `move_to_last(Entity)` - Moves one of the children to the end.
+  - `remove_all_children()` - Disconnects all children from a parent. All
```
alice-i-cecile, line 121:
```suggestion
  - `disconnect_children()` - Disconnects all children from a parent. All
```

rfcs/53-generalized-hierarchy.md
```diff
@@ -0,0 +1,273 @@
...
+    children become roots.
+      - A HierarchyEvent is added to the queue to reflect this update.
+
+All of these commands will also update the four marker components appropriately.
+Optionally, we could also mark these components as changed (even though they're
```
alice-i-cecile, line 126: Hmm. I think I like this.

In general, stuff like this belongs in the "Open Questions" section.

rfcs/53-generalized-hierarchy.md
```diff
@@ -0,0 +1,273 @@
...
+ZSTs) to leverage change detection.
+
+One optional extension to this design is to create dedicated `HierarchyQuery<Q,
+F>`, which is a custom SystemParam that wraps multiple Query objects to simplify
```
alice-i-cecile, line 130:
```suggestion
F>`, which is a custom `SystemParam` that wraps multiple `Query` objects to simplify
```

rfcs/53-generalized-hierarchy.md
```diff
@@ -0,0 +1,273 @@
...
+querying for items in the hierarchy:
+
+ - `iter_descendants(_mut)` - returns an iterator over `(parent, child)` results
+   from a query parameter. This does a depth-first traversal over the hierarchy
+   starting at some entity.
+ - `iter_ancestors(_mut)` - returns an iterator over `(parent, child)` results
+   from a query parameter. This bubbles up queries from the bottom to the top
+   starting at some entity.
+
+As these operations are handled at any command buffer flush. The hierarchy
```
alice-i-cecile, line 140:
```suggestion
As these operations are handled at any command buffer flush, the hierarchy
```

rfcs/53-generalized-hierarchy.md
```diff
@@ -0,0 +1,273 @@
...
+The biggest benefit to this new design is guarenteed global consistency. The
+hierarchy cannot be arbitrarily mutated between synchronization points.
+
+The second benefit is that any update to the hierarchy is "atomic" and is `O(1)`
+for the majority of the operations handled.
+
+## Drawbacks
+There are a notable number of drawbacks to this design:
+
+ - Counting children is now `O(n)`. Finding and editing children order is still
```
alice-i-cecile, line 153: It's a) not clear when we want to count children and b) not clear to me that the number of children will ever be greater than say 10 in realistic use cases.

rfcs/53-generalized-hierarchy.md
```diff
@@ -0,0 +1,273 @@
...
+   `O(n)`, that has not changed.
+ - The fact that the hierarchy is not optimally stored in memory is still an
+   issue, and all hiearchy traversals require heavy random access into memory.
+ - Updates are still not immediately visible and deferred until the next command
+   buffer flushes.
+ - Hierarchy updates are now single threaded.
```
alice-i-cecile, line 159: Can't we can add internal parallelism to resolve this?

rfcs/53-generalized-hierarchy.md
```diff
@@ -0,0 +1,273 @@
...
+## Rationale and alternatives
+The primary goal of this proposal is to get rid of the hierarchy maintenance
+system and the consistency issues that come with it's design, which it delivers
+on.
+
+### Keep `Parent`/`Children` as is. Implement Commands/Events.
+This simply keeps the components as is but restricts both of them to be
+read-only. The commands and events seen in this proposal can also be implemented
+on top of them. The internals of `Children` could be replaced with a `HashSet<T>`
+to keep operations close to `O(1)`, but a consistent ordering will be lost here.
```
alice-i-cecile, line 175: UI badly needs consistent ordering of children.

rfcs/53-generalized-hierarchy.md
```diff
@@ -0,0 +1,273 @@
...
+
+## Rationale and alternatives
+The primary goal of this proposal is to get rid of the hierarchy maintenance
+system and the consistency issues that come with it's design, which it delivers
+on.
+
+### Keep `Parent`/`Children` as is. Implement Commands/Events.
+This simply keeps the components as is but restricts both of them to be
+read-only. The commands and events seen in this proposal can also be implemented
+on top of them. The internals of `Children` could be replaced with a `HashSet<T>`
```
alice-i-cecile, line 174: I think that replacing this with a `HashSet` is likely to be dramatically slower in practice: asymptotics are irrelevant for tiny n.

rfcs/53-generalized-hierarchy.md
```diff
@@ -0,0 +1,273 @@
...
+        the start of the linked list, and the sibling pointers of both will be
+        updated.
+      - If the child already had a parent, that previous parent will be updated
+        too.
+      - A HierarchyEvent is added to the queue to reflect this update.
+  - `move_to_first(Entity)` - Moves one of the children to the front.
+  - `move_to_last(Entity)` - Moves one of the children to the end.
+  - `remove_all_children()` - Disconnects all children from a parent. All
+    children become roots.
+      - A HierarchyEvent is added to the queue to reflect this update.
```
alice-i-cecile, line 123:
```suggestion
      - A `HierarchyEvent` is added to the queue to reflect this update.
```

rfcs/53-generalized-hierarchy.md
```diff
@@ -0,0 +1,273 @@
...
+There are a notable number of drawbacks to this design:
+
+ - Counting children is now `O(n)`. Finding and editing children order is still
+   `O(n)`, that has not changed.
+ - The fact that the hierarchy is not optimally stored in memory is still an
+   issue, and all hiearchy traversals require heavy random access into memory.
+ - Updates are still not immediately visible and deferred until the next command
+   buffer flushes.
+ - Hierarchy updates are now single threaded.
+ - The marker components may cause heavy archetype fragmentation.
```
alice-i-cecile, line 160: I'm not convinced that these marker components are worth it. IMO their use is rare enough (and their external-implementation simple enough) that I would prefer to exclude them from the design to save overhead and complexity, and have helper methods on `HierarchyQuery` instead. If needed, we can even cache the entities that are e.g. "roots" in a resource that gets pulled in there.

rfcs/53-generalized-hierarchy.md
```diff
@@ -0,0 +1,273 @@
...
+ - Some potential implementations of this hierarchy requires normal ECS commands
+   to be aware of the hierarchy. For example, despawning a child in the middle of
+   the linked list without updating it's immediate siblings will violate several
+   invariants of the system.
+
+## Rationale and alternatives
```
alice-i-cecile, line 166: IMO this section should cover "Hierarchy in a resource" as well. Cart's reasoning against this in the linked thread was quite compelling, but this should be documented further here.

alice-i-cecile (changes requested):
A fair bit of editing feedback, along with some technical critiques.

Major thoughts:

1. This should explicitly call out the creation of a `bevy_hierarchy` crate, and the ability to use a single hierarchy for distinct properties.
2. I don't think there's enough discussion of property inheritance here. How do we handle inheriting seperate properties? How do we ensure those values stay in sync? Can we opt-out of inheriting properties? This was listed as one of the motivations, but the proposals don't address it.
3. We *really* need a half-decent benchmark suite for this: these decisions are very performance driven and discussing O(n) complexity for single digit n doesn't really make sense.
4. I'm strongly in favor of incrementally improving the hierarchy situation ASAP: there are a lot of subtle, painful bugs, and it's seriously blocking UI work.
5. I rather like the "commands-only" approach to mutation: this is much safer than the existing design.

rfcs/53-generalized-hierarchy.md
james7132: Part of this I think is also to ensure child uniqueness, which has been noted as an issue in the past. We can't just `push` a child on to a Vec and call it a day here. Though I do think the most egregious child counts will be seen in UI use cases (i.e. a list of items in a scroll view could easily number up several hundred if not properly controlled). Though even then the entire children array fits in cache, so it may not be that bad of a check.

rfcs/53-generalized-hierarchy.md
james7132: Good point. The fact that we're privating basically the entire API behind commands means we can privately use resources for caching this too.

rfcs/53-generalized-hierarchy.md
alice-i-cecile: Sure, but we can use other forms of sets that have better small-n performance :)

rfcs/53-generalized-hierarchy.md
alice-i-cecile: Upon reflection: I think we should kill these marker components entirely.

TheRawMeatball:
While I like the commands-only idea, the issue with it is it can't be fully enforced since despawning entities can easily break this system. The hooks mentioned in bevyengine/bevy#3742 is a possible solution to that.

alice-i-cecile:
> it can't be fully enforced since despawning entities can easily break this system

Can't we just use `RemovedComponents` to correct for this? I suppose we'd have to run a check each time after commands are processed.

james7132:
> While I like the commands-only idea, the issue with it is it can't be fully enforced since despawning entities can easily break this system. The hooks mentioned in [bevyengine/bevy#3742](https://github.com/bevyengine/bevy/issues/3742) is a possible solution to that.

It's a start, I guess. One of the main drawbacks listed is that bevy_ecs is not aware of the structure, and will require something like hooks to fully polish out. However, the merits of this current design still stand, even if not a complete 100% solution in it's current form.

james7132:
Made an initial draft implementation that addresses a number of the issues found here, listed under one of the current alternatives in the RFC. This seems to be the absolute minimum changes to get some of the benefits from this RFC.

With some of the optimizations being made to the transform propagation system (see bevyengine/bevy#4180), it's probably best to minimize `Query/World::get` usage when traversing the hierarchy, which the main proposed changes here do the exact opposite of.

I'll update the RFC itself with these changes soon.

rfcs/53-generalized-hierarchy.md
james7132: Removed the mention of these given the recent optimization insights.

rfcs/53-generalized-hierarchy.md
james7132: I can think of quite a few cases where 10 is reasonable and realistic, but this is no longer an issue as I switched the design up to keep `Parent`/`Children`.

rfcs/53-generalized-hierarchy.md
james7132: Unless we somehow parallelize commands, I don't see how this is possible.

rfcs/53-generalized-hierarchy.md
james7132: Addressed and mentioned. Added "HashSet Children Internals" as an alternative and mentioned this as a drawback.

rfcs/53-generalized-hierarchy.md
james7132: Added!

rfcs/53-generalized-hierarchy.md
james7132: done!

BoxyUwU:
It was not super clear to me whether this rfc is propsing to make `insert` special cased to update the hierarchy, or whether to have separate `insert_child`/`insert_parent`-type commands. I don't have a preference either way as once hooks are a thing we can have `.insert::<Parent>` `.remove::<Children>` just work:tm: without special casing the commands for bevy's built in hierarchy.

Overall I like the changes here towards making the hierarchy more robust by having everything be immutable and requiring changes to be made via commands, this is stuff we will have to do even in a future with hooks/relations :)

The `Type Markers in Components` section does not really seem like an "alternative" it seems like an additional change we could make ontop of the changes this RFC is arguing for so I don't think theres point in me leaving comments about `Children<T>` and `Parent<T>`

james7132:
Updated the title/filename and clarified which commands we will be using or adding.

rfcs/53-consistent-hierarchy.md
```diff
@@ -0,0 +1,410 @@
+# Feature Name: `consistent-hierarchy`
+
+## Summary
+The current transform hierarchy operations currently allow for it to remain in a
+inconsistent state. This RFC proposes a few changes to make it globally
+consistent at any given time, as well as make it easier to support non-Tranform
+uese cases.
```
MinerSebas, line 7:
```suggestion
use cases.
```

rfcs/53-consistent-hierarchy.md
```diff
@@ -0,0 +1,410 @@
...
+
+ - `roots(_mut)` - returns an iterator over entities that match the provided
+   query parameters that do not have a parent.
+ - `parent(_mut)` - returns single result from the parent of an entity.
+ - `iter_children(_mut)` - returns an iterator over `(parent, child)` results
+   from a query parameter. This does a depth-first traversal over the hierarchy
+   starting at some entity.
+ - `iter_descendants(_mut)` - returns an iterator over `(parent, child)` results
+   from a query parameter. This does a depth-first traversal over the hierarchy
+   starting at some entity.
```
MinerSebas, line 121: Currently, they sound like they do the exact same thing.

```suggestion
 - `iter_children(_mut)` - returns an iterator over `(parent, child)` results
   from a query parameter. This does not return Children of Children.
 - `iter_descendants(_mut)` - returns an iterator over `(parent, child)` results
   from a query parameter. This does a depth-first traversal over the hierarchy
   starting at some entity.
```

As for `iter_descendants` which `depth-first traversal` does it perform?
Inorder (Left, Root, Right), Preorder (Root, Left, Right) or Postorder (Left, Right, Root)?
I would expect Preorder here.

rfcs/53-consistent-hierarchy.md
```diff
@@ -0,0 +1,410 @@
...
+   developer and third-party crates to make their own independent transform
+   system without compiling the default transform system.
+
+## Drawbacks
+
+ - The fact that the hierarchy is not optimally stored in memory is still an
+   issue, and all hiearchy traversals require heavy random access into memory.
+ - Updates are still not immediately visible within a stage and deferred until
+   the next command buffer flush.
+ - Hierarchy updates are now single threaded.
```
MinerSebas, line 200: Not 100% correct, the existing Hierarchy update system already only used a single thread, though it could be run in parallel with other systems.

rfcs/53-consistent-hierarchy.md
james7132: Good point. It should be pre-order, and  iter_children should also be from first to last. Will update the text to reflect that.

rfcs/53-consistent-hierarchy.md
james7132: Haha! Here is definitely where the current inconsistencies come in. Maintenance is single threaded, mutation could be parallelized, by using disjoint queries for Parent in multiple systems. I'll update the text to reflect this.

nicopap:
On a purely use-facing perspective, the proposal is solid. I think it solves a lot of issues. And doesn't introduce any meaningful limitations. As a heavy user of the hierarchy system, it seems fine. I'll have a go with a multiple-hierarchy system RFC similar to the one introduced here. I got multiple use cases where this could be very useful.

rfcs/53-consistent-hierarchy.md
```diff
@@ -0,0 +1,421 @@
...
+pub enum ChildRemoved {
+  pub parent: Entity,
+  pub child: Entity,
+}
+
+pub enum ChildMoved {
+  pub parent: Entity,
+  pub previous_parent: Entity,
+  pub new_parent: Entity,
+}
```
nicopap, line 153:
```suggestion
pub enum HierarchyEvent {
  ChildAdded {
    parent: Entity,
    child: Entity,
  },
  ChildRemoved {
    parent: Entity,
    child: Entity,
  },
  ChildMoved {
    parent: Entity,
    previous_parent: Entity,
    new_parent: Entity,
  },
}
```

You probably meant this, right?

james7132: The intention is to keep them separate so listeners can independently listen to each kind of event separately instead of listening on instances of all 3. Plus moving a child suggests the other two.

rfcs/53-consistent-hierarchy.md
```diff
@@ -0,0 +1,421 @@
...
+   `Children` or `Parent` will break the invariants of the system without some
+   hooks to enforce the invariants on removal. This is partially mitigatable via
+   documentation until hooks or relations lands.
+
+## Rationale and alternatives
+The primary goal of this proposal is to get rid of the hierarchy maintenance
+system and the consistency issues that come with it's design, which it delivers
+on.
+
+### Alternative: Hierarchy as a Resource
```
cart, line 218: The biggest downside here is definitely scene management. We'd need to either:

1. Add the concept of "resource merging". In practice I think this could be reasonably cheap, especially if we assume that scenes have a single root entity (and I'm relatively convinced at this point that they should).
2. Add a secondary component-based hierarchy representation for scenes, which then gets "converted" into a runtime centralized Hierarchy resource when the scene is spawned.

Both of these are fundamental changes to the mental model that require serious thought.

> Iteration requires mutative access to the resource, which can limit parallelism of systems traversing the hierarchy. The alternative requires a dedicated O(n) sweep over the hierarchy whenever it's mutated.

We don't necessarily need to couple these operations. I get that we're deferring topological sorts to avoid really expensive insertion (we don't want to be constantly shifting items around on each insert). But this is an optimization issue, not a correctness issue. We could choose to do it once per frame with `traverse_hierarchy_and_toposort(&mut self)` (maybe in the transform propagation) and encourage users to just do `traverse_hierarchy(&self)` unless they really want to toposort earlier. On average, iterations would be approximately topo-sorted. Or we could have a "dirty" bit and just topo sort on frames where it is dirty.

rfcs/53-consistent-hierarchy.md
```diff
@@ -0,0 +1,421 @@
...
+
+#### Drawbacks
+
+ - Merging and splitting hierarchies loaded from scenes can be quite difficult to
+   maintain.
+ - Iteration requires mutative access to the resource, which can limit
+   parallelism of systems traversing the hierarchy. The alternative requires a
+   dedicated O(n) sweep over the hierarchy whenever it's mutated.
+
+### Alternative: Linked List Hierarchy
```
cart, line 256: Linked list `Hierarchy` components are growing on me. Immediate, always-in-sync O(1) hierarchy updates are compelling. I've fixated on needing `Without<Parent>` queries in the past, but thats only important if we allow more than one root and/or treat "hierarchy membership / rootness" as a direct result of Hierarchy data. If we force explicit roots, this becomes a non-issue.

This creates a number of problems though:
1. How do we choose roots? Do we just spawn a single "canonical" root entity on our users' behalf? Do users denote a `Root` component on entities? Do we just ignore "hierarchy orphans" for things like transform propagation? Do we detect `parent == None` cases for entities without `Root` markers and warn?
2. Immediate hierarchy updates only work when both parent and child entities already have Hierarchy components. We would still need to rely on deferred parenting commands for the initial spawn of entities, or for entities that don't yet have the hierarchy component.
3. In light of (2), to actually make "immediate hierarchy updates" possible and worthwhile in practice, we would need to ensure that all common entities/bundles/prefabs intended to be hierarchy members have Hierarchy when they are spawned. This would mean a lot more "unnecessary" / empty hierarchy data in the ECS and more complicated bundles. Users defining their own bundles would need to add Hierarchy to get the "immediate" benefits. This feels brittle / in some cases users might not be sure if the parent is "ready", so they might need to try both (or just default to commands).
4. We would need to measure the extra `Query::get` child traversal perf relative to pre-computed child list iteration and decide if it is "worth it".

rfcs/53-consistent-hierarchy.md
```diff
@@ -0,0 +1,421 @@
...
+As suggested by the generic parameters, the query results and filters are
+identical to the ones used by a normal Query or QueryState.
+
+The main motivation for this is to make creating systems like
+`transform_propagate_system` easier. These queries can be generically used for
+any system that needs some form of hierachy traversal. These queries can be quite
+expensive to run, so the documentation for them should reflect this and
+discourage misuing it.
+
+### HierarchyEvents
```
cart, line 134: I think its worth pointing out that in practice, this will increase the memory usage of the hierarchy system to be something like `2 * size_of::<Entity>() * MAX_HIERARCHY_CHANGES_ACROSS_TWO_FRAMES`. Given that hierarchical spawning for scenes will be pretty burst-ey, in practice this will be a per-hierarchical-entity `2 * size_of::<Entity>()` price unless we add garbage collection to Events.

Although that should be weighed against the cost of `PreviousParent(Entity)`, which includes both additional memory footprint and (sometimes) additional runtime costs for archetype moves.

cart (commented):
As a short term solution to our hierarchy woes, this seems like a reasonable / scoped approach that we could probably land before the next release.

rfcs/53-consistent-hierarchy.md
cart: The downside to separate queues here is that order matters for these events. ChildRemoved -> ChildAdded is different from ChildAdded -> ChildRemoved. If the goal is to maintain some consistent "view" of the hierarchy, I think we need a centralized queue.

I'd also like some clarity on when ChildMoved vs ChildRemoved+ChildAdded events will be sent.

cart:
I'm down to approve and merge this once all outstanding comments have been resolved. The event conversation is the most important to sort out.

rfcs/53-consistent-hierarchy.md
james7132: `PreviousParent` is a `size_of::<Entity>()` memory cost and arbitrary archetype moves. Given that `(Global)Transform` is 48 bytes, and one additional `Entity` is 8, just in memory write costs alone, we'd need to write 6x (16 bytes vs 96) the events to compare an archetype move for your typical transform hierarchy change, and that's not counting the archetype/table move overhead, the other components being moved, and the additional archetype fragmentation from the two-fold moves. It's just we incur the cost of `PreviousParent` one frame later instead of immediately.

Building automatic event garbage collection seems perfectly workable given that events already track readers.

rfcs/53-consistent-hierarchy.md
james7132: Great point, updated to match this. Also this should really simplify the event code (looking at the prototype PR) haha.

rfcs/53-generalized-hierarchy.md
james7132: With the advent of https://github.com/bevyengine/bevy/pull/4749, this is mostly (?) mitigated. The exclusive nature of commands is likely not going to change anytime soon.

rfcs/53-consistent-hierarchy.md
james7132: Updated the RFC to reflect some of these + some pros/cons I initially missed. I actually like this a lot more now that it's clear that this even further addresses the archetype fragmentation and makes hierarchy edits on par with a SparseSet lookup.

If we know the roots of a scene, assuming we don't need to ensure it's sorted immediately afterwards, copy-pasting the mapped contents of a scene hierarchy at the end of the "global hierarchy" then updating the spawned scene roots to their new parent(s) should be `O(n)` in the number of entities in the hierarchy, and should be very fast. It can then rely on the frame-to-frame topological sort to optimize out the kinks.

rfcs/53-consistent-hierarchy.md
james7132: After doing some work to parallelize transform propagation, I like this approach a lot less. Responses inline.

> How do we choose roots? Do we just spawn a single "canonical" root entity on our users' behalf? Do users denote a Root component on entities? Do we just ignore "hierarchy orphans" for things like transform propagation? Do we detect parent == None cases for entities without Root markers and warn?

A SparseSet marker component, maintained by the commands proposed here could work. We're already incurring heavy random read costs, so archetypal iteration at the root level likely isn't going to be much of a perf impact here.

I'd strongly recommend against a single root. Any propagation system we have now is parallelizing at the root level, and is reliant on efficiently identifying them and going wide splitting on the individual trees in the forest.

Other than the marker component, the only way I can think of is a resource root, which introduces more consistency problems.

> Immediate hierarchy updates only work when both parent and child entities already have Hierarchy components. We would still need to rely on deferred parenting commands for the initial spawn of entities, or for entities that don't yet have the hierarchy component.

The design here still calls for a command-based approach, so iteration is read-only. The `O(1)` nature just changes the costs of running those commands. Within those commands, we can update the state of the new child/sibling/parents by adding the `Hierarchy` component.

> In light of (2), to actually make "immediate hierarchy updates" possible and worthwhile in practice, we would need to ensure that all common entities/bundles/prefabs intended to be hierarchy members have Hierarchy when they are spawned. This would mean a lot more "unnecessary" / empty hierarchy data in the ECS and more complicated bundles. Users defining their own bundles would need to add Hierarchy to get the "immediate" benefits. This feels brittle / in some cases users might not be sure if the parent is "ready", so they might need to try both (or just default to commands).

Agreed here. This seems like a major downside. I'll add this to the RFC.

> We would need to measure the extra Query::get child traversal perf relative to pre-computed child list iteration and decide if it is "worth it".

This is the real kicker. `Query::get` has a lot of overhead when you're using it to iterate over a large portion of the `World`. The tight cache-locality of `Children` simplifies this by a long shot.

cart:
> Great point, updated to match this and address the potential duplication issues. Also this should really simplify the event code (looking at the prototype PR) haha.

I'm not seeing these changes. Did you push them?

james7132:
Looks like I didn't. Sorry about that! Should be up now.

PR #54, opened 2022-04-07
Global Task Pools

james7132:
[RENDERED](https://github.com/james7132/rfcs/blob/global-task-pools/rfcs/54-global-task-pools.md)

This proposes changes to the way the three main TaskPools are initialized and used in Bevy.

### DIFF
--- a/rfcs/54-global-task-pools.md
+++ b/rfcs/54-global-task-pools.md
@@ -1,0 +1,118 @@
+# Feature Name: `global-task-pools`
+
+## Summary
+Globally initialize the newtyped TaskPools (ComputeTaskPool,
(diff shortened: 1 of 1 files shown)


cart (approved):
I'm convinced! The ergonomics wins are hard to ignore and given that this is pretty much the pattern adopted everywhere, it seems like a safe bet. Globals do create "conflict" problems and feel more than a little bit dirty. I generally don't want to embrace this pattern in Bevy, but task pools do seem like an exceptional case. As we've recently discussed, I do think that the unified task pool in https://github.com/bevyengine/bevy/pull/4740 is the right general direction, but I'm comfortable merging this in the current state as it feels orthogonal from a design standpoint.

PR #56, opened 2022-05-20
bevy_reflect: Unique `Reflect`

nicopap:
[RENDERED](https://github.com/nicopap/rfcs/blob/bevy-reflect-api/rfcs/56-better-reflect.md)

- create a new trait: `PartialReflect`
- `Reflect: PartialReflect`
- All methods on `Reflect` that **does not** depend on the uniqueness of the
  underlying type are moved to `PartialReflect`
- Most things should now accept a `dyn PartialReflect` rather than
  `dyn Reflect`.
- It is possible to convert a `PartialReflect` into a `Reflect` using a
  `into_full` method.
- `FromReflect` becomes `FromPartialReflect`.

## Unresolved questions
- [X] `CanonReflect` name? ⇒ Went with `Reflect` and `PartialReflect`
- [X] `CanonReflect` or `ReflectProxy`? Remove all underlying-dependent methods from
  `Reflect` and adding them to a new trait, or keep the underlying-dependent
  methods but moved all the non-dependent methods to a new trait. Which one to
  go with? ⇒ Went with a bit of both.
- [X] Implementation (partial: https://github.com/bevyengine/bevy/pull/7207)
- [ ] Consider making more things `Reflect`.
- [X] Is a `TypeRegistry` entry for `FromPartialReflect` necessary? (no, as per partial implementation)

### DIFF
--- a/rfcs/56-better-reflect.md
+++ b/rfcs/56-better-reflect.md
@@ -1,0 +1,261 @@
+# bevy_reflect: Unique `Reflect`
+
+This builds on https://github.com/bevyengine/bevy/pull/4042 and the `TypeInfo`
+API.
+
+## Summary
+
+Add a `ReflectProxy` enum that generalizes the `Dynamic***` structs from
+`bevy_reflect` to all types, and remove `Reflect` implementation for
+`Dynamic***` structs.
+
+I also suggest renaming all `Dynamic***` into `***Proxy` and the reflect
+traits `clone_dynamic` methods to `proxy_clone`.
+
+## Lexicon/Terminology/Nomenclature/Jargon
+
+#### pass for
+
+We say that `T: Reflect` **passes for** `U: Reflect` when `T`'s `type_name` is
+equal to `U`'s, and it is possible to convert from a `T` to a `U` using
+`apply`, `set` or `FromReflect::from_reflect`.
+
+#### underlying
+
+The **underlying** value of a dynamic object (eg: `dyn Reflect`) is the
+concrete type backing it. In the following code, the **underlying type** of
+`reflect_a` is `MysteryType`, while the **underlying value** is the value of
+`a`.
+
+```rust
+let a = MysteryType::new();
+let reflect_a: Box<dyn Reflect> = Box::new(a.clone()) as Box<dyn Reflect>;
+```
+
+When using `clone_value`, the underlying type changes: if `MysteryType` is a
+struct, the underlying type of `reflect_clone` will be `DynamicStruct`:
+
+```rust
+let reflect_clone: Box<dyn Reflect> = reflect_a.clone_value();
+```
+
+#### Underlying-sensible method
+
+Those are methods of `Reflect` that act differently between `reflect_clone` and
+`reflect_a` from the previous examples.
+
+
+#### `Dynamic***`
+
+The structures holding generic structural information about some proxied type. 
+* `DynamicMap`
+* `DynamicList`
+* `DynamicTuple`
+* `DynamicStruct`
+* `DynamicTupleStruct`
+
+#### Reflect subtraits
+
+The top level traits in `bevy_reflect`: `Struct`, `TupleStruct`, `Tuple`,
+`List` and `Map`. They are all subtraits of `Reflect`.
+
+## Motivation
+
+The `Reflect` trait is a trap, it doesn't work as expected due to `Dynamic***`
+mixing up with the "regular" types under the `dyn Reflect` trait object.
+
+To solve this, we do not allow more than a single underlying type to "pass for"
+a concrete type in `Reflect`. This implies that the various `Dynamic***` types
+shouldn't implement `Reflect`.
+
+When the underlying type of a `Box<dyn Reflect>` is a `Dynamic***`, a lot of
+the `Reflect` methods become invalid, such as `any`, `reflect_hash`,
+`reflect_partial_eq`, `serializable`, `downcast`, `is` or `take`. However,
+those methods are provided regardless.
+
+### What's under the `dyn Reflect`?
+
+Currently, `Reflect` has very a confusing behavior. Most of the methods on
+`Reflect` are highly dependent on the underlying type, regardless of whether
+they are supposed to stand for the requested type or not.
+
+For example
+
+```rust
+let a = MysteryType::new(); // suppose MysteryType implements Reflect
+let reflect_a: Box<dyn Reflect> = Box::new(a.clone()) as Box<dyn Reflect>;
+let dynamic_a: Box<dyn Reflect> = a.clone_value();
+// This more or less always panic
+assert_eq!((&*reflect_a).type_id(), (&*dynamic_a).type_id())
+```
+
+This is because multiple different things can pretend to be the same thing. In
+itself this shouldn't be an issue, but the current `Reflect` API is leaky, and
+you need to be aware of both this limitation and the underlying type to be able
+to use the `Reflect` API correctly.
+
+
+### `reflect_hash` and co. with `Dynamic***`
+
+The problem is not limited to `type_id`, but also extends to the `reflect_*`
+methods.
+
+```rust
+let a = MysteryType::new(); // suppose MysteryType implements Reflect
+let reflect_a: Box<dyn Reflect> = Box::new(a.clone()) as Box<dyn Reflect>;
+let dynamic_a: Box<dyn Reflect> = a.clone_value();
+let reflect_hash = reflect_a.reflect_hash();
+let dynamic_hash = dynamic_a.reflect_hash();
+// This more or less always panic if `MysteryType` implements `Hash`
+assert_eq!(reflect_hash, dynamic_hash)
+```
+
+
+## Proposition
+
+The problem stems from the duplicity of `dyn Reflect`. We want to rely on the
+underlying methods of the same thing to be unique.
+
+To solve this, we do not allow more than a single underlying type to "pass for"
+a concrete type in `Reflect`. This implies that the various `Dynamic***` types
+shouldn't implement `Reflect`.
+
+## User-facing explanation
+
+* `Reflect` maps 1-to-1 with an underlying type meaning that if a
+  `Box<dyn Reflect>` cannot be downcasted to `T`, you cannot build a `T` from
+  it.
+* There is a `ReflectProxy` enum you can use to construct any `Reflect` type
+  based on some dynamic structure. This can be used, for example, for
+  deserialization of types that implement `Reflect` but not `Deserialize`.
+  * You can "clone reflectively" a `Box<dyn Reflect>` by using
+    `Reflect::proxy_clone`. It returns a `ReflectProxy`.
+  * You can use `ReflectProxy::sidecast` to create a `Box<dyn Reflect>` from
+    a `ReflectProxy`. The created trait object's underlying type is a concrete
+    type you can downcast to.
+* `FromReflect` doesn't exist anymore. Instead, `Reflect::downcast` provides
+  now the same guarentees that `FromReflect::from_reflect` had. If you need to
+  get a `T` from a `ReflectProxy`, use a combination of
+  `ReflectProxy::sidecast_type<T>` and `downcast`.
+
+
+## Implementation strategy
+
+The `Reflect` API stays identical to the one we have today, at the exception of
+`clone_value`.
+
+We remove the `Dynamic***: Reflect` implementations and we introduce a
+`ReflectProxy` enum.
+
+```rust
+pub enum ReflectProxy {
+    Struct(DynamicStruct),
+    TupleStruct(DynamicTupleStruct),
+    Tuple(DynamicTuple),
+    List(DynamicList),
+    Map(DynamicMap),
+    AlreadyReflect(Box<dyn Reflect>),
+}
+```
+
+`ReflectProxy` mirrors the `ReflectRef` and `TypeInfo` enums.
+
+`ReflectProxy` has a method to convert the underlying value into the
+value it is proxying. This allows you to get a `Reflect` from a `ReflectProxy`
+while upholding the `Reflect` uniqueness rule.
+
+```rust
+impl ReflectProxy {
+  pub fn type_name(&self) -> &str {
+    match self { /* etc. */ }
+  }
+  pub fn can_sidecast(&self, registry: &TypeRegistry) -> Result<(), SidecastError> {}
+  pub fn sidecast(self, registry: &TypeRegistry)
+    -> Result<Box<dyn Reflect>, SidecastError> {
+    let type_name = self.type_name();
+    let registration = registry.get_with_name(type_name).ok_or(SidecastError::Unregistered)?;
+    Ok(registration.constructor(self)?)
+  }
+  pub fn sidecast_type<T: Reflect + Typed>(self) -> Result<Box<dyn Reflect>, SidecastError> {
+    let registration = TypeRegistration::of::<T>();
+    Ok(registration.constructor(self)?)
+  }
+}
+```
+ 
+`sidecast` requires introducing a new field to `TypeRegistration`:
+```rust
+pub struct TypeRegistration {
+    short_name: String,
+    // new: constructor
+    constructor: fn(ReflectProxy) -> Result<Box<dyn Reflect>, ConstructorError>,
+    data: HashMap<TypeId, Box<dyn TypeData>>,
+    type_info: TypeInfo,
+}
+impl TypeRegistration {
+  pub fn constructor(&self, proxy: ReflectProxy) ->  Result<Box<dyn Reflect>, ConstructorError> {
+    (self.constructor)(proxy)
+  }
+  // ...
+}
+```
+
+`constructor` would simply construct the `T` and wrap it into a `Box` before
+returning it.
+
+`constructor` can be derived similarly to `TypeInfo`. It will however
+require additional consuming methods on `Dynamic***` structs in order to be
+able to move their `Box<dyn Reflect>` fields into the constructed data
+structure.
+
+## Drawbacks
+
+- Users cannot define their own proxy type, they must use the `Dynamic***`
+  types.
+- `Dynamic***` do not implement the `Reflect` subtraits anymore.
+- There is some added complexity (although it fixes what I believe to be a
+  major limitation of the `bevy_reflect` API)
+- `Reflect::apply` now only accepts a `ReflectProxy` (see next section)
+
+## Rationale and alternatives
+
+- Divide `Reflect` in two distinct traits, one with the underlying-sensible
+  method, and one with the other methods. (see next points)
+- An earlier version of this RFC proposed `ReflectProxy` as a **trait**,
+  however, I reformulated it as an enum when considering how to implement it.
+  (How would you define the `TypeRegistration::constructor` field if passed a
+  `Box<dyn ReflectProxy>`? You'd need something like `ReflectProxyRef`?)
+
+I believe this change is needed, since `Reflect` duplicity is a large footgun.
+
+I don't believe that having `Dynamic***` implement `Reflect` makes sense, I
+think it introduces much more issues than it solves. The only use-case for
+`Dynamic***` being `Reflect` are the `reflect_mut`, `reflect_ref` and `apply`
+methods. However, those could as well be implemented outside of the `Reflect`
+trait, as mentioned in the "Future possibilities" section.
+
+## Unresolved questions
+
+- How does this interact with `#[reflect(ignore)]`?
+- Is a `trait ReflectProxy` more desirable than an `enum` and how to implement
+  it?
+- Should `Dynamic***` inner values be `Box<ReflectProxy>` or `Box<dyn Reflect>`?
+  With `Box<dyn Reflect>`:
+  * You must have a `TypeRegistry` and you must call `sidecast` both when
+    _building_ the `Dynamic***` and when converting from `ReflectProxy` to
+    `Reflect` (note that you most often already need the `TypeRegistry` when
+    building the `Dynamic***` if you want to access `TypeInfo`)
+  * The heavy lifting is mostly done when constructing the `ReflectProxy`,
+    `sidecast` is only a shallow conversion, calling `downcast` on the top
+    level data structure fields.
+  * There is less change to do if we keep it as it is today.
+  * With `Box<ReflectProxy>`, the opposite is true.
+- Should we try to implement the reflect subtraits for the `Dynamic***`?
+  (this would require removing the `Subtrait: Reflect` bounds)
+
+## Future possibilities
+
+We might be able to move `reflect_ref` and `reflect_mut` out of `Reflect` into
+its own trait. (`Overlay`), make `Reflect: Overlay` and implement `Overlay` for
+the `Dynamic***` types and `ReflectProxy`. `apply` could then accept a 
+`T: Overlay` instead of a `ReflectProxy`.
--- a/rfcs/56-better-reflect.md
+++ b/rfcs/56-better-reflect.md
@@ -5,7 +5,7 @@
 - Add a `CanonReflect: Reflect` trait that replaces `Reflect`
 - Remove from `Reflect` all methods that assume uniqueness of underlying type.
 - Add a `from_reflect` method to `Typed` that takes a `Box<dyn Reflect>` and
-  returns a `Self` 
+  returns a `Self` to convert a `Reflect` into a `CanonReflect`.
 
 ## Lexicon/Terminology/Nomenclature/Jargon
 
@@ -115,7 +115,7 @@ but to keep to scope of this RFC short, we will keep them in `Reflect`.
 
 ### `CanonReflect` trait
 
-We remove all methods methods assuming uniqueness and `Any` trait bound from
+We remove all methods assuming uniqueness and `Any` trait bound from
 `Reflect` and only implement them on "canonical" types: ie: the ones implemented
 by the `#[derive(Reflect)]` macros.
 
@@ -146,10 +146,11 @@ checking if the underlying type is the one we expect OR converting into the
 underlying type described by the `Reflect::type_name()` method.
 
 To do so, it's necessary to either provide an expected type or access the type
-registry to invoke the constructor associated with teh `Reflect::type_name()`.
+registry to invoke the constructor associated with the `Reflect::type_name()`.
 
-To define a constructor We move the `from_reflect` method from `FromReflect` to `Typed`. For no
-particular reasons but to reduce how many traits we export from `bevy_reflect`.
+To define a constructor We move the `from_reflect` method from `FromReflect` to
+`Typed`. For no particular reasons but to reduce how many traits we export from
+`bevy_reflect`.
 
 **Danger**: Using `type_name` from `Reflect` is error-prone. For the conversion
 to pick up the right type, we need `type_name()` to match exactly the one
@@ -167,11 +168,11 @@ trait Typed {
 }
 pub struct TypeRegistration {
     short_name: String,
+    data: HashMap<TypeId, Box<dyn TypeData>>,
+    type_info: TypeInfo,
     // new: constructor
     constructor: fn(Box<dyn Reflect>)
         -> Result<Box<dyn CanonReflect>, Box<dyn Reflect>>,
-    data: HashMap<TypeId, Box<dyn TypeData>>,
-    type_info: TypeInfo,
 }
 ```
 
@@ -189,7 +190,7 @@ impl TypeRegistration {
     pub fn make_canon(&self, reflect: Box<dyn Reflect>)
         -> Result<Box<dyn CanonReflect>, Box<dyn Reflect>>
     {
-        (self.constructor)(proxy)
+        (self.constructor)(reflect)
     }
     // ...
 }
@@ -232,7 +233,7 @@ impl dyn Reflect {
     pub fn make_canon<T: Typed>(self: Box<Self>)
         -> Result<Box<dyn CanonReflect>, Box<dyn Reflect>>
     {
-        T::from_reflect(from).map(|t| Box::new(t))
+        T::from_reflect(self).map(|t| Box::new(t))
     }
 }
 ```
@@ -245,24 +246,16 @@ only `Reflect`.
 
 ## User-facing explanation
 
-- `CanonReflect` maps 1-to-1 with an underlying type meaning that if a
-  `Box<dyn Reflect>` cannot be downcasted to `T`, you cannot build a `T` from
-  it.
-- `FromReflect` doesn't exist anymore. Instead, `Reflect::downcast` provides
-  now the same guarantees that `FromReflect::from_reflect` had. If you need to
-  get a `T` from a `ReflectProxy`, use a combination of
-  `ReflectProxy::sidecast_type<T>` and `downcast`.
+`Reflect` let you navigate any type regardless independently of their shape
+with the `reflect_ref` and `reflect_mut` methods. `CanonReflect` is a special
+case of `Reflect` that also let you convert into a concrete type using the
+`Any` methods. To get a `CanonReflect` from a `Reflect`, you should use any
+of the `make_canon` methods on `dyn Reflect`, `TypeRegistry` or
+`TypeRegistration`.
 
-
-`sidecast` requires introducing a new field to `TypeRegistration`:
-
-`constructor` would simply construct the `T` and wrap it into a `Box` before
-returning it.
-
-`constructor` can be derived similarly to `TypeInfo`. It will however
-require additional consuming methods on `Dynamic***` structs in order to be
-able to move their `Box<dyn Reflect>` fields into the constructed data
-structure.
+Any type derived with `#[derive(Reflect)]` implements `CanonReflect`. The only
+types that implements `Reflect` but not `CanonReflect` are "proxy" types such
+as `DynamicStruct`, or any third party equivalent.
 
 ## Drawbacks
 
@@ -275,7 +268,7 @@ structure.
 - The addition of `constructor` to `Typed` will increase the size of generated
   code.
 - `constructor` makes `Typed` not object safe. But it should be fine
-  since `Typed` is not intended to be used with dynamic trait objects.
+  since `Typed` is not intended to be used as a dynamic trait objects.
 
 ## Unresolved questions
 
@@ -286,12 +279,18 @@ structure.
   Technically, `Canon` doesn't mean "normal representation", as "canonical"
   does, but it's short and close enough to be understood. I thought about
   using `canonize` instead of `make_canon`, but it felt too religious.
+- An earlier version of this RFC had the exact opposite approach to fix the
+  uniqueness issue: Instead of removing all underlying-dependent methods from
+  `Reflect` and adding them to a new trait, we kept the underlying-dependent
+  methods but moved all the non-dependent methods to a new trait. Which one to
+  go with?
 
 ## Future possibilities
 
 - Add a `target_type(&self) -> TypeId` or `full_type_name` method to `Reflect`
   such that it's easier to check if we are converting into the proper canonical
-  type.
+  type. We might also benefit from a method that is capable of taking two
+  type paths and comparing them accounting for shortcuts.
 - Performance: currently we force-convert from `CanonReflect` to `Reflect`,
   then deeply traverse the data structure for all our reflect methods. It could
   be greatly optimized if we add a way to tell if the underlying type is
--- a/rfcs/56-better-reflect.md
+++ b/rfcs/56-better-reflect.md
@@ -1,13 +1,22 @@
-# bevy_reflect: `CanonReflect` as a unique `Reflect`
+# bevy_reflect: `Reflect` as a unique `Reflect`
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,261 @@
...
+`constructor` can be derived similarly to `TypeInfo`. It will however
+require additional consuming methods on `Dynamic***` structs in order to be
+able to move their `Box<dyn Reflect>` fields into the constructed data
+structure.
+
+## Drawbacks
+
+- Users cannot define their own proxy type, they must use the `Dynamic***`
+  types.
+- `Dynamic***` do not implement the `Reflect` subtraits anymore.
```
MrGVSV, line 215: I wouldn't call this a drawback since this is the whole reason for this RFC.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,261 @@
...
+  however, I reformulated it as an enum when considering how to implement it.
+  (How would you define the `TypeRegistration::constructor` field if passed a
+  `Box<dyn ReflectProxy>`? You'd need something like `ReflectProxyRef`?)
+
+I believe this change is needed, since `Reflect` duplicity is a large footgun.
+
+I don't believe that having `Dynamic***` implement `Reflect` makes sense, I
+think it introduces much more issues than it solves. The only use-case for
+`Dynamic***` being `Reflect` are the `reflect_mut`, `reflect_ref` and `apply`
+methods. However, those could as well be implemented outside of the `Reflect`
```
MrGVSV, line 234: I'd argue that the actual biggest use-case for them is deserialization. Since we can't guarantee that all types implement `FromReflect` we're forced to return a `Dynamic***`.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,261 @@
...
+trait, as mentioned in the "Future possibilities" section.
+
+## Unresolved questions
+
+- How does this interact with `#[reflect(ignore)]`?
```
MrGVSV, line 239: From what I gather, it really just encapsulates what `FromReflect` did. I imagine we'll have to do a similar thing. The tricky part is handling stuff beyond just `Default`, such as `FromWorld`.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,261 @@
...
+- Is a `trait ReflectProxy` more desirable than an `enum` and how to implement
```
MrGVSV, line 240: Personally, I much prefer the enum. It's easier to work with and make sense of, and since we know the exact number of variants available, it just makes more sense to me.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,261 @@
...
+  it?
+- Should `Dynamic***` inner values be `Box<ReflectProxy>` or `Box<dyn Reflect>`?
```
MrGVSV, line 242: I think it would have to be `Box<ReflectProxy>`, right? Since creating the proxy requires fields be recursively `clone_proxy`'d.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,261 @@
...
+  * With `Box<ReflectProxy>`, the opposite is true.
+- Should we try to implement the reflect subtraits for the `Dynamic***`?
+  (this would require removing the `Subtrait: Reflect` bounds)
+
+## Future possibilities
+
+We might be able to move `reflect_ref` and `reflect_mut` out of `Reflect` into
+its own trait. (`Overlay`), make `Reflect: Overlay` and implement `Overlay` for
+the `Dynamic***` types and `ReflectProxy`. `apply` could then accept a 
+`T: Overlay` instead of a `ReflectProxy`.
```
MrGVSV, line 261: How could something like `DynamicStruct` implement `Overlay` if it doesn't implement `Struct`? What would it return for `reflect_ref`?

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,261 @@
...
+    `sidecast` is only a shallow conversion, calling `downcast` on the top
+    level data structure fields.
+  * There is less change to do if we keep it as it is today.
+  * With `Box<ReflectProxy>`, the opposite is true.
+- Should we try to implement the reflect subtraits for the `Dynamic***`?
+  (this would require removing the `Subtrait: Reflect` bounds)
+
+## Future possibilities
+
+We might be able to move `reflect_ref` and `reflect_mut` out of `Reflect` into
```
MrGVSV, line 258: Do we need to move these methods out?

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,261 @@
...
+`ReflectProxy` has a method to convert the underlying value into the
+value it is proxying. This allows you to get a `Reflect` from a `ReflectProxy`
+while upholding the `Reflect` uniqueness rule.
+
+```rust
+impl ReflectProxy {
+  pub fn type_name(&self) -> &str {
+    match self { /* etc. */ }
+  }
+  pub fn can_sidecast(&self, registry: &TypeRegistry) -> Result<(), SidecastError> {}
```
MrGVSV, line 172: What does this method do? Is it just checking that a type is registered?

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,261 @@
...
+`ReflectProxy` enum.
+
+```rust
+pub enum ReflectProxy {
+    Struct(DynamicStruct),
+    TupleStruct(DynamicTupleStruct),
+    Tuple(DynamicTuple),
+    List(DynamicList),
+    Map(DynamicMap),
+    AlreadyReflect(Box<dyn Reflect>),
```
MrGVSV, line 157: We could just keep this as `Value` (maybe `Primitive` or something similar). I think a user will be able to ascertain that a variant that returns `Box<dyn Reflect>` is already an implementor of `Reflect`. At least, that's my opinion.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,261 @@
...
+  deserialization of types that implement `Reflect` but not `Deserialize`.
+  * You can "clone reflectively" a `Box<dyn Reflect>` by using
+    `Reflect::proxy_clone`. It returns a `ReflectProxy`.
+  * You can use `ReflectProxy::sidecast` to create a `Box<dyn Reflect>` from
+    a `ReflectProxy`. The created trait object's underlying type is a concrete
+    type you can downcast to.
+* `FromReflect` doesn't exist anymore. Instead, `Reflect::downcast` provides
+  now the same guarentees that `FromReflect::from_reflect` had. If you need to
+  get a `T` from a `ReflectProxy`, use a combination of
+  `ReflectProxy::sidecast_type<T>` and `downcast`.
```
MrGVSV, line 139: This will be really nice. Though, I'm sure we'll run into a lot of the same questions surrounding `FromReflect` as well.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,261 @@
...
+* `DynamicMap`
+* `DynamicList`
+* `DynamicTuple`
+* `DynamicStruct`
+* `DynamicTupleStruct`
+
+#### Reflect subtraits
+
+The top level traits in `bevy_reflect`: `Struct`, `TupleStruct`, `Tuple`,
+`List` and `Map`. They are all subtraits of `Reflect`.
```
MrGVSV, line 60: Forgetting the newly added `Array` I see 😜

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,261 @@
...
+API.
+
+## Summary
+
+Add a `ReflectProxy` enum that generalizes the `Dynamic***` structs from
+`bevy_reflect` to all types, and remove `Reflect` implementation for
+`Dynamic***` structs.
+
+I also suggest renaming all `Dynamic***` into `***Proxy` and the reflect
+traits `clone_dynamic` methods to `proxy_clone`.
```
MrGVSV, line 13: I like the idea of a rename. The one potential issue with it is that it's not inherently true. 

You can use `DynamicStruct` without ever needing to convert to a "proxied type". It can just be a tool to use in general and on its own apart from proxying.

MrGVSV (commented):
This is really great! I think you hit the nail on the head with one of reflection's biggest footguns right now: guarantees that a `dyn Reflect` is actually the type you think it is.

Using something like `ReflectProxy`, I think the API will just be a lot safer and easier to use. This involves a bit more work on the users' end (more passing around of the `TypeRegistry`), but overall I think it's pretty worth it.

Good work!

rfcs/56-better-reflect.md
nicopap: `sidecast` is consuming (takes `Box<Self>`). So maybe you want to make sure you can actually sidecast before calling it (note that `Reflect::downcast` returns a `Result` where the `Err` is the original value if it can't be converted). I imagine it as a "dry run" version of `sidecast`.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,261 @@
...
+```rust
+impl ReflectProxy {
+  pub fn type_name(&self) -> &str {
+    match self { /* etc. */ }
+  }
+  pub fn can_sidecast(&self, registry: &TypeRegistry) -> Result<(), SidecastError> {}
+  pub fn sidecast(self, registry: &TypeRegistry)
+    -> Result<Box<dyn Reflect>, SidecastError> {
+    let type_name = self.type_name();
+    let registration = registry.get_with_name(type_name).ok_or(SidecastError::Unregistered)?;
```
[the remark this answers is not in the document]
nicopap: So there are two failure mode:
1. The type itself is not registered
2. The struct the `Dynamic***` is supposed to represent cannot be constructed from the the fields that were added to it. (for example we have `struct Foo { bar: usize }` and a `DynamicStruct::add_field("baz", Box::new("Hi".to_owned()))`)

In the case of storing `Box<ReflectProxy>` instead of `Box<dyn Reflects` in `Dynamic***` fields, the errors could happen when trying to convert the nested value inside one of the fields.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,261 @@
...
+    level data structure fields.
+  * There is less change to do if we keep it as it is today.
+  * With `Box<ReflectProxy>`, the opposite is true.
+- Should we try to implement the reflect subtraits for the `Dynamic***`?
+  (this would require removing the `Subtrait: Reflect` bounds)
+
+## Future possibilities
+
+We might be able to move `reflect_ref` and `reflect_mut` out of `Reflect` into
+its own trait. (`Overlay`), make `Reflect: Overlay` and implement `Overlay` for
```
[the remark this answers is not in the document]
nicopap: My first draft had `StructuralExplorable`. I swapped it to `Overlay` last minute :smile: . As you mentioned in a different comment, there is more than one usecase for `Dynamic***`. I tried to capture the one used in the `Reflect::apply` method. You are right `Overlay` is pretty bad, it doesn't work with the `reflect_ref` usecase.

rfcs/56-better-reflect.md
nicopap: I think it's important to have some sort of replacement method on `Dynamic***` that does what `reflect_ref` does. Since it's not `Reflect` anymore, we'd lose those methods. Moving the `reflect_ref` to a trait that is both implemented by `dyn Reflect` (does rust even allow that???) and `ReflectProxy` would solve that.

rfcs/56-better-reflect.md
nicopap: The proxy would only be shallow with `Box<dyn Reflect>`. No recursion. I personally prefer `Box<dyn Reflect>` because the ways `sidecast` can fail is limited and more predictable. While if you have a deeply nested element that cannot be sidecast, it's going to be hard to attach the correct context to the error type, and it's going to be annoying to debug.

rfcs/56-better-reflect.md
nicopap: I think _they should_ implement `Struct`, `Tuple` etc. to be fair. The `Overlay` discussion goes in more details.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,261 @@
...
+
+To solve this, we do not allow more than a single underlying type to "pass for"
+a concrete type in `Reflect`. This implies that the various `Dynamic***` types
+shouldn't implement `Reflect`.
+
+## User-facing explanation
+
+* `Reflect` maps 1-to-1 with an underlying type meaning that if a
+  `Box<dyn Reflect>` cannot be downcasted to `T`, you cannot build a `T` from
+  it.
```
bjorn3, line 127: Won't that prevent `Reflect` from working for non-rust types like in case of scripting?

rfcs/56-better-reflect.md
MrGVSV: Ah sorry I missed the subtrait part. Yes that could potentially be a drawback.

rfcs/56-better-reflect.md
MrGVSV: How do you clone the value of a `Box<dyn Reflect>`, though? Assuming it's underlying type is a complex structure like a struct. Won't it need the deeply nested proxies?

rfcs/56-better-reflect.md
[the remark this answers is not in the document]
MrGVSV: Maybe `ReflectLike` or `PartialReflect`? Idk something that ties it back to reflection would be good I think 😅

rfcs/56-better-reflect.md
MrGVSV: Yeah I see that point. I’m just hesitant on having `Reflect` be unique but not doing the same for its subtraits. Again, calling `as_reflect` on a `dyn Struct` should just work, but can't if the underlying type is a `DynamicStruct` according to this RFC.

rfcs/56-better-reflect.md
MrGVSV: That makes sense and seems pretty reasonable to have. 

On the subject, though, could `sidecast` return both the error type and the owned value?

rfcs/56-better-reflect.md
MrGVSV: Well it couldn't be a struct unless it's marked as `#[reflect_value]`. So the only reflected types you should ever get in those cases are "value" types that don't fall into any of the other categories.

So I still think it should be consistent with that idea (and with the other enums). Again, this might just be a matter of preference, though.

rfcs/56-better-reflect.md
nicopap: Could you elaborate. How do you use a non-rust type in rust? Wouldn't it need a conversion at one point?

bjorn3: The idea would be that you have for example a `PyObject` which is how any python object is internally represented and then it's `Reflect` implementation would present a different type depending on the type the `PyObject` has on the python side. For example for

```python
class Foo(object):
    __slots__ = ('a', 'b')

class Bar(object):
    __slots__ = ('c', 'd')
```

a `PyObject` holding a `Foo` would present the fields `a` and `b`, while a `PyObject` holding a `Bar` would present the fields `c` and `d` and a `PyObject` holding an array would present itself as an array through `Reflect`.

MrGVSV: But this only exists on the Python side, right? What would this map to in Rust currently if there's no matching struct?

bjorn3: It doesn't need to map to a rust type I think. `downcast` could return None for any type you attempt to downcast to. `type_name` would give the python name. It shouldn't map to `DynamicStruct` as that would lose the python type.

MrGVSV: What's the actual Rust type though? What are you passing around? Could _that_ type be reflected?

Like `PyObject` -To Rust-> `PyObjectContainer` -Reflected-> `dyn Reflect` -Downcast-> `PyObjectContainer` -To Python-> `PyObject`.

Sorry if I’m misunderstanding 😅

bjorn3: That would work, but it would have the exact same issue as what the `Dynamic*` types have right now in that the downcasted type doesn't match the semantic type that determines the layout which `.reflect()` gives.

rfcs/56-better-reflect.md
nicopap: That is an issue! Calling `Reflect::proxy_clone` and then `ReflectProxy::sidecast` them would work. I think the `Box<dyn Reflect>` vs `Box<ReflectProxy>` is a question of whether we want to force evaluation at construction time, or have supporting both usecases. It seems in the bevy spirit to support both use cases, so `Box<ReflectProxy>` seems now the better option to me.

rfcs/56-better-reflect.md
nicopap: Yeah I think I get the issue here. What about the proposition of moving `reflect_ref` and `reflect_mut` to their own trait independent from `Reflect` (call it the `Bavy` trait) and also move all the `Reflect` subtraits to `Bavy`? You could then implement `Bavy` for `PyObject` (or a wrapper type, per orphan rule). Maybe also move the `sidecast` proposition and implement `Bavy` for `ReflectProxy`?

bjorn3: I would rather go the opposite way and move all methods except `reflect_ref` and `reflect_mut` into separate traits as not every type may have an implementation for them and `reflect_ref` and `reflect_mut` are the core reflection api.

rfcs/56-better-reflect.md
nicopap: I think not. The user can build arbitrary struct, upcast it to `Box<dyn Reflect>` and insert it there. And I think it would be a reasonable thing to do (eg: fine grained control of when to convert which part of a `ReflectProxy` into `Reflect`).

MrGVSV: Okay I think I understand. So like in the code below?

```rust
let mut dyn_struct = DynamicStruct::default();
dyn_struct.insert("foo", Box::new(SomeStruct::new()));
let field_proxy = dyn_struct.field("foo");
assert!(matches!(field_proxy, ReflectProxy::AlreadyReflect(..)));
```

> Pretend `DynamicStruct` stores `Box<ReflectProxy>`

I think if that's the case, then I do agree that `AlreadyReflect` is a readable and reasonable naming choice. 

We could _maybe_ consider just `Reflect` or `Boxed` for the sake of keystrokes, but yeah `AlreadyReflect` is pretty good already.

rfcs/56-better-reflect.md
nicopap: @bjorn3  I think this is a good idea. Maybe I could revise this RFC to have a `UniqueReflect` vs `Reflect`?

colepoirier:
> palliate

@nicopap could you clarify this by using a simpler word than “palliate”?

rfcs/56-better-reflect.md
kabergstrom: why would you want to reuse Reflect as a trait for this as opposed to writing specific code for Python? what's the use case? I thought Reflect as a trait is supposed to model how you can represent data in Rust, not in any language ever.

rfcs/56-better-reflect.md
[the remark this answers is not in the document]
PROMETHIA-27: I think `Inspectable` is a good name for it. Probably would want to rename the functions then

rfcs/56-better-reflect.md
PROMETHIA-27: I think it's entirely possible that a custom implementation of `ReflectProxy` might be preferable, but I also can't think of any concrete examples.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,261 @@
...
+  * You must have a `TypeRegistry` and you must call `sidecast` both when
+    _building_ the `Dynamic***` and when converting from `ReflectProxy` to
+    `Reflect` (note that you most often already need the `TypeRegistry` when
+    building the `Dynamic***` if you want to access `TypeInfo`)
+  * The heavy lifting is mostly done when constructing the `ReflectProxy`,
+    `sidecast` is only a shallow conversion, calling `downcast` on the top
+    level data structure fields.
+  * There is less change to do if we keep it as it is today.
+  * With `Box<ReflectProxy>`, the opposite is true.
+- Should we try to implement the reflect subtraits for the `Dynamic***`?
```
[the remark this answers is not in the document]
PROMETHIA-27: What part of the subtraits requires `Subtrait: Reflect`? They're basically just interfaces for basic data structures like a `Dict<String, Box<dyn Reflect>>` or a `Vec<Box<dyn Reflect>>`. I don't think it's necessary to keep the `: Reflect` bound.

rfcs/56-better-reflect.md
PROMETHIA-27: (see above; I prefer `Inspectable` to `Overlay`)
I think `DynamicStruct` *should* implement `Struct`; with that, it would return an `Inspectable<'a>(&'a dyn Struct)` containing a reference to itself.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,310 @@
...
+- Performance: currently we force-convert from `CanonReflect` to `Reflect`,
+  then deeply traverse the data structure for all our reflect methods. It could
+  be greatly optimized if we add a way to tell if the underlying type is
+  already canon. (For example `make_canon` could be a no-op if the underlying
+  type is already the canonical form)
+  \
+  This also interacts interestingly with `ReflectRef` and `ReflectMut`. Should
+  the `Value` variant hold a `dyn Reflect` or a `dyn CanonReflect`?
+- `impl<T: Struct> Reflect for T {}` and other subtraits rather than
+  `Struct: Reflect`.
```
MrGVSV, line 303: How would this work? Wouldn't this prevent implementing/deriving `Reflect` on implementors of `Struct`?

rfcs/56-better-reflect.md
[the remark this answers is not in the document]
MrGVSV: But say you had a `&mut dyn Struct`, there would be no way to call `clone_value` or `apply` on it. You'd only be able to do that if you had access to the original `&mut dyn Reflect` (even though they're the same value under the hood).

rfcs/56-better-reflect.md
MrGVSV: Yeah I prefer `Inspectable` over `Overlay` as well.

rfcs/56-better-reflect.md
MrGVSV: What would the implementation look like to make that work, though?

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,310 @@
...
+
+**Danger**: Using `type_name` from `Reflect` is error-prone. For the conversion
+to pick up the right type, we need `type_name()` to match exactly the one
+returned by `std::any::type_name`. Ideally we have an alternative that defines
+the underlying type without ambiguity, such as `target_type_id` or
+`full_type_name`.
+
+We introduce a new field to `TypeRegistration`: `constructor`. It allows
+converting from a `Reflect` to a `CanonReflect`, so that it's possible to
+use its underlying-sensible methods.
```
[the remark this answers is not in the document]
nicopap: Yeah. I think we can avoid using `FromReflect` for this, if we add a `into_canon` method to `PartialReflect`.

nicopap:
Next revision:
* `CanonReflect`, `Reflect` ⇒ `Reflect`, `PartialReflect`
* add `into_canon` to `PartialReflect`
* Remove `FromReflect` change (though it will be renamed `FromPartialReflect`)

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,310 @@
...
+  type. We might also benefit from a method that is capable of taking two
+  type paths and comparing them accounting for shortcuts.
+- Performance: currently we force-convert from `CanonReflect` to `Reflect`,
+  then deeply traverse the data structure for all our reflect methods. It could
+  be greatly optimized if we add a way to tell if the underlying type is
+  already canon. (For example `make_canon` could be a no-op if the underlying
+  type is already the canonical form)
+  \
+  This also interacts interestingly with `ReflectRef` and `ReflectMut`. Should
+  the `Value` variant hold a `dyn Reflect` or a `dyn CanonReflect`?
```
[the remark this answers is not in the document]
nicopap: I was thinking: maybe add another variant to the `ReflectRef` and co. enums? It could be useful for example partially "evaluating" a struct tree.

rfcs/56-better-reflect.md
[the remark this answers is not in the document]
nicopap: It's still problematic though. Since there is no way to get a `CanonReflect` from a `Reflect` "proxy" this way. The only way I can see is if we add the `FromReflect::from_reflect` method to the type registry the same way we do for other traits.

rfcs/56-better-reflect.md
nicopap: Yes. But with a lowered surface API (only `reflect_ref` and `reflect_mut`), there is not much point to implement `Reflect` in any specific way for implementors of `Struct`. There might be other issues, especially with `into_full`, but it's for the future.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,310 @@
...
+```rust
+pub trait CanonReflect: Reflect + Any {
+  fn any_mut(&mut self) -> &mut dyn Any {
+    // implementation
+  }
+  // etc.
+}
+```
+
+This trait is automatically implemented by the `#[derive(Reflect)]` macro.
```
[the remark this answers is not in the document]
nicopap: This is probably solved with the new draft using the `PartialReflect` naming

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,310 @@
...
+## Drawbacks
+
+- Users must be aware of the difference between `CanonReflect` and `Reflect`,
+  and it must be explained.
+- You cannot directly convert a `Reflect` into a `Any` or `T`, you must first
+  make sure it is conforms a canonical representation by using either
+  `TypeRegistry::make_canon` or `<dyn Reflect>::make_canon<T: Typed>(self:
+  Box<self>)`.
+- The addition of `constructor` to `Typed` will increase the size of generated
+  code.
```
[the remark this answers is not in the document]
nicopap: Correct. Updated the draft to not rely on `FromReflect` though

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,310 @@
...
+a way to get a `CanonReflect` from a `Reflect`. This is only possible by
+checking if the underlying type is the one we expect OR converting into the
+underlying type described by the `Reflect::type_name()` method.
+
+To do so, it's necessary to either provide an expected type or access the type
+registry to invoke the constructor associated with the `Reflect::type_name()`.
+
+To define a constructor We move the `from_reflect` method from `FromReflect` to
+`Typed`. For no particular reasons but to reduce how many traits we export from
+`bevy_reflect`.
```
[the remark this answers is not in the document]
nicopap: I've updated the RFC to not rely on `FromReflect`.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,310 @@
...
+    pub fn make_canon<T: Typed>(self: Box<Self>)
+        -> Result<Box<dyn CanonReflect>, Box<dyn Reflect>>
+    {
+        T::from_reflect(self).map(|t| Box::new(t))
+    }
+}
+```
+
+Note that [trait upcasting] would allow us to remove the `as_reflect` and
+`as_reflect_mut` methods. In fact, I don't think they are necessary at all.
```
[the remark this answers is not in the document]
nicopap: Good point, removed that from the RFC

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,318 @@
...
+
+    fn as_partial_reflect(&self) -> &dyn PartialReflect;
+    fn as_partial_reflect_mut(&mut self) -> &mut dyn PartialReflect;
+    fn reflect_ref(&self) -> ReflectRef;
+    fn reflect_mut(&mut self) -> ReflectMut;
+
+    fn apply(&mut self, value: &dyn PartialReflect);
+    // Change name of `clone_value`
+    // Ideally add documentation explaining that the underlying type changes.
+    fn clone_partial(&self) -> Box<dyn PartialReflect>;
```
MrGVSV, line 208: This is a good idea. I think this makes it much clearer that this can return a proxy type rather than the original. Although, it should also be pointed out in the documentation that this doesn't always return proxy types (i.e. for value types like `i32`).

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,318 @@
...
+  knowing the underlying type.
+- This splits the reflect API, that used to be neatly uniform. It will be
+  more difficult to combine various use cases of `bevy_reflect`, requiring
+  explicit conversions. (Note that however, the previous state of things would
+  result in things getting silently broken)
+
+## Unresolved questions
+
+- Should `Reflect` still be named `Reflect`, does it need to be a subtrait of
+  `PartialReflect`?
```
MrGVSV, line 297: Yeah I think it should definitely be a subtrait. Because then easily have access to all `PartialReflect` methods. Plus, it forces manual implementors to implement both, which might be necessary for certain reflect-based functionality (e.g. third-party crate assumptions).

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,318 @@
...
+- Is the ability to convert from a proxy `PartialReflect` to `Reflect` a
+  hard-requirement, ie: blocker? (I think only a first attempt at implementation
+  can answer this)
```
MrGVSV, line 300: I personally don't think this should be a blocker. `FromPartialReflect` will work "good enough", just as it has up until now. I do think the next big step after this RFC will be sorting that all out, though.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,318 @@
...
+
+## Future possibilities
+
+- Could we modify the `ReflectRef` and `ReflectMut` enums to enable a sort of
+  "partial" evaluation of structures by adding a `&dyn Reflect` variant?
```
MrGVSV, line 305: Like this?
```rust
enum ReflectRef<'a> {
  // ...
  FullReflect(&'a dyn Reflect)
}
```

It sounds like this might only work for value types. Structs and other types will need to still use their proper variant then be converted using `into_full` if needed.

Or is that the idea? If so, we could maybe have `ReflectRef::Value` hold an enum itself between partial and full reflect objects. Or maybe that's too much nesting haha

MrGVSV (commented):
Really liking the direction this is heading :D

rfcs/56-better-reflect.md
nicopap: You are right. It doesn't really make sense for `ReflectRef` and `ReflectMut`. I still had the `ReflectProxy` enum in mind. In which a `FullReflect` would make much more sense.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,318 @@
...
+- `Reflect: PartialReflect`
+- All methods on `Reflect` that **does not** depend on the uniqueness of the
+  underlying type are moved to `PartialReflect`
+- Most things should now accept a `dyn PartialReflect` rather than
+  `dyn Reflect`.
+- It is possible to convert a `PartialReflect` into a `Reflect` using a
+  `into_full` method.
+- `FromReflect` becomes `FromPartialReflect`.
+
+## Jargon
```
PROMETHIA-27, line 15: Could "pass_for" be re-added to the jargon? It was useful terminology, imo, for referencing the represented type.

afonsolage:
Nice! Just wanted to note that another advantage of using `PartialReflect` is that it will be possible to optimize the serde format by not serializing default values.

This will be very useful for scenes and networking

rfcs/56-better-reflect.md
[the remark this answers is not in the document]
MrGVSV: As pointed out by @jakobhellermann, I am going to take back my stance on this. Doing `Subtrait: PartialReflect` makes more sense overall, as it allows proxies to be used properly and also puts the burden of uniqueness wholly on `Reflect`.

That said, I still think we should implement the subtraits for the Dynamic types. If we didn't, then we couldn't really make use of them as proxies.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,324 @@
...
+sense in the context of (2) and (3).
+
+This makes `Reflect` "canonical", in that the underlying type is the one being
+described all the time. And now the `Any` bound and any-related methods are all
+"safe" to use, because the assumptions requires to be able to use them are
+backed into the type system.
+
+We still probably want a way to convert from a `PartialReflect` to a `Reflect`,
+so we add a `into_full` method to `PartialReflect` to convert them into
+`Reflect` in cases where the underlying type is canonical.
```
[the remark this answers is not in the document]
MrGVSV: Yeah I think that's the idea. To convert a proxy, you'll still need to use `FromPartialReflect` like you do now with `FromReflect`.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,324 @@
...
+methods, because changing the trait reflection design is complex and we want to
+keep the scope of this RFC minimal.
+
+```rust
+pub trait PartialReflect: Send + Sync {
+    // new
+    fn as_full(&self) -> Option<&dyn Reflect>;
+    fn into_full(self: Box<Self>) -> Option<Box<dyn Reflect>>;
+
+    fn type_name(&self) -> &str;
```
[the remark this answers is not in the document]
MrGVSV: Mentioned this on Discord, but essentially `type_name` still has a use for deserialization.

However, for Point 2 I think you could intentionally use a non-real type name and be fine. As long as you aren't hoping to deserialize it (unless you `#[reflect(Deserialize)]`), I think it should be okay.

rfcs/56-better-reflect.md
MrGVSV: Yes, it would be solved by it (see #59 and its PoC implementation https://github.com/bevyengine/bevy/pull/6056)

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,324 @@
...
+
+    fn as_partial_reflect(&self) -> &dyn PartialReflect;
+    fn as_partial_reflect_mut(&mut self) -> &mut dyn PartialReflect;
+    fn reflect_ref(&self) -> ReflectRef;
+    fn reflect_mut(&mut self) -> ReflectMut;
```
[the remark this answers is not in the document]
nicopap: Regarding this, https://github.com/bevyengine/bevy/pull/7207 does store `&dyn PartialReflect` in proxies and leaves `PartialReflect: Any`. I think it will require a different vision on `ReflectRef` and proxies to complete the RFC and have a proper separation between proxy and canonical types.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,324 @@
...
+trait: `PartialReflect`. `PartialReflect` also has a `as_full` and `into_full`
+to convert them into "full" `Reflect`. We also keep the "trait reflection"
+methods, because changing the trait reflection design is complex and we want to
+keep the scope of this RFC minimal.
+
+```rust
+pub trait PartialReflect: Send + Sync {
+    // new
+    fn as_full(&self) -> Option<&dyn Reflect>;
+    fn into_full(self: Box<Self>) -> Option<Box<dyn Reflect>>;
```
bjorn3, line 202: I think this should return `Result<Box<dyn Reflect>, Box<dyn PartialReflect>>` to allow getting back the original `PartialReflect` in case `Reflect` is not implemented.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,324 @@
...
+pub trait Reflect: PartialReflect + Any {
+    fn as_reflect(&self) -> &dyn Reflect;
+    fn as_reflect_mut(&mut self) -> &mut dyn Reflect;
+
+    fn any_mut(&mut self) -> &mut dyn Any {
+      // implementation
+    }
+    // etc.
+    fn any
+    fn any_mut
```
bjorn3, line 240: Duplicated above.

rfcs/56-better-reflect.md
```diff
@@ -0,0 +1,324 @@
...
+### New `Reflect` trait
+
+`Reflect` now, instead of implementing those methods, has `PartialReflect` as
+trait bound:
+
+```rust
+pub trait Reflect: PartialReflect + Any {
+    fn as_reflect(&self) -> &dyn Reflect;
+    fn as_reflect_mut(&mut self) -> &mut dyn Reflect;
+
```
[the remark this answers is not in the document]
nicopap: As trait upcasting is not merged, it means I could give those methods the same name as the ones on `PartialReflect`?

Such as:

- `fn as_partial_reflect(&self) -> &dyn PartialReflect`
- `fn as_partial_reflect_mut(&mut self) -> &mut dyn PartialReflect`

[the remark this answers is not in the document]
bjorn3: Hang on, it might actually be possible to call those methods of `PartialReflect` on an `Box<dyn Reflect>` already.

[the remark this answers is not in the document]
soqb: it is possible. bevyengine/bevy#7207 uses this.

MrGVSV:
This can be merged since it's since been implemented in https://github.com/bevyengine/bevy/pull/7207

PR #57, opened 2022-06-12
Deferred mutation queries

MiniaczQ:
[RENDERED](https://github.com/MiniaczQ/rfcs/blob/deferred-mutation-queries/rfcs/57-deferred-mutation-queries.md)
Queries that defer mutation of certain components.


infmagic2047:
"Defer mutation for better parallelism" is basically the purpose of `Command`. Can we unify the implementation and/or API with that?

Nilirad:
I'm worried that the `DeferMut` can be easily confused with `DerefMut`.

MiniaczQ:
> I'm worried that the `DeferMut` can be easily confused with `DerefMut`.

Yeah, we can bikeshed a better name if we decide to keep the mechanic.

> "Defer mutation for better parallelism" is basically the purpose of `Command`. Can we unify the implementation and/or API with that?

Didn't consider `Command`, but yes, it seems to do exactly what is needed, assuming the mutations are batched together.
In fact, I think this piece of code would do exactly that:
```rs
fn foo(
  mut commands: Commands,
  bars: Query<(Entity, Cow<Bar>)>,
) {
  for (entity, bar) in vars.iter() {
    bar.0 = 0.0; // Copy happens
    commands.entity(entity).insert(bar.to_owned()); // We (re)insert the component
  }
}
```
2 problems with that would be:
- bit boilerplate-y, you need to add commands as argument
- order of mutation, with the original idea we'd keep the order the same as system order,
  but if it's commands and few of those get batched, what would the order of application be?

MiniaczQ:
Although bit non-obvious, using Commands seems like the way to go here.
I'm closing the request

PR #59, opened 2022-07-02
`FromReflect` Ergonomics

MrGVSV:
[RENDERED](https://github.com/MrGVSV/rfcs/blob/main/rfcs/59-from_reflect_ergonomics.md)

Derive `FromReflect` by default when deriving `Reflect`, add `ReflectFromReflect` type data, and make `ReflectDeserializer` easier to use.


rfcs/59-from_reflect_ergonomics.md
```diff
@@ -0,0 +1,284 @@
...
+
+If any type in the serialized data did not register `ReflectFromReflect`, this should result in an error being returned indicating that `ReflectFromReflect` was not registered for the given type. With the changes listed in **[Derive Complexity](#derive-complexity)**, this should be less likely to happen, since all types should register it by default. If it does happen, the user can choose to either register said type data, or forgo the automatic conversion with **dynamic deserialization**.
+
+#### Dynamic Deserialization
+
+There may be some cases where a user has data that doesn't conform to `FromReflect` but is still serializable. For example, we might not want to make `Entity` constructible, but we could still pass along data that *represents* an `Entity` so we it can be constructed in some Bevy-controlled way.
+
+These types require manual reconstruction of the real type using the deserialized data. This is what the previous behavior of `ReflectDeserializer::deserialize`. We can re-enable this behavior on the deserializer be disabling automatic conversion:
```
PROMETHIA-27, line 183: NIT: This could be edited, took me a couple tries to understand.

rfcs/59-from_reflect_ergonomics.md
```diff
@@ -0,0 +1,284 @@
...
+
+## Rationale and alternatives
+
+The biggest alternative is doing nothing. Again, most of these are ergonomics and ecosystem wins. The engine will work pretty much the same with or without these changes. The hope is that we can make reflection less intimidating, confusing, and frustrating, while making it simpler, more intuitive, and more consistent across crates.
+
+One concern about integrating the `FromReflect` derive into the `Reflect` derive is that we might want to support something like a `FromWorldReflect`. However, using the opt-out method this should still be possible.
+
+## Unresolved questions
+
+- Should `ReflectFromReflect` be replaced by a method on `TypeRegistration`, such as `TypeRegistration::construct`?
```
PROMETHIA-27, line 278: I don't think so, but I wouldn't be opposed to adding `construct()` as a helper function which wraps over `ReflectFromReflect`.

PROMETHIA-27 (approved):
Love this! I am fully behind and have wanted every feature here for a while.

rfcs/59-from_reflect_ergonomics.md
```diff
@@ -0,0 +1,284 @@
+# Feature Name: `from_reflect_ergonomics`
+
+## Summary
+
+Derive `FromReflect` by default when deriving `Reflect`, add `ReflectFromReflect` type data, and make `ReflectDeserializer` easier to use.
+
+## Terminology
```
alice-i-cecile, line 7: I feel like this might be clearer as a single section, comparing and constrasting at each step. Perhaps a table even?

rfcs/59-from_reflect_ergonomics.md
```diff
@@ -0,0 +1,284 @@
...
+
+#### *Real Types*
+
+The term "**real type**" will be used to distinguish Dynamics and other concrete types. We can define it as any concrete type known at compile time that is *not* one of the Dynamic structs. 
+
+In example code, variables containing real types will be denoted as `real_Y`, where `Y` corresponds to their data structure. For example, `real_list` would represent a list-like type, such as `Vec<T>`.
+
+## Background
+
+### What is `FromReflect`?
```
alice-i-cecile, line 25: I think we might want to start even earlier, with "what is `Reflect`". Remember that the audience is the average Bevy developer, who hasn't touched reflection at all.

If we had good introductory module docs (or those on `Reflect`), I'd link those, but we don't.

rfcs/59-from_reflect_ergonomics.md
```diff
@@ -0,0 +1,284 @@
...
+
+For those unfamiliar, here's a quick refresher on `FromReflect`.
+
+The `FromReflect` trait is meant to allow types to be constructed dynamically (well, not completely, but we'll get into that later). Specifically, it enables a real type to be generated from a Dynamic one. As an example, we can convert a `DynamicTuple` to a real tuple like so:
```
alice-i-cecile, line 29: Users won't understand what a dynamic type is at this point, or why we might want to do this.

rfcs/59-from_reflect_ergonomics.md
```diff
@@ -0,0 +1,284 @@
...
+
+This is why, when deriving `FromReflect`, all fields *must* also implement `FromReflect`[^1]. Without this, the trait itself won't just "not work"— it won't even compile.
+
+### Why is `FromReflect` useful?
+
+So why does this matter? If we need the real type we can just avoid using Dynamics, right?
+
+Well for any users who have tried working with serialized reflection data, they've likely come across the biggest use case for `FromReflect`: deserialization.
+
+When reflected data is deserialized, all of it is placed in a Dynamic type[^2]. This means that, even if we know the real type ahead of time, we don't quite have it yet. This is where `FromReflect` shines, as it allows us to convert that Dynamic into our desired real type.
```
alice-i-cecile, line 65: It's not clear what "don't quite have it yet" means precisely.

alice-i-cecile (commented):
Please try to split your lines at the sentence / clause level; it makes reviewing and suggestions much more pleasant :)

Nilirad (commented):
Have you considered using the “static type” instead of “real type”? It seems to be a direct translation from dynamic vs static typing.

rfcs/59-from_reflect_ergonomics.md
MrGVSV: Yeah I agree that we should keep RFR. I decided to leave out the `Future Possibilities` section, but in it I originally described adding a `ReflectFromWorldReflect` type data to support a proposed `FromWorldReflect` trait. And my thought was that this would pair nicely with RFR without needing to create another method on `TypeRegistration`. (Left out that section because it seemed a bit off-topic, though, haha.)

And also, by keeping RFR we keep open the possibility of cloning and storing the type data outside the registry if needed.

MrGVSV:
> Have you considered using the “static type” instead of “real type”? It seems to be a direct translation from dynamic vs static typing.

I did, but chose not to use it. My reasoning was that "static" feels like we should know the exact type at compile time. However, a `dyn Reflect` could still be a "real type", even though it isn't statically known.

rfcs/59-from_reflect_ergonomics.md
```diff
@@ -0,0 +1,284 @@
...
+dyn_tuple.insert(321usize);
+
+let real_tuple = <(usize, usize)>::from_reflect(&dyn_tuple).unwrap();
+```
+
+A Dynamic value to a real one. Neat!
+
+### How does `FromReflect` work?
+
+Basically it's `FromReflect` all the way down.
```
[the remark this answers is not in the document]
MrGVSV: Hm, what does "subtype" refer to here? Should we use "field/entry" instead?

rfcs/59-from_reflect_ergonomics.md
```diff
@@ -0,0 +1,284 @@
...
+struct Foo {
+  value: (f32, f32)
+}
+```
+
+Calling `Foo::from_reflect` will kick off a chain of `FromReflect`:
+
+`Foo::from_reflect` calls `<(f32, f32)>::from_reflect` which itself calls `f32::from_reflect`.
+
+This is why, when deriving `FromReflect`, all fields *must* also implement `FromReflect`[^1]. Without this, the trait itself won't just "not work"— it won't even compile.
```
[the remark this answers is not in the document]
MrGVSV: To be fair, "this invariant is validated at compile time" is not much simpler haha.

I can reduce the wordiness, but I'm not sure about using language like "invariant" as I think it could be confusing to newer/intermediate devs (including myself who still struggles with understanding variance lol 😅). 

Thoughts on this?

MrGVSV:
> Please try to split your lines at the sentence / clause level; it makes reviewing and suggestions much more pleasant :)

Okay I split bascially every sentence across multiple lines (where I could). Typora hates me now lol, but hopefully that's more readable/reviewable 🙂

rfcs/59-from_reflect_ergonomics.md
MrGVSV: I tried the table, but it wasn't helping that much. Hopefully this is closer to what you wanted.

rfcs/59-from_reflect_ergonomics.md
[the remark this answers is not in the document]
PROMETHIA-27: I think field would be clearer here, `subtype` has a pretty defined meaning in both OOP and more functional type systems and I don't think it's relevant here.

rfcs/59-from_reflect_ergonomics.md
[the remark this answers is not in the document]
PROMETHIA-27: I'm in favor of just cutting it down to "Without this, the trait implementation won't compile". 

Irrelevant to this PR, but I'm considering if it would be a good idea to rename `ArchetypeInvariants` to `ArchetypeRequirements` for the same reason. I feel like a newbie would prefer to see "{Jargon}Requirements" and get at least some idea as to what it means rather than "{Jargon}{Jargonier}" and just ignore it entirely.

raffaeleragni:
Any update on this? I am trying to deserialize some bevy basic types (eg Transform) but can't rewire the type from DynamicStruct, because it doesn't have ReflectFromReflect.

MrGVSV:
> Any update on this?

Unfortunately, there hasn't been much progress. While `@cart` has suggested more of an openness to merging `FromReflect` into the derive for `Reflect`, I haven't taken the time to update the RFC.

One reason I haven't updated it is because after coding the initial implementation in https://github.com/bevyengine/bevy/pull/6056, I just found the RFC to be overkill. The actual changes are only a few hundred lines and most of it is cleanup. On top of that, `ReflectFromReflect`, which was part of this RFC, got merged on its own pretty easily.

So I could update this, but I may just close it out in favor of just using https://github.com/bevyengine/bevy/pull/6056 to respond to any feedback. I'll probably rebase and update that PR relatively soon. Thoughts?

---

> I am trying to deserialize some bevy basic types (eg Transform) but can't rewire the type from DynamicStruct, because it doesn't have ReflectFromReflect.

If you want to make a PR adding in a `ReflectFromReflect` registration, I'm sure that will be easily accepted (basically a four line change to cover both `Transform` and `GlobalTransform`).

In the meantime, however, you can just manually register it for all your known types:

```rust
app.register_type_data::<Transform, ReflectFromReflect>();
```

raffaeleragni:
Didn't know there was a workaround, that's actually great, thanks.

MrGVSV:
Closing this out. See https://github.com/bevyengine/bevy/pull/6056#issuecomment-1575798530 for details.

PR #63, opened 2022-08-08
Bevy nursery: a path to upstreaming ecosystem crates

alice-i-cecile:
[**RENDERED**](https://github.com/alice-i-cecile/rfcs/blob/nursery/rfcs/63-nursery.md)

Work done within the [broader Bevy ecosystem](https://bevyengine.org/assets/) is sometimes a good fit as part of the core engine.
These crates should live in a **nursery** owned by the Bevy org for a period of time,
where they can be polished and refined by the broader community,
before eventually being merged in completely.
When this occurs, nursery crate maintainers will be granted scoped merge rights to the main engine.

### DIFF
--- a/rfcs/63-nursery.md
+++ b/rfcs/63-nursery.md
@@ -1,0 +1,169 @@
+# Feature Name: `bevy-nursery`
+
+## Summary
+
+Work done within the [broader Bevy ecosystem](https://bevyengine.org/assets/) is sometimes a good fit as part of the core engine.
+These crates should live in a **nursery** owned by the Bevy org for a period of time,
+where they can be polished and refined by the broader community,
+before eventually being merged in completely.
+
+## Motivation
+
+Third-party crates, commonly plugins, are often used to extend the engine in ways that are broadly applicable.
+Polished crates that are a good fit for the engine should eventually be merged in wholesale,
+rather than reimplemented, to avoid the [second system effect](https://en.wikipedia.org/wiki/Second-system_effect)
+and minimize breakage for their users.
+
+However, merging in these crates directly will result in very large, unreviewable PRs,
+and may compromise Bevy standards for quality.
+But if we delay, critical ecosystem crates may go stale, pointlessly splitting efforts, or making it hard for beginners to discover the de-facto standards.
+
+## User-facing explanation
+
+In addition to the core crates included within `bevy` itself, there are a number of **nursery** crates,
+which are being prepared for inclusion in the engine.
+
+These crates follow exactly the same technical rules as any other third-party crate, but:
+
+1. Live in the [`bevyengine` organization](https://github.com/bevyengine) on GitHub.
+2. Follow Bevy's standards for code quality and project management.
+3. Explicitly welcome community contributions to help get them ready for release.
+4. Will have a branch prepared for the newest release of Bevy before we release a new major version.
+
+You can find these crates by looking for the special `Nursery` tag on [Bevy Assets](https://bevyengine.org/assets/),
+or by browsing the [`bevyengine` organization](https://github.com/bevyengine)'s list of repositories.
+
+### Submitting a crate to the nursery
+
+If you would like to submit a crate for inclusion in the nursery, please open an issue in the `bevy-nursery` repository.
+While you do not need to be the crate's maintainer to do so, the following criteria must be met before it will be accepted to the nursery:
+
+1. The existing code must be MIT + Apache 2 licensed.
+2. The crate maintainers must consent.
+   1. While this is not strictly necessary due to licenses, buy-in and consent is important to us.
+3. The Bevy maintainers must collectively agree to accept the crate as a nursery crate.
+
+A crate is a good fit for the nursery if:
+
+1. The crate covers functionality that we eventually want to include in the engine itself. Factors include:
+   1. Broadly useful to engine users.
+   2. A direct replacement or improvement for an existing engine feature.
+2. The crate is reasonably mature.
+3. The crate is well-loved.
+4. The crate follows a high standard for testing, code quality and documentation.
+
+Very small or uncontroversial crates may be suitable for a direct PR via the ordinary mechanism.
+
+### Setting up a nursery crate
+
+Once accepted, the repository will be transferred to the Bevy organization under the same name and `crates.io` identity.
+
+The existing crate maintainers will have complete admin permissions over the new crate,
+and any other Bevy maintainers will be granted write permissions (or the `bors` equivalent).
+
+A new milestone will be created, called `upstreaming`, which will track the work that needs to be completed before the crate can be submitted as a PR to `bevy` itself.
+
+Finally, the issues and PRs from that crate will be added to the `#bevy-nursery` channel on Discord to improve discoverability.
+
+### Preparing a crate to be upstreamed
+
+Before a nursery crate can be accepted to Bevy itself, the following conditions must be met:
+
+1. All items in the `upstreaming` milestone must be complete.
+2. A final release must be published, encompassing all PRs merged to date.
+3. The Bevy maintainers must collectively agree to accept the crate upstream in its current form.
+
+This checklist may be helpful when considering issues for the `upstreaming` milestone:
+
+- [] are the examples and documentation helpful and complete?
+- [] do all features work as intended?
+- [] is the code quality acceptable?
+- [] is there any urgent highly-experimental work that should be completed before attempting to integrate more closely with the rest of the engine?
+
+Crates **do not have to be complete** to be merged into Bevy itself: merely polished and clearly useful.
+
+The PR to merge the crate into Bevy itself should be very simple, and contain only the bare minimum change set required to enable interoperability.
+Further integration work should be done in follow-up PRs.
+If you are an interested community member, raise issues and submit fixes on the nursery repository **before** the crate is being upstreamed.
+
+Once the upstreaming PR is merged, the nursery crate's repo will be archived and any still-relevant issues will be migrated to the main Bevy repo.
+
+### Maintaining upstreamed crates
+
+When a nursery crate is upstreamed into the main engine,
+the core maintainer team of the nursery crate will be offered (but do not need to accept)
+**scoped merge rights** for the Bevy project.
+
+Nursery crate maintainers have shown themselves to have the judgement and reliability needed to manage part of the core engine:
+upstreaming a crate should not impose any additional barriers to the crate's development and maintenance.
+Following the standard rules for community reviews, they will be able to merge PRs that affect their nursery crate and directly related areas of the engine.
+Controversial PRs in their area will require either the sign-off of the project lead *or* the nursery crate maintainer, although like always, consulting experts is always wise!
+
+Like always, the project lead may choose to hand out (or revoke) additional permissions on a case-by-case basis where warranted, but the expansion of merge rights outlined above is the default process when a nursery crate is upstreamed.
+
+## Drawbacks
+
+1. Clearly picks winners within the crate ecosystem.
+   1. These winners typically already exist, but it's hard for beginners to discover that critical information.
+2. Hands out additional merge rights to the engine.
+   1. May increase chaos and dilute product direction.
+   2. Tracking areas of expertise may get onerous.
+   3. Marginally increases security risks (2FA + only write permissions) reduce this risk substantially.
+3. Likely to increase the scope of the core engine.
+   1. Feature flags, template games and plugin sets help ensure users only compile and ship the features they care about.
+   2. Adding maintainers ensures that the workload does not increase significantly.
+
+## Rationale and alternatives
+
+### Why should we upstream crates?
+
+In other modular ecosystems, it's very common to have critical functionality outside of the core project.
+This is unfortunate because:
+
+1. It's very challenging for beginners to discover these essential tools.
+2. These tools cannot readily be used in official documentation.
+3. These tools may suddenly go unmaintained.
+4. Work may be divided across competitors (including work done by the core).
+5. Integration with the core may be poor, or suddenly break.
+
+By upstreaming essential tools, we can avoid these problems, and gracefully scale the core organization.
+
+### Why do we need a "nursery" stage?
+
+Ecosystem crates are huge, complex endeavours.
+We want to ensure that the appropriate quality bar is met, but still signal early once a clear candidate has been identified.
+
+The nursery stage serves as a distributed PR review process for the entire crate, breaking down the necessary changes into manageable units of work.
+The added attention helps ensure that community stakeholders have a chance to weigh in before the crate itself is upstreamed.
+
+Finally, as Bevy stabilizes, nursery crates can be rapidly iterated on without following Bevy's stability guarantees.
+
+### Why shouldn't we rename the crate to `bevy_nursery_*` for improved discoverability?
+
+This results in significant bureaucratic work (transferring issues and PRs), risks user confusion and adds pointless migration pain.
+The discoverability issue can be tackled via the Discord channel and the prominent label on Bevy Assets.
+
+### Why do we need to grant merge rights to nursery crate maintainers?
+
+If we do not:
+
+1. Maintainers of popular ecosystem crates are reluctant to submit their crates to the nursery due to the reduced ability to iterate on and fix their projects.
+2. Maintainers of the core engine may become overloaded with work as the scope of the engine grows.
+
+The maintainers of nursery crates have shown their ability to identify, create and lead a high-value project: they're natural candidates for additional responsibility in the core engine.
+
+### Why not use GitHub's Code Owners feature to help scope merge rights?
+
+While [bors has code owners integration](https://bors.tech/rfcs/0357-bors-support-for-codeowners.html),
+it's not clear that it's a useful tool over simple social contract.
+
+Setting up (and maintaining) these lists is tedious bureaucracy.
+This approach would also encourage a less-integrated architecture for organizational reasons, rather than technical ones.
+
+## Prior art
+
+The [rust-nursery](https://github.com/rust-lang-nursery) exists for similar reasons, although there is a greater emphasis on stability.
+
+Unity has a long history of buying out important ecosystem tools,
+which are then often left to rot.
+By bringing on the existing maintainer team and allowing for direct community contributions we can hopefully avoid this fate.
--- a/rfcs/63-nursery.md
+++ b/rfcs/63-nursery.md
@@ -6,6 +6,7 @@ Work done within the [broader Bevy ecosystem](https://bevyengine.org/assets/) is
 These crates should live in a **nursery** owned by the Bevy org for a period of time,
 where they can be polished and refined by the broader community,
 before eventually being merged in completely.
+When this occurs, nursery crate maintainers will be granted scoped merge rights to the main engine.
 
 ## Motivation
 


alice-i-cecile:
This RFC came about due to a discussion with @cart, @aevyrie and @alercah at RustConf :)

It presents an alternate approach to some of the goals in #52, to enable faster and more experimental iteration on large well-contained projects.

BoxyUwU (approved):
I don't know if "approving" RFCs is a thing but I like this :laughing:

rfcs/63-nursery.md
```diff
@@ -0,0 +1,170 @@
...
+
+These crates follow exactly the same technical rules as any other third-party crate, but:
+
+1. Live in the [`bevyengine` organization](https://github.com/bevyengine) on GitHub.
+2. Follow Bevy's standards for code quality and project management.
+3. Explicitly welcome community contributions to help get them ready for release.
+4. Will have a branch prepared for the newest release of Bevy before we release a new major version.
+
+You can find these crates by looking for the special `Nursery` tag on [Bevy Assets](https://bevyengine.org/assets/),
+or by browsing the [`bevyengine` organization](https://github.com/bevyengine)'s list of repositories.
```
CleanCut, line 35: If you like, you could also assign a `nursery` _topic_ to the repositories themselves. A GitHub _topic_ is like a label on a repository, and can be specified as a criteria in a search (if you wanted to, for example, link to a search displaying all the nursery repositories in the bevyengine org). For example, `rust`, `game-engine`, and `2d` are topics on one of my own repositories in the screenshot below. You can add/edit topics by editing the _About_ section of a repository.

image

djeedai (approved):
I'll note that Druid has a nursery repo exactly for that purpose, with pretty much the same rules (except the upstreaming ones which are not written).

This looks like a well thought RFC with direct benefit to all parties involved. All agreed.

rfcs/63-nursery.md
```diff
@@ -0,0 +1,170 @@
...
+   2. A direct replacement or improvement for an existing engine feature.
+2. The crate is reasonably mature.
+3. The crate is well-loved.
+4. The crate follows a high standard for testing, code quality and documentation.
+
+Very small or uncontroversial crates may be suitable for a direct PR via the ordinary mechanism.
+
+### Setting up a nursery crate
+
+Once accepted, the repository will be transferred to the Bevy organization under the same name and `crates.io` identity.
```
aevyrie, line 60: In many cases the name may change, and it may also be a chance to make breaking changes that would otherwise be painful with a direct crate update.

For example, the name `bevy_mod_picking` was chosen to not conflict with the now reserved `bevy_picking`. Alternatively, it may even result in a merge with an existing crate (such as `bevy_ui`) instead of a simple crate addition to the monorepo.

rfcs/63-nursery.md
```diff
@@ -0,0 +1,170 @@
...
+This checklist may be helpful when considering issues for the `upstreaming` milestone:
+
+- [] are the examples and documentation helpful and complete?
+- [] do all features work as intended?
+- [] is the code quality acceptable?
+- [] is there any urgent highly-experimental work that should be completed before attempting to integrate more closely with the rest of the engine?
+
+Crates **do not have to be complete** to be merged into Bevy itself: merely polished and clearly useful.
+
+The PR to merge the crate into Bevy itself should be very simple, and contain only the bare minimum change set required to enable interoperability.
```
aevyrie, line 86: Maybe note that upstreaming is the process of integrating a crate with the `bevy` monorepo under `bevy/crates`?

djeedai:
Ah yes 1 comment: I don't think "pick the winner" is necessarily a drawback so long as those crates are integrated under a feature flag. If there's a "second winner" or even 2 strong crates to begin with, they likely each have strengths and weaknesses and are more suited for some use cases, so I don't think it's bad to have say 2 crates for audio that the user can choose from.

rfcs/63-nursery.md
```diff
@@ -0,0 +1,170 @@
...
+Nursery crate maintainers have shown themselves to have the judgement and reliability needed to manage part of the core engine:
+upstreaming a crate should not impose any additional barriers to the crate's development and maintenance.
+Following the standard rules for community reviews, they will be able to merge PRs that affect their nursery crate and directly related areas of the engine.
+Controversial PRs in their area will require either the sign-off of the project lead *or* the nursery crate maintainer, although like always, consulting experts is always wise!
+
+Like always, the project lead may choose to hand out (or revoke) additional permissions on a case-by-case basis where warranted, but the expansion of merge rights outlined above is the default process when a nursery crate is upstreamed.
+
+## Drawbacks
+
+1. Clearly picks winners within the crate ecosystem.
```
aevyrie, line 107: Could bevy allow competing crates into the nursery? What about allowing two crates to merge into a single, new, nursery crate with multiple owners?

I can see pros/cons in both directions.

rfcs/63-nursery.md
```diff
@@ -0,0 +1,170 @@
...
+### Maintaining upstreamed crates
+
+When a nursery crate is upstreamed into the main engine,
+the core maintainer team of the nursery crate will be offered (but do not need to accept)
+**scoped merge rights** for the Bevy project.
+
+Nursery crate maintainers have shown themselves to have the judgement and reliability needed to manage part of the core engine:
+upstreaming a crate should not impose any additional barriers to the crate's development and maintenance.
+Following the standard rules for community reviews, they will be able to merge PRs that affect their nursery crate and directly related areas of the engine.
+Controversial PRs in their area will require either the sign-off of the project lead *or* the nursery crate maintainer, although like always, consulting experts is always wise!
```
CleanCut, line 101: This language does not make it clear that @cart as [BDFALAICMI](https://bevyengine.org/news/scaling-bevy/#bdfalaicmi) can veto the now-upstreamed-nursery-crate's maintainer, assuming that is true.

aevyrie (commented):
I really like this proposal because it provides a very clear path for upstreaming a crate, as well as explicit guidelines for those interested in doing so.

`bevy_mod_picking` was created based on `@cart`'s prescient request to allow the picking API to be tested and mature in a parallel crate before integrating into the engine. With the nursery in place, it gives me a very clear path to contributing my work back to the engine, which is not currently the case. Instead of putting the work behind a flag, or in a PR that constantly needs to be rebased, this would allow for actual user testing like any other third party crate.

rfcs/63-nursery.md
```diff
@@ -0,0 +1,170 @@
...
+
+While [bors has code owners integration](https://bors.tech/rfcs/0357-bors-support-for-codeowners.html),
+it's not clear that it's a useful tool over simple social contract.
+
+Setting up (and maintaining) these lists is tedious bureaucracy.
+This approach would also encourage a less-integrated architecture for organizational reasons, rather than technical ones.
+
+## Prior art
+
+The [rust-nursery](https://github.com/rust-lang-nursery) exists for similar reasons, although there is a greater emphasis on stability.
```
alice-i-cecile, line 166:
```suggestion
The [rust-nursery](https://github.com/rust-lang-nursery) exists for similar reasons, although there is a greater emphasis on stability.

[Druid](https://github.com/linebender/druid) has a similar [widget nursery](https://github.com/linebender/druid-widget-nursery), and explicitly follows a policy of [optimistic merging](http://hintjens.com/blog:106) within that repo.
```

rfcs/63-nursery.md
```diff
@@ -0,0 +1,170 @@
...
+
+The maintainers of nursery crates have shown their ability to identify, create and lead a high-value project: they're natural candidates for additional responsibility in the core engine.
+
+### Why not use GitHub's Code Owners feature to help scope merge rights?
+
+While [bors has code owners integration](https://bors.tech/rfcs/0357-bors-support-for-codeowners.html),
+it's not clear that it's a useful tool over simple social contract.
+
+Setting up (and maintaining) these lists is tedious bureaucracy.
+This approach would also encourage a less-integrated architecture for organizational reasons, rather than technical ones.
```
CleanCut, line 162: I encourage you to reconsider the [`.github/CODEOWNERS`](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) functionality.

You have noted in this document that:

> 2. Tracking areas of expertise may get onerous.

And here you have decided against using `CODEOWNERS` to define _merge rights_.

However, `CODEOWNERS` is a tool aimed at raising awareness of a change to an area of code that you care about. That sounds like something you _do_ want. The core functionality of `CODEOWNERS` is to automatically add a set of users, teams, and/or orgs that own the code as a reviewer to a PR. The ability to _require approvals_ from codeowners is a feature of branch protection, but you don't have to switch that on.

---

On a separate topic, [Merge Queue](https://github.com/github/roadmap/issues/370) is a new GitHub feature that is in beta and should replace the functionality of `bors` with much better integration with the rest of GitHub. For examples, PR's merged via a merge queue will show as "merged", not "closed", and the UI for Merge Queue will be part of GitHub itself. I encourage you to try it out when it goes GA (or if you're really curious, maybe try out the beta).

CleanCut (commented):
Excellent idea. I have a couple comments.

rfcs/63-nursery.md
CleanCut: Agreed. It would be an excellent chance to evaluate whether it makes sense to change the repo name and/or the crate name (as published to crates.io).

rfcs/63-nursery.md
```diff
@@ -0,0 +1,172 @@
...
+### Maintaining upstreamed crates
+
+When a nursery crate is upstreamed into the main engine,
+the core maintainer team of the nursery crate will be offered (but do not need to accept)
+**scoped merge rights** for the Bevy project.
+
+Nursery crate maintainers have shown themselves to have the judgement and reliability needed to manage part of the core engine:
+upstreaming a crate should not impose any additional barriers to the crate's development and maintenance.
+Following the standard rules for community reviews, they will be able to merge PRs that affect their nursery crate and directly related areas of the engine.
+Controversial PRs in their area will require either the sign-off of the project lead *or* the nursery crate maintainer, although like always, consulting experts is always wise!
```
cart, line 101: I strongly disagree with backdooring our decision making process here, especially as the default course of action. Controversial decisions are often controversial because they have implications in a wider context. While in some cases nursery crate owners _may_ have proven their ability to make consistent / generally likeable decisions, when slotted into a wider context of Bevy the correct choice can change. Someone could make a "perfect" crate that is beloved, consistent, fully featured, well documented, etc and it could _still_ be a terrible choice for Bevy. Nursery crate authors have not proven themselves in that context.

Consider this example (which I expect will be common): pretend we add a crate to the nursery and then as part of the nursery process, we significantly alter the design to fit Bevy's needs. If the "vision" required substantial alternations due to feedback from maintainers, can we really trust the author to make the right calls going forward?

rfcs/63-nursery.md
CleanCut: Getting into that level of detail is probably unnecessary for the scope of this document.

rfcs/63-nursery.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Preparing a crate to be upstreamed
+
+Before a nursery crate can be accepted to Bevy itself, the following conditions must be met:
+
+1. All items in the `upstreaming` milestone must be complete.
+2. A final release must be published, encompassing all PRs merged to date.
+3. The Bevy maintainers must collectively agree to accept the crate upstream in its current form.
+
+This checklist may be helpful when considering issues for the `upstreaming` milestone:
```
cart, line 77: I think this checklist should include things in the vein of:
* Is this the right implementation for Bevy? (Someone could make a "perfect" crate that is beloved, consistent, fully featured, well documented, etc and it could still be a terrible choice for Bevy because they've optimized for different constraints / philosophies)
* If we were to start from scratch, would we have changed anything? Is there a path forward for this crate or would "doing the right thing" require a rewrite?

rfcs/63-nursery.md
aevyrie: Now that you mention it, I'm uncomfortable giving scoped merge rights to nursery owners *at all*. I don't think it should function any different from a PR - it's not as if we give those contributors merge rights to the area of the code they added.

rfcs/63-nursery.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Drawbacks
+
+1. Clearly picks winners within the crate ecosystem.
+   1. These winners typically already exist, but it's hard for beginners to discover that critical information.
+2. Hands out additional merge rights to the engine.
+   1. May increase chaos and dilute product direction.
+   2. Tracking areas of expertise may get onerous.
+   3. Marginally increases security risks (2FA + only write permissions) reduce this risk substantially.
+3. Likely to increase the scope of the core engine.
```
cart, line 113: I'm glad you called it out, but its worth mentioning (at least, via pull request comment) that nursery crates should never expand the "planned" scope of the engine (ex: crates in the nursery would have been manually implemented in-engine eventually). So I don't see this a concerning drawback.

rfcs/63-nursery.md
```diff
@@ -0,0 +1,172 @@
...
+1. All items in the `upstreaming` milestone must be complete.
+2. A final release must be published, encompassing all PRs merged to date.
+3. The Bevy maintainers must collectively agree to accept the crate upstream in its current form.
+
+This checklist may be helpful when considering issues for the `upstreaming` milestone:
+
+- [] are the examples and documentation helpful and complete?
+- [] do all features work as intended?
+- [] is the code quality acceptable?
+- [] is there any urgent highly-experimental work that should be completed before attempting to integrate more closely with the rest of the engine?
```
TimJentzsch, line 82: The rendering of the checklist is broken
```suggestion
- [ ] are the examples and documentation helpful and complete?
- [ ] do all features work as intended?
- [ ] is the code quality acceptable?
- [ ] is there any urgent highly-experimental work that should be completed before attempting to integrate more closely with the rest of the engine?
```

rfcs/63-nursery.md
TimJentzsch: It could become a problem though if a crate gets merged upstream and then stagnates in development due to stricter and slower processes.
Then someone might create a new crate to solve this and we get ecosystem duplication again.

Not sure if this _will_ happen, but definitely something to consider IMO.

rfcs/63-nursery.md
cart, line 82: Probably worth adding "does this crate need to be renamed prior to upstreaming?". Ex: `bevy_mod_picking` -> `bevy_picking`

rfcs/63-nursery.md
```diff
@@ -0,0 +1,172 @@
...
+Like always, the project lead may choose to hand out (or revoke) additional permissions on a case-by-case basis where warranted, but the expansion of merge rights outlined above is the default process when a nursery crate is upstreamed.
+
+## Drawbacks
+
+1. Clearly picks winners within the crate ecosystem.
+   1. These winners typically already exist, but it's hard for beginners to discover that critical information.
+2. Hands out additional merge rights to the engine.
+   1. May increase chaos and dilute product direction.
+   2. Tracking areas of expertise may get onerous.
+   3. Marginally increases security risks (2FA + only write permissions) reduce this risk substantially.
```
TimJentzsch, line 112: This parenthesis seems misplaced to me: 
```suggestion
   3. Marginally increases security risks (2FA + only write permissions reduce this risk substantially).
```

rfcs/63-nursery.md
```diff
@@ -0,0 +1,172 @@
...
+2. A final release must be published, encompassing all PRs merged to date.
+3. The Bevy maintainers must collectively agree to accept the crate upstream in its current form.
+
+This checklist may be helpful when considering issues for the `upstreaming` milestone:
+
+- [] are the examples and documentation helpful and complete?
+- [] do all features work as intended?
+- [] is the code quality acceptable?
+- [] is there any urgent highly-experimental work that should be completed before attempting to integrate more closely with the rest of the engine?
+
```
cart, line 83: We should probably have our "version mismatch" plan laid out here. In some cases nursery crates will have versions that are "compatible" with snapping to the bevy versioning scheme (or a crate rename gives us a clean version history). But sometimes they will not be compatible. What should we do in that case?

cart (commented):
Broadly in favor of this effort, but I do have comment (left above instead of below :smile:)

rfcs/63-nursery.md
cart: I'm very curious about the Merge Queues beta and applied as soon as it was announced. Havent heard back / been accepted :)

rfcs/63-nursery.md
cart: Imo the slower process (and the ability for the 3rd party ecosystem to out compete it in development velocity) is a feature, not a bug. If we are making controversial decisions at the same rate as someone unencumbered by Bevy quality / consistency standards, that would be concerning.

rfcs/63-nursery.md
alice-i-cecile: Neat, but I think this is probably redundant.

alice-i-cecile:
Feedback addressed :) The main point worth addressing explicitly at this point is "nursery crate owners get scoped merge rights". No override for controversial PRs any more; if that process needs accelerating we should address that separately.

We could grant:

1. No merge rights.
2. Merge rights to directly relevant code.
3. Merge rights to adjacent areas of the engine (e.g. all of audio).
4. Full merge rights for non-controversial PRs.

My personal feeling is that 2/3, possibly graduating to 4 is the right way to go :) In my experience, experimenting with authority over small areas has been a great model for both sides of the equation.

aevyrie:
> The main point worth addressing explicitly at this point is "nursery crate owners get scoped merge rights". 

Why should contributors going through the nursery process get merge rights, when PR contributors do not? I think we need a good answer to that question before deciding on what, if any, merge rights to grant.

Does this imply that large enough PRs should be rejected, and told to go through the nursery process? What are the criteria for this threshold?

cart:
> Why should contributors going through the nursery process get merge rights, when PR contributors do not? I think we need a good answer to that question before deciding on what, if any, merge rights to grant.

I agree with this. The fact that someone wrote code does not qualify them to maintain it within the context of Bevy. That is a separate concern and should have its own process / criteria.

Large PR vs Nursery are two paths that accomplish the same goal. They should be treated the same in this regard.

I'm open to considering opening the doors to scoped merge rights (maybe via codeowners). But my vote is to remove things of this nature from this RFC and then have that conversation on its own time.

alice-i-cecile:
After consideration, I agree with you @cart and @aevyrie. I've updated the RFC to match that position now :)

rfcs/63-nursery.md
```diff
@@ -0,0 +1,168 @@
...
+In addition to the core crates included within `bevy` itself, there are a number of **nursery** crates,
+which are being prepared for inclusion in the engine.
+
+These crates follow exactly the same technical rules as any other third-party crate, but:
+
+1. Live in the [`bevyengine` organization](https://github.com/bevyengine) on GitHub.
+2. Follow Bevy's standards for code quality and project management.
+3. Explicitly welcome community contributions to help get them ready for release.
+4. Will have a branch prepared for the newest release of Bevy before we release a new major version.
```
afonsolage, line 31: I have a few questions about this:

1- If some `bevy` repo PR introduce a breaking change into some `nursery` crate, there will besome defined process to detect and warn `nursery` crate maintainers? Like a custom CI on every `nursery` crate which will detect it?

2- What happens if at the end of the 3-month cycle, `bevy` is ready for a new release, but some `nursery` crates doesn't, either because of some complex breaking change or because maintainers didn't have time to update?

3- When releasing patch notes/migration guides/new, updates on `nursery` crates will be included?

afonsolage (commented):
It's a great idea, I love it, but I have some questions, 3 related to nursery crates versioning, and another 4º:

4 - When a `nursery` crate will be removed/discontinued from `bevy-nursery`? In case of removing a `nursery` crate, it will be archived or "returned" to original author?

Nilirad (commented):
I don't have any corrections to suggest, however I have some questions that I think are not answered in the RFC.

- What about more defined rules on branch structure? For example:
  - Must the default branch be named `main`?
  - Should one branch constantly track Bevy's `main` branch?
  - If yes, what it should be? The default branch or a dedicated one?
- What happens to crates that are initially adopted by the nursery but at one point they become no longer the best solution for upstreaming (e.g. new better competitor, neglect, changes in Bevy vision)?

rfcs/63-nursery.md
```diff
@@ -0,0 +1,168 @@
...
+
+In other modular ecosystems, it's very common to have critical functionality outside of the core project.
+This is unfortunate because:
+
+1. It's very challenging for beginners to discover these essential tools.
+2. These tools cannot readily be used in official documentation.
+3. These tools may suddenly go unmaintained.
+4. Work may be divided across competitors (including work done by the core).
+5. Integration with the core may be poor, or suddenly break.
+
```
wilk10, line 142: I would like to add that when depending on external plugins, especially more than one, it is often very painful to track Bevy main (even pointing to a specific commit), and a benefit of upstreaming plugins is that CI is guaranteed to pass.

wilk10 (approved):
I am very much in favour of this. Thanks a lot for moving this forward.

I am often very wary of adding new external plugins because of the mentioned drawbacks. 

Having Bevy's official stamp of approval and guarantee that they are going to be maintained empowers users to try more functionality out.

rfcs/63-nursery.md
alice-i-cecile: I would not expect nursery crates to track main tightly (they'll live in seperate crates), so I don't think this is actually an advantage.

This is the right call IMO to avoid increasing burden and friction too heavily.

aevyrie: At the very least, a nursery might give the community a set of crates to more vigilantly keep up to date with main. However, as Alice said, nursery crates won't be in lockstep with bevy commits.

wilk10: Yeh i understand, I just meant once they're fully upstreamed, not during the nursery phase. 

Nevermind, it's not really important.

PR #64, opened 2022-10-19
Add schematics RFC

MDeiml:
Rendered

Unresolved questions:
* How to synchronize the other direction
* How to be able to have "requirements" in schematics. I propose something like https://github.com/bevyengine/bevy/issues/1481 which could also be used outside this RFC
* How to serialize the schematic world
* How to check if:
    * Every component gets converted
    * Conversions don't conflict

See also https://github.com/bevyengine/bevy/issues/3877


SamPruden:
It's great to see this moving! I'm glad to see my initial proposal is well received. This looks like a great start!

Thoughts in no particular order:

* This fairly tightly couples a schematic entity to having a single corresponding runtime entity. I think that's probably reasonable, but I just wanted to call it out explicitly. The lowest level version of this feature should probably support scenarios where that correspondence doesn't exist, even if they're very rare.
* Having the iterator provide the `commands` is much neater than the convoluted idea I was developing.
* The resource situation is a little tricky. Any resources touched by the system would need to become dependencies of the conversions, and trigger reconversion if they change.
* The `spawn_or_select_child` idea with labels is nice.
* This doesn't tackle the problem of multiple schematics coordinating in non trivial ways.
* Do we want schematic systems to be able to mutate the schematic world? This feels very dangerous. A bug in a system could corrupt the editor save file in ways that may not be immediately noticed. It should be possible to do this from code somehow, but it should be very hard to do this by accident.
* This is covered by "stability against changes", but I would emphasise that one of the goals is to allow significant technical changes to be made without concern for breaking compatibility with already authored content.
* I was originally including the customisation of schematic component UI as part of this proposal, but you're absolutely right to separate it out from this RFC. If schematics are just components, that can be a completely orthogonal feature.
* This design directly mutates the runtime world. My instincts say that this is a bad idea. When it comes to the parts which aren't tackled yet (conflict resolution and coordination, which I think are actually the hardest parts apart from readback) I think it's going to be necessary to maintain an intermediate _constraint specification_ structure representing the runtime world. I'm concerned that if we try to add those later, we'll end up having to scrap the work we've already done. I can't see a way around a three pass (collect, resolve, apply) approach to schematic conversion, although I'd be very happy to be proved wrong.

@MDeiml I'd be very interested to hear your thoughts on immediate conversion vs collect-resolve-apply.

One thing I would emphasise is that we should be thinking of this as a feature aimed at a non-technical audience. Obviously implementing a schematic is a technical process, but a designer consuming them should not be assumed to have any technical experience. For this reason, things like invariant enforcement with good error descriptions are essential.

MDeiml:
Thanks for your feedback! This was really fast. Gonna reorder your points to organize my answer better

> * This fairly tightly couples a schematic entity to having a single corresponding runtime entity. I think that's probably reasonable, but I just wanted to call it out explicitly. The lowest level version of this feature should probably support scenarios where that correspondence doesn't exist, even if they're very rare.

You could just leave the corresponding entity empty. Maybe we could even spawn the entity lazily for the first inserted component. Loosing this correspondence would mean one extra call to `spawn_or_select` if you want a corresponding entity, which is what you want 99% of the time. Having this corresponding entity makes it just way more easy to use IMO.

> * The resource situation is a little tricky. Any resources touched by the system would need to become dependencies of the conversions, and trigger reconversion if they change.

That is tricky indeed. There is no `Changed` or `ChangeTrackers`. Also most of the time the behavior of `Changed` wouldn't even be practical here. I don't think there's a way around giving responsibility to the developer here.

That being said in most cases this should be fine. A typical case where you would `Res` in the first place is getting a handle to an asset. In this case changes to the `AssetServer` wouldn't really matter.

> * Do we want schematic systems to be able to mutate the schematic world? This feels very dangerous. A bug in a system could corrupt the editor save file in ways that may not be immediately noticed. It should be possible to do this from code somehow, but it should be very hard to do this by accident.

Similar to the last point I think. This might make sense in some cases, but we would really need to trust the developer to do it right. Imagine for example a schematic that generates some randomness like a random id when it is converted. It then wants to remember this id somewhere, which can only be in the schematic world.

The other thing to consider here is that schematics as I described them are just systems. Currently it is not possible to restrict systems to be read-only at the schedule level. `SchematicQuery` is intended to be read-only (didn't explicitly say that I think). If you still think this is important I suggest forbidding mutating queries as system params so that systems that want to mutate components need to be exclusive (i.e. take `&mut World`).



> * This is covered by "stability against changes", but I would emphasise that one of the goals is to allow significant technical changes to be made without concern for breaking compatibility with already authored content.

Sounds good. I also think I didn't stress the "stable interface" part enough. Feel free to suggest specific changes to the document btw :)




> One thing I would emphasise is that we should be thinking of this as a feature aimed at a non-technical audience. Obviously implementing a schematic is a technical process, but a designer consuming them should not be assumed to have any technical experience. For this reason, things like invariant enforcement with good error descriptions are essential.

Completely agree. There should also be error detection for e.g. not every schematic component being read, runtime components not being converted back, ...


> * This doesn't tackle the problem of multiple schematics coordinating in non trivial ways.
> * This design directly mutates the runtime world. My instincts say that this is a bad idea. When it comes to the parts which aren't tackled yet (conflict resolution and coordination, which I think are actually the hardest parts apart from readback) I think it's going to be necessary to maintain an intermediate _constraint specification_ structure representing the runtime world. I'm concerned that if we try to add those later, we'll end up having to scrap the work we've already done. I can't see a way around a three pass (collect, resolve, apply) approach to schematic conversion, although I'd be very happy to be proved wrong.
>
> @MDeiml I'd be very interested to hear your thoughts on immediate conversion vs collect-resolve-apply.

I agree on both points. This question is not yet resolved with what we got so far. Gonna make an extra comment to keep things organized ;)

MDeiml:
Now to the question of schematics interacting / resolving conflicts. My intuition here is the following:

(almost) All runtime data needs to be converted to data in the schematic world somehow. Otherwise the scene format that is created using schematics would not really be "complete". You could also think about this from the other direction: Every change you want to make to the runtime world should be achievable through changing something in the schematic world. That's the whole point of having schematics as a editor representation.

Now IMO it would be good to ensure at least to some degree that there is no duplicate data in the schematic world. As an example imagine an entity that has both a "mesh renderer" component and a "particle system" component. Both of these depend on a "visibility" component. My argument is that (in the schematic world) "visibility" should neither be saved in the "particle system" component nor in the "mesh renderer" component, as otherwise we would save this data twice in the schematic world. For me the better design here would be to have a "visibility" component in the schematic world that is seperate from "mesh renderer" and "particle system".

For me what these design choices naturally lead to, is that every piece of data in the runtime world has one and only one corresponding piece of data in the schematic world. Now the term "piece of data" is not very exact, but my suggestion would be to interpret that as "component" as components (in the runtime world) already should be "atmoic" i.e. be so small that breaking them apart into multiple pieces makes them kind of useless. A component that can be separated into multiple pieces IMO are bad design. Summarizing what I say is that "every component in the runtime world has one and only one corresponding component in the schematic world".

Of course this immediately makes the process of conversion quite a bit easier. There is no need for "conflict resolution" any more other than checking that every component in the runtime world is only touched by one schematic (one system). As such there's also no need for "collect-resolve-apply". (Checking what I just mentioned should be possible by just looking at the commands that are produced by each system)

On the other hand it doesn't solve the problem of enforcing envariants, i.e. every entity with a "mesh renderer" component should also have a "visibility" component (in the schematic world). My suggestion here would be to solve this problem completely in the schematic world. We could for example look for the "mesh renderer / visibility" invariant in the schematic world and then resolve it there or give out an error message, whatever is more appropriate. See also https://github.com/bevyengine/bevy/issues/1481.

Sry for the wall of text. Happy to hear your opinion on this.

SamPruden:
> This was really fast. Gonna reorder your points to organize my answer better

A fair subtle roast of my organisational habits.
 
> That is tricky indeed. There is no `Changed` or `ChangeTrackers`. Also most of the time the behavior of `Changed` wouldn't even be practical here. I don't think there's a way around giving responsibility to the developer here.

Is there any fundamental reason why there couldn't be change trackers for resources? On the "mutable access is an assumed change" principle of course. I can imagine "schematic resources" also being used for things like per-level configurations in the editor. In that case, we definitely would need to automatically rerun conversion when they changed. For me, giving responsibility to the developer is a last resort - the API should be as simple to use and as hard to break as possible.

> The other thing to consider here is that schematics as I described them are just systems. Currently it is not possible to restrict systems to be read-only at the schedule level. `SchematicQuery` is intended to be read-only (didn't explicitly say that I think). If you still think this is important I suggest forbidding mutating queries as system params so that systems that want to mutate components need to be exclusive (i.e. take `&mut World`).

Ah, `SchematicQuery` being read-only does help! Requiring something explicit like `.add_mutating_schematic()` seems reasonable.

Your second comment will receive a second reply. Spoilers; we're going to disagree on that one.

alice-i-cecile:
FYI, resources do use change tracking: they're accessible via the `.is_changed()` and `.is_added()` methods.

SamPruden:
One-to-one schematic components to runtime components fundamentally contradicts the design goals for me, for a number of reasons.

Schematic components are intended to be logical units of functionality from the perspective of a designer. Often, that logical unit will be significantly larger than the atomic components used at runtime. My goto example is a Rigidbody Schematic. The designer wants one big block they can add to an object, with neat checkboxes for things like "Is Kinematic". Dynamic and kinematic rigidbodies likely have quite different sets of runtime components, and certainly break into many atomic parts.

One-to-one is also incompatible with the goal of being flexible to changes (both small and large) in the runtime data layout. You may design the schematic as being one-to-one, then want to break the runtime data into multiple components later.

I see two broad categories of cooperation between schematics:

- Multiple schematics each adding the same component. A trivial example might be that multiple types of rendering schematic all want to add a `DoesRenderingTag` component.
- Schematics requiring a component to be added by another schematic, but not adding it themselves. An example might be that the Rigidbody Schematic complains about needing a `Transform` component to be added, and ideally suggests "Try adding a Transform Schematic". (That would require the editor to know which schematics are capable of adding which components, which is an idea that I'm exploring.)

On the principle that we should empower users as much as possible, my inclination is towards a powerful underlying schematic "engine", capable of tracking all of these dependencies and managing constraints. We then build simple APIs for the common case on top of that engine. It's harder but (IMO) it's better.

MDeiml:
Sry, I was not very clear. I proposed a "one-to-many" structure. I'll make a concrete example cause it's much easier to talk about it that way.

```rust
struct MeshRendererSchematic {
    mesh: Handle<Mesh>,
    material: Handle<Material>,
}

struct VisibilitySchematic {
    visible: bool
}

struct TransformSchematic {
    translation: Vec3,
    rotation: Vec3,
    scale: Vec3,
}

fn main() {
    App::new()
    // ...
        .add_schematic(mesh_render_schematic)
        .add_schematic(visibility_schematic)
        .add_schematic(transform_schematic)
        // results in `VisibilitySchematic` automatically being added
        .add_archetype_invariant(MeshRenderSchematic::implies(VisibilitySchematic))
        // results in an error message if entity doesn't have a `TransformSchematic`
        // (`implies` would also fit better here, just for example purposes)
        .add_archetype_invariant(MeshRenderSchematic::requires(TransformSchematic))
    // ...
        .run();
}

fn mesh_render_schematic(query: SchematicQuery<MeshRendererSchematic>) {
    for (mesh_renderer_schematic, commands) in query {
        commands.insert_or_update(mesh_renderer_schematic.mesh);
        commands.insert_or_udpate(mesh_renderer_schematic.material);
    }
}

fn transform_schematic(query: SchematicQuery<TransformSchematic>) {
    for (transform_schematic, commands) in query {
        commands.insert_or_update(Transform {
            translation: transform_schematic.translation,
            rotation: Quaternion::from_euler(
                EulerRot::default(),
                transform_schematic.rotation.x,
                transform_schematic.rotation.y,
                transform_schematic.rotation.z,
            ),
            scale: transform_schematic.scale,
        });
        // This could maybe even be an `ArchetypeInvariant` on the runtime world
        commands.insert_or_update(GlobalTransform::default());
    }
}

fn visibility_schematic(query: SchematicQuery<VisibilitySchematic>) {
    // you get the point
}
```

I guess this even allows a "many-to-many" relationship between components. Just not a "many-to-many" relationship between schematic **systems** and runtime components.

Are we getting closer to agreeing? xD

MDeiml:
> FYI, resources do use change tracking: they're accessible via the `.is_changed()` and `.is_added()` methods.

That's great. Still means that developers need to check this manually. Do you think that's ok? Otherwise we could maybe do some magic and have something like `SchematicParam` which is different from `SystemParam` and automatically has change tracking, but that would maybe limit the user and add a lot of complexity to the implementation. We would also need `IntoSchematic`, `FunctionSchematic`, ... basically duplicate the whole `ecs::system` module.

MDeiml:
> On the principle that we should empower users as much as possible, my inclination is towards a powerful underlying schematic "engine", capable of tracking all of these dependencies and managing constraints. We then build simple APIs for the common case on top of that engine. It's harder but (IMO) it's better.

To summarize, we're disagreeing in two points (probably):
1. "one-to-many" vs "many-to-many" for schematic systems <-> runtime components
2. Resolving conflicts during the conversion process (what you call "schematic engine" I guess) vs resolving conflicts inside the schematic world (what I call "archetype invariants")

SamPruden:
Ah, our positions are much closer than I thought then! The "every component in the runtime world has one and only one corresponding component in the schematic world" line threw me off.

I've certainly been imagining that something like schematic implies/requires relationships would be possible, so we're agreed on that.

However, this brings up the question of whether we're expecting people to only have one canonical schematic to achieve a certain goal, and whether we're comfortable with those schematics being tightly coupled to each other.

If there is one single `TransformSchematic`, then doing a basic `requires` on it makes sense. But, if we want the user to be able to build their own `MyTransformSchematic`, that should also be compatible. This probably doesn't make much sense for transforms, but for more complex things it could happen. Using the archetype invariant method, the `MeshRenderSchematic` is tightly coupled to the `TransformSchematic`, and incompatible with `MyTransformSchematic`. This probably doesn't make much sense for transforms, but for more complex things it could happen.

The other approach is to define compatibility purely on the basis of the interpreted runtime data. This approach is purer, more versatile, and (I think) strictly more powerful. However, it does require a significantly more complex implementation.

One issue with the runtime data compatibility approach is that it's possible for the compatibility between schematics to change accidentally if their implementation changes. Personally, I think I'm happy saying that if programmers mess that up then they've introduced a bug, go fix it.

I'll make a follow up comment with a very rough sketch of how I imagine schematic systems may look, attempting to deal with neatly piping through the necessary information for change tracking, constraint resolution, and nice editor error messages and suggestions.

SamPruden:
> Note that this takes a single schematic component, not a query .

To motivate why I prefer this over `for (a, commands) in query`:

Using the query approach, it's possible to cheat the tracking by storing information from previous iterations in a local. I don't believe that there's any way to 100% stop people from ever breaking the tracking, but forcing them to write to global data to break it feels more robust to me. It's more rusty.

MDeiml:
Also seems like a good design.

* I **really** like your naming using `require`. Seems a lot more concise than my `insert_or_update`.

* I'm not sure I understand the generic argument in `SchematicCommands`. Can you add more parameters to common interpreter functions? Otherwise I don't think this is needed.

* For the query interpreters I quite liked my design of having iteration also return the commands. With the function signature where `SchematicQuery` and `SchematicCommands` are separate code would have to look like

  ```rust
  for (entity, component) in query {
      let commands = schematic_commands.get(component);
      // ...do stuff with `commands`
  }
  ```

I think we should talk about the duplicate API of "schematic functions" and "schematic systems" though. Tbh I think there's good reasons for and against both (feel free to add points, I'll edit this comment to stay up to date):

Pros:
* Less verbose in 99% of all cases (this is a big win)
* Harder to break correspondence

Cons:
* Duplicate API (biggest disadvantage IMO)
* Might lead users to write less efficient systems (not sure if this is a concern) if the schematic needs memory access / calculation that is common to all schematic components of the same type
* Schematics are not just systems, so there needs to be some translation between the two
* Breaking change tracking is possible using `SchematicQuery`, but I don't think it would happen unintentionally. People writing schematics are technical users, so probably it's fine to trust them a bit. I guess what could make the API a lot more clear is to require syntax like `for (component, commands) in query.iter_changed()` with an `for (component, commands) in query.iter_every_single_component_you_might_not_want_to_do_this()` alternative.

Finally regarding resolving conflicts in the schematic world vs as part of conversion: (The complexity of implementation should be roughly similar.)

Pro "schematic world" / archetype invariants:
* Can be used for things other than schematics
* Persistent, i.e. the result of the resolution is saved as part of the schematic world
* Apparent to user. Components don't get added quietly but instead would pop up in the editor. This is similar to Unity's `[RequireComponent(...)]` I think.
* Invariants can be added to schematic world as well as runtime world as opposed to only in the runtime world

Pro "collect-resolve-apply":
* Better integrated syntactically with interpreters / schematics. Uses the `require` syntax inside schematic functions as opposed to `add_invariant` in setup code
* As such probably easier to understand for the coder
* More customizable, as it is defined using code rather than structs

Btw I'm amazed at how quickly this is moving :). Seems like together we'll be able to come up with a pretty nice design here.

SamPruden:
> * I'm not sure I understand the generic argument in `SchematicCommands`. Can you add more parameters to common interpreter functions? Otherwise I don't think this is needed.

You can also do a query style `SchematicCommands<(ComponentA, ComponentB)>`.

The intent is to allow the engine to know which interpreter functions can create which components.

This is useful for editor UX reasons. If we have a schematic with a `require::<TransformComponent>()` constraint (i.e. it requires a transform component to be present, but doesn't add one itself), the editor can help the designer by saying "I know that `TransformSchematic` produces that component that you need! Why don't you try that?"

SamPruden:
> For the query interpreters I quite liked my design of having iteration also return the commands. With the function signature where SchematicQuery and SchematicCommands are separate code would have to look like

Yep, I also thought that was very pretty.

As a matter of personal style, I always like to encode constraints into the type system as heavily as possible. I wil happily go to extra effort to move documented restrictions and runtime errors into compiler errors. Given that personal leaning - which I think Rust shares - I prefer the more locked down approach.

I don't think it's a dealbreaker either way, it's a surface level difference.

SamPruden:
> Schematics are not just systems, so there needs to be some translation between the two

I think maybe this is a good thing. In fact, maybe we want to encapsulate the schematic `World` and only allow access through schematic APIs.

If we want to enforce proper use of dependency/change tracking, we probably want to disallow people from doing untracked queries, which they would be able to do if it were just a plain system. This is my personal tendency towards locking things down (without restricting power) again.

MDeiml:
> As a matter of personal style, I always like to encode constraints into the type system as heavily as possible

Oh, I'm not sure I explained myself good here. My comment was regarding your query interpreters where `SchematicQuery` and `SchematicCommands` are separate. Having `SchematicQuery` **return** the `SchematicCommands` seems to be a bit more safe to me. The generic parameter from `SchematicCommands` would then need to be in `SchematicQuery` I guess. But also this is not a very important point.

Btw thanks for clarifying my question about the generic parameter, I understand it now. It's solving a problem though, that only comes up with the collect-resolve-apply approach, not with archetype invariants.

Otherwise I think we reached a point where we just have different but equally valid opinions, so we should maybe write down both designs with their pros and cons and then get feedback from other people. Do you agree?

We could add two sections for archetype invariants and your collect-resolve-apply approach with a comment explaining that this is a "pick one out of two" situation. Also add a section for your "common interpreters" with a comment explaining the feature is optional.

SamPruden:
> Having `SchematicQuery` **return** the `SchematicCommands` seems to be a bit more safe to me.

Oh I see. Yes, I think this is just a matter of API style, they seem equivalent in all meaningful ways. 

> It's solving a problem though, that only comes up with the collect-resolve-apply approach, not with archetype invariants.

Agreed. 

> Otherwise I think we reached a point where we just have different but equally valid opinions, so we should maybe write down both designs with their pros and cons and then get feedback from other people. Do you agree?

There's a big scary monster that we haven't addressed yet, and that's readback.

If we assume that it's possible to drop in and out of a "play mode" in the editor, we would ideally like changes to runtime state to somehow be reflected back into the schematics. Schematics are the interface that designers are used to using to interact with that data, and we'd like them to be able to stay in that familiar territory for basic observation of runtime state. We also want to be able to do things like edit the `TransformSchematic` live in the editor during gameplay to move an object while testing gameplay.

I don't believe an automated solution is possible in the general case, because we can't (and don't want to) enforce that the schematic -> runtime mapping is bijective. We could trivially create an interpreter that ignores a schematic field, and now there's no way to recreate the schematic from the runtime data.

I expect that this will require some type of opt-in manually implemented "schematic inference" functionality. It won't be possible for everything, but for something like `TransformSchematic` it would be easy to implement.

A big question here is whether we attempt to fold that API into the existing interpreters (so an interpreter also specifies how to invert itself as part of the command) or whether the solution is to have a completely separate set of "inference systems" that go the other way.

We should probably do some thinking about whether this part of the problem impacts the choice between the two designs before we solicit feedback on them.

MDeiml:
> We should probably do some thinking about whether this part of the problem impacts the choice between the two designs before we solicit feedback on them.

Good point. I think I'll still write down both solutions for future reference.

> I don't believe an automated solution is possible in the general case

Agreed

> I expect that this will require some type of opt-in manually implemented "schematic inference" functionality

I also think that's the best tactic here. It would be nice though if the editor could output something like "TransformSchematic is not being read back" in an appropriate place, so there should be a mechanism to detect schematics that don't have this implemented.

> A big question here is whether we attempt to fold that API into the existing interpreters (so an interpreter also specifies how to invert itself as part of the command) or whether the solution is to have a completely separate set of "inference systems" that go the other way.

Good question. Intuitively I'd suggest something like:
* Common interpreters are implemented as a trait on a unit struct
* Inference for those interpreters is another trait (that extends the interpreter trait) that you would implement on the same struct. This way both can be added together
* Query interpreters are a bit more crazy, basically just systems, so probably it's ok if their corresponding inference is not grouped with them

Thinking about this now (as you predicted) makes me more keen to include common interpreters for that exact reason. :sweat_smile:

SamPruden:
One possible solution to this is to have two classes of schematics, Edit and Play. Edit Schematics are the ones already discussed in this thread. Play Schematics are the ones that get inferred from the runtime state during play. Often, one schematic will hold both roles - but there's room for both Edit Only and Play Only schematics.

* Edit Only Schematics are those with interpretation but not inference.

* Play Only Schematics are those with inference but not interpretation. They would be used to display (and potentially edit) runtime data that's not part of the editor/serialized state. They only exist while in play mode.

Inference functions would be like inverse interpreters. They can perform queries over the runtime world, and infer schematics into the edit world.

In the case of something like `TransformSchematic`, the same identical schematic would have both interpretation and inference, meaning there's no UX difference between edit and play.

Every time I've tried to solve this readback problem previously, I've ended up concluding that my solution was fatally flawed in some way. So I wonder what's wrong with this one - any ideas?

SamPruden:
> Good point. I think I'll still write down both solutions for future reference.

I'm happy to do the write up of mine if you want to do yours - it's probably best if each is presented by its strongest advocate.

> Thinking about this now (as you predicted) makes me more keen to include common interpreters for that exact reason. 😅

Amusingly I actually don't think this is a good reason! In the inference model proposed above, I'm quite happy for them to be implemented independently, even if they're commonly side-by-side in code.

An advantage of that inference approach is that it's not tied too closely to the schematic that generated the data. We should assume that runtime data will drift away from what the schematic specified (including components being added and removed) and our play UI should attempt to be as tolerant of that as possible. A little conceptual separation here may not be a bad thing.

MDeiml:
> During gameplay, some entities may be spawned in

Yes definitely. This shouldn't be thought of as "readback" but definitely as "conversion the other direction" for that reason probably.

Still both directions are not separate. If an instance of a schematic component has a corresponding component in the runtime world, then that link should also be accessible from the runtime world.

The inference equivalent to "query interpreters" is actually not to hard to define I think. It would be systems that look like
```rust
// using the components from my example in the rfc
fn complicated_schematic_inference(query: InferenceQuery<(&MainA, &Children), SchematicA>, child_query: Query<&MainAChild>) {
    for ((a, children), inference_commands) in query {
        let a_child: &MainAChild = (
            // not sure on the specific syntax here.
            // Use `child_query` to find `MainAChild` components in all children
        );
        let entity = inference_commands.map_entity(a_child.0);
        inference_commands.require(SchematicA(a.0, entity));
    }
}
```

`InferenceQuery` could access some global resource to find out the entity mapping (or maybe that could be done later not too important). `InferenceCommands` more or less only has the two methods I just used I guess.

MDeiml:
Now if we find a good way to put this into "common interpreter" from I'd be happy to include those. I think in this case it would be fine to bundle both functions into one object, since they should be at least kind of inverse. The "common" interface is just there to translate stuff into the system form anyways.

MDeiml:
I think we should try to keep the tracking, schematics might contain extra information that can't be reconstructed and inference might not have a unique result in the case of sth like `MyTransformSchematic`.

Keeping this tracking is IMO not the hard thing if we have "one-to-many" relations, but that's just a feeling. I imagine a solution where this ownership is tracked per component / entity.

Now thinking about it the design should be a bit more elaborate than my code example just now. Inference should be thought of as two parts:
1. Run over all known instances of schematics and either update them or decide that they were deleted
2. Run special systems to find new instances of schematics. The important interface here is a query that can filter out things that are already in schematics 

Using different representations for runtime and editor is what unity does and people really don't like it. We should try to do better.

And for the most common situation where a schematic is more or less just a bundle this isn't too hard. I don't really see runtime and schematic drifting apart over time for that reason. It's not like any sensible user is going to write some monsterous big schematic that has effects all over the place. I actually imagine that schematics most of the time would be derived, for which we could probably guarantee correctness.

Now finally tracking doesn't have to be perfect. One advantage of the "one-to-many" design is that the schematic world could be "see through" i.e. where schematics are missing the editor could display runtime data.

MDeiml:
I added a section explaining my take on inference a bit more

SamPruden:
If I understand your version of inference correctly, it looks something like:

- When entering play mode, we make a clone of the schematic world. (so that we can restore it when we finish playing)
- During gameplay, maintain both a schematic and a runtime world.
- The editor's scene hierarchy and related UI primarily shows the schematic world.
- When we select an object to view and edit its properties, that's a schematic entity.
- Every frame, the state of the runtime world is inferred back into the schematic world.
- When we know that a runtime entity is the result of schematic conversion, we do our inference process back into that original schematic entity.
- The schematic inference process does some kind of "patching" operation - it's not inferring the whole schematic value, it's just updating bits of it where it has the information available.
- Where an entity isn't the result of conversion, a new schematic entity is spawned and inferrence is done into schematics on that entity. In this case it's not only patching, because the schematic needs to be wholly inferred.
- When schematics are non-flat - meaning they correspond to child entities as well as a main entity - we somehow still do sensible inference on them.
- When a runtime entity has components for which no schematic exists, we just don't show that data? Or we have some type of default inferred schematic to show it?
- All of this works with some type of two-way binding. We can edit the schematic values live to edit the runtime data in arbitrary ways.

Is that a reasonable summary?

If you can see a way to make this work then fair enough! But personally I've never managed to envision a way of doing this that isn't hugely burdensome for schematic authors.

I completely agree that the goal should be to have the UI experience for designers be as similar during editing and playback as possible. However, I'm not sure that it's okay to be hiding the runtime data behind our imperfect schematic abstraction during gameplay. That recieved some (IMO fair) pushback in the original proposal thread.

SamPruden:
> We can edit the schematic values live to edit the runtime data in arbitrary ways.

This is problematic, because it means that if a runtime component is inferred back into a schematic and edited, that runtme -> schematic -> runtime round-trip must maintain 100% fidelity every step of the way, otherwise you're corrupting other bits of runtime data. This may usually be possible, but it feels very brittle. Probably too brittle.

MDeiml:
Very good summary. You said it better than me :) This is just my vision though, I guess most of it is up to whoever writes the editor UI eventually. It's just our job that something sensible is possible.

> * When a runtime entity has components for which no schematic exists, we just don't show that data? Or we have some type of default inferred schematic to show it?

I guess it would be good to show a placeholder for the missing schematic. If the runtime component supports the "displayable in UI" trait it could be displayed as if it were copied maybe. I see some possibilities here. Also maybe you could switch between "schematic" and "runtime" views in the editor for debugging purposes.

> But personally I've never managed to envision a way of doing this that isn't hugely burdensome for schematic authors

I agree that all of this adds A LOT of work for schematic authors. I propose to mediate that by making schematics derivable. See my most recent update for how that could look.

> Yet we still need the round-trip to be perfect

In the most common cases including every schematic that can be derived this is very much achievable. In other cases authors (as mentioned above) would already need to invest a lot of time and probably read into schematics. As such I think it would be possible to warn them about the consequences of imperfect synchronization.

> That recieved some (IMO fair) pushback in the original proposal thread

Also fair in my opinion. We need to provide a solution where it's really easy for users to get 100% perfect round trips in almost all cases. That could be solved with the derives and other tools we could provide though. Maybe we can find a way to make schematics composable in a way, that composing 100% perfect schematics gives 100% perfect schematics.

SamPruden:
> This is just my vision though, I guess most of it is up to whoever writes the editor UI eventually. It's just our job that something sensible is possible.

I think that in order to make a compelling case for schematics, we should probably provide a sensible high level proposal for how the editor would consume them. It doesn't need to be an actual full RFC for the real editor workflow, but it needs to prove that schematics are up to the task by breaking down the types of operations that the editor will want to do and showing that schematics can do them.

> I agree that all of this adds A LOT of work for schematic authors. I propose to mediate that by making schematics derivable. See my most recent update for how that could look.

I too have always envisioned the common case being automated.

However, from what I've seen in the Unity world, people will want to make custom schematics quite often. There are many core components for which custom schematics will be necessary - e.g. `RigidbodySchematic` will likely have a lot of complexity if it wants to provide great UX.

One of the primary goals of schematics is to make refactoring the runtime layout a low friction operation. Non trivial refactorings will generally require the derived implementation to be dropped in favour of a manual one which knows the mapping to the new data layout. If that requires writing burdensome inference logic, we have a problem.

If people **ever** avoid making changes to runtime data because the schematic burden of that work is too great, we've completely failed at our jobs. That's an anti-goal.

Covering 85% of cases is easy, 95% is hard, 100% might not be possible.

SamPruden:
There's a different way of looking at this which is somewhat of a radical departure. I don't know whether it makes any sense.

Schematics actually serve two related purposes - they're the serialization/authoring data format, and the UI/UX experience.

It's assumed that there will be some way to customise the UI of a schematic's editor panel, but we've been leaving that for a separate RFC.

The serialisation part is irrelevant in play mode. Actually, the only thing that we care about being the same during normal editing and during play mode editing, is the UI.

What if we think of this as three separate layers?

Edit mode: Bind UI to schematic, convert schematic to runtime data
Play mode: Bind same UI directly to runtime data

This avoids all of the issues that come from roundtripping the runtime data back through the schematic struct. However, we would need to separately specify the UI layer and write two versions of the UI binding. But hopefully we could share the UI layout, so the UX would remain the same.

It's more code to write, but it's much easier code to write correctly. And I think it can handle more situations without breaking, and would still be derivable in the trivial case.

I don't know - am I crazy?

MDeiml:
> we should probably provide a sensible high level proposal for how the editor would consume them

I agree

> What if we think of this as three separate layers?
> 
> Edit mode: Bind UI to schematic, convert schematic to runtime data Play mode: Bind same UI directly to runtime data

That's what I meant with the debug view in my earlier comment. I think that is not a full solution though. As I mentioned earlier Unity has gotten bad feedback for the same approach that you are describing here. It should be an important piece in the eventual UI design though. It could even be possible to mix data from the schematic and runtime world in the UI if it is done smart enough. And it gives us a safety net for when schematic fail. In the current version I say that the inference is optional. Components without inference should probably displayed in the way you are describing here.

> The serialisation part is irrelevant in play mode

It would be nice to have serialization in runtime as well, as this could be used to save game state, but obviously that's not a very important goal.

> It's assumed that there will be some way to customise the UI of a schematic's editor panel, but we've been leaving that for a separate RFC.

I think there's solution here somewhere though. Generally what is complicated is not the actual schematic component, but how to display that data. An example:

> `RigidbodySchematic` will likely have a lot of complexity

If you look at the [components that rapier uses](https://docs.rs/bevy_rapier2d/latest/bevy_rapier2d/dynamics/index.html) you will see that `RigidbodySchematic` will be more or less
```rust
#[derive(Component, Schematic)]
struct RigidbodySchematic {
    rigidbody: Rigidbody,
    velocity: Velocity,
    mass: MassProperties,
}
```
plus a `ColliderSchematic` that is required by this somehow. The data structure itself is not that complicated.

What is complicated is the UI associated with that data. You probably want to display a bounding box and velocity vectors and center of mass etc. but that does not complicate the conversion between the two worlds in a meaningful way.

> One of the primary goals of schematics is to make refactoring the runtime layout a low friction operation. Non trivial refactorings will generally require the derived implementation to be dropped in favour of a manual one which knows the mapping to the new data layout. If that requires writing burdensome inference logic, we have a problem.

That is a **very** important problem with relying too much on `derive` though. I think the solution to that is threefold:
1. Provide more tools for writing schematics than just `derive`. E.g.
   * make it possible to add together schematics for individual fields of structs and individual variants of enums
   * create a `map(forwards: Fn(...), backwards: Fn(...)` method
   * Have builtin multi purpose schematics like `struct PrefabSchematic(Scene)`
   
   I think these are all implementable
2. Make sure that a schematic does not necissarily need inference. Have a fallback that uses runtime components and their `Reflect` instance. This is similar to what you suggested in your comment
3. Don't be to strict on the "stable to changes" part. It is unrealistic that a design would be resistant to **any** implementation changes. Tell users it's also ok to rewrite their schematic.

MDeiml:
See my newest commit for my suggestion which combinators we could add.

SamPruden:
>> Edit mode: Bind UI to schematic, convert schematic to runtime data Play mode: Bind same UI directly to runtime data

> That's what I meant with the debug view in my earlier comment.

I think we've had a miscommunication here - I'm wondering what it would look like if this were the standard architecture of all schematics. They have five separate pieces:

1. A schematic component - stable serializable struct
2. An interpreter that does schematic component  -> runtime data
3. A UI panel
4. A binding between the UI panel and the schematic component
5. A separate binding between the same UI panel and the runtime data

When in edit mode, we use parts 1, 2, 3, 4.
When in play mode, we use parts 3, 5 - there is no schematic world in play mode.

The UI panel is actually the only piece that's common between the two workflows, but it creates continuity in the designer's UX.

I believe this technically eliminates all of the problems - roundtripping fidelity issues are 100% gone as long as the UI binding system is reasonable. Those issues come from trying to force the flexible runtime data back through the schematic struct under faulty bijectivity assumptions, which we're not doing anymore. 
 
But it comes at the (huge!?) cost of having five whole separate pieces to a schematic. This is probably completely unreasonable, unless we can find ways to heavily automate even non-trivial cases.

> If you look at the [components that rapier uses](https://docs.rs/bevy_rapier2d/latest/bevy_rapier2d/dynamics/index.html)

I'm shocked that's so simple! I expected to see a fair number of conditional components, e.g. different components between kinematic and dynamic bodies. I know that [Unity's ECS Physics](https://docs.unity3d.com/Packages/com.unity.physics@1.0/manual/core_components.html) has a lot going on in terms of combining different components for different behaviours.

> Combinators

I'm afraid I'm not following the purpose of these, sorry! 

> Don't be to strict on the "stable to changes" part. It is unrealistic that a design would be resistant to any implementation changes. Tell users it's also ok to rewrite their schematic.

I'm afraid I'm a zealot on this point. It's not okay to break schematic serialization compatibility in anything but the most extreme scenario.

One of the primary goals of schematics is to allow designers and programmers to work side-by-side simultaneously. An already built scene should never break because of a change to the runtime representation. Changing the schematic logic should be common, but changing the schematic's data layout in ways that break compatibility with assets should almost never happen. If we haven't achieved this goal, the feature isn't that useful.

Of course sometimes changes will need to happen because the underlying functionality has changed, and additional information is needed. There's no avoiding that.

MDeiml:
Ah ok, I understood wrong then.

> 5\. A separate binding between the same UI panel and the runtime data

I'm pretty sure that's the same thing with extra steps. You would have
```
Runtime <-- Schematic
   ^           ^
    \         /
     v       v
     UI Panel
```
And the arrows need to commute. I draw the left and right arrows to be reversible since you can update both worlds from the UI. Now you could go `Runtime -> UI -> Schematic` in that diagram and I'd expect it to be reasonable only if it's the inverse of the `Schematic -> Runtime` arrow. Otherwise couldn't UI values change just by starting the game? I don't think that's nice behavior.

> roundtripping fidelity issues are 100% gone as long as the UI binding system is reasonable

We would just reduce the problem of having perfect `Schematic <-> Runtime` conversions to having perfect `Runtime <-> UI` conversions, which I feel are the same thing since UI and schematic data representations should be almost identical, if not the same thing.

> An already built scene should never break because of a change to the runtime representation

> Of course sometimes changes will need to happen because the underlying functionality has changed, and additional information is needed. There's no avoiding that.

"never" is a bit to strong here, but you already said that yourself. Also the cases were you need additional information is more what I was thinking of with my "3."

> If we haven't achieved this goal, the feature isn't that useful

I disagree. It's still very much useful for driving editor UI which is the primary purpose. It is also still useful to make the scene format more stable. You would just have to put into updating your scenes if your underlying representation changes to a reasonable degree.

> > Combinators
> 
> I'm afraid I'm not following the purpose of these, sorry!

Oh I guess that's my fault. Those are methods that take in one or more schematics and spit out another schematic. Think of something like `Iterator::map` or [`Iterator::zip`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.zip) which takes one iterator and creates another one. There are some schematic combinators that could be implemented such that the schematics that they create are always 100% perfect conversions if the schematics that you put in are 100% perfect (maybe up to some restrictions on other arguments).
My idea was that schematics can be created by combining trivial ones like `DefaultSchematic`.

SamPruden:
> We would just reduce the problem of having perfect Schematic <-> Runtime conversions to having perfect Runtime <-> UI conversions, which I feel are the same thing since UI and schematic data representations should be almost identical, if not the same thing.

So my thinking was that the UI could be more flexible. E.g. if some field's data is missing in the runtime, we can do something like disabling the field in the UI. The UI just has to take the data and display something sensible and consistent-ish, but it can adapt a little to its circumstance in a way that can't happen if the UI is only seeing the schematic, which is a very rigid struct.

It also allows the UI to expose runtime data that doesn't necessarily want a schematic - i.e. stuff that doesn't need a serializable representation.

From a separation of concerns perspective, I actually think it might be more elegant. But 5 things is a lot of things. It's so much boilerplate. Many schematics will want to customise their UI anyway, so some of that work will already be being done...

This is getting messy, but I just can't imagine the approach of trying to force runtime data through schematic shaped holes is ever going to be a good experience. I've taken three runs at designing this feature and I always end up back at this point, getting stuck.

> You would just have to put into updating your scenes if your underlying representation changes to a reasonable degree.

But if we ignore the two-way binding problem, our above discussed proposal for schematic interpretation doesn't have this problem, does it? I think that's already pretty good at maintaining long-term stability.

SamPruden:
> Combinators

I'm sorry, I think I'm going to need a code example of them in use! Or coffee, maybe I just need coffee.

SamPruden:
There is another option - we just throw away two-way binding. Play mode can expose the raw runtime data (and UI customisation can still be done there) but it's separate from schematics. We lose the continuity of UX - but maybe that's not too bad.

If the trade-off is weakening other aspects of schematics to get a half-decent runtime schematic UI, vs beautifully elegant schematics but raw data during runtime, I think I might take the latter.

MDeiml:
```rust
DefaultSchematic::<A>.add(DefaultSchematic::<B>).map(|(a, b)| S { a, b }, |s| (s.a, s.n))
```
would be the same as
```rust
struct S {
    a: A,
    b: B,
}
```
The result of the `add` call would be a `SimpleSchematic<(A, B)>` which is "isomorphic" to `S` so it can be converted to that using `map`.

Obviously oftentimes you would want to combine more complex schematics. Maybe ones that were derived.

MDeiml:
> There is another option - we just throw away two-way binding. Play mode can expose the raw runtime data (and UI customisation can still be done there) but it's separate from schematics. We lose the continuity of UX - but maybe that's not too bad. (For core and high-effort components, the UIs can be built to be virtually identical.)
> 
> If the trade-off is weakening other aspects of schematics to get a half-decent runtime schematic UI, vs beautifully elegant schematics but raw data during runtime, I think I might take the latter.

Yes, I had somethings similar in mind, but mix it with schematics that have conversions the other way. Gonna make a mockup as soon as I have time.

SamPruden:
Okay, so if we're on the same page, we're looking at a constellation of features that conspire to make the editor experience.

- Schematics - this is everything we discussed in this thread before I mentioned "readback" and ruined everything. They do stable representation and interpretation, and they're the unit of functionality that designers interact with. They're 100% one-way.
- In edit mode, we maintain two worlds - schematic and runtime. The schematic world is live interpreted into the runtime world, which is what we're actually looking at in the scene view.
- We can toggle the scene hierarchy between the schematic and runtime worlds. We can view property panels for both types of entity, which (by default) just shows their components.
- In edit mode, the runtime data view is readonly. In play mode, the schematic data view is disabled, because we're not syncing back.
- We have a mechanism for optionally building custom UI for components - this same mechanism can be applied to both the schematic world and runtime world views. In the simple case, a custom UI panel corresponds to one component - but it can also correspond to a collection of components, allowing them to be combined into one UI panel in the same way they are in schematics.
- If people want to put effort into giving a particular feature a good runtime UX, they can build a custom runtime UI to look the same / very similar to the schematic UI. Perhaps they would even be able to share code between the two.
- Scene gizmos/handles are either part of the UI, or use a very similar system - again needing to be reimplemented between the two views.

Pros:
- All of the major technical dataflow problems are solved - by just not doing them.
- We're no longer trying to persuade that runtime data can be viewed through the lens of schematic data - and maybe that's a good thing actually.
- We have a natural way of providing runtime UI for components that aren't part of any schematics.
- No extra work is required to deal with entities that have been spawned in without coming from a schematic.
- It's possible to have a consistent UX between editing and runtime views if a little effort is put in - and hopefully it's not too much work.
- For common trivial cases, it may still be possible to auto-implement the runtime UI with derive?

Cons:
- Achieving good UX in both views means duplication of effort.
- It's possible that people will be habitually lazy about creating runtime UI, leading to bad designer experience in play mode.

I think this is the direction I'm leaning in. It's technically feasible, split out into separate parts (UI customisation can be a separate RFC, although I have ideas for it), and avoids unwanted abstraction that hides what's going on underneath. It's strictly more powerful than any system that tries to use schematics for the runtime representation. I dislike the duplication of effort in the UI, but I think it's the least bad outcome. Imperfect but working UX is better than theoretically great UX that's restrictive and broken.

MDeiml:
We're almost on the same page. :smile: 

I'm still of the opinion though that synchronizing between runtime and schematic worlds during runtime is a thing we should try to achieve. Otherwise we're "just as good as Unity"

I'm also convinced that:
* Readback can be implemented with enough effort for 100% of sensible schematics
* Plugin authors can be trusted to write good implementations for readback. Even in situations where derives don't help
* Game authors will be fine with derives for 95% of all schematics
* It is ok if readback is only implemented for 95% percent of schematics. It's also fine if readback is only implemented for 1% or 0%. This is because "schematic" and "runtime" views can be merged easily and in a way that is intuitive to the user

I didn't have time to draw a mock up for this yet, but maybe I can explain in words how I imagine the "merged" UI to look like:

* With the interface currently described in the RFC, schematic entities will be a subset of runtime entities. Also (maybe with some minor changes) it is possible to detect, which components are read back.

* In a "hierarchy / entity" view, runtime entities that are not schematic entities are displayed in a different color or with a specific icon. Runtime entities of which all components are read back could even be hidden / greyed out.

* In a "inspector" view, runtime components that are read back are hidden. Other runtime components are displayed in the same list as schematics components on this entity, but with a different color or with an icon to indicate that.

* You can still switch to "runtime view", which behaves as predicted. The default for both edit and play mode is "schmatic view" though. All built in components have readback implemented, so in most cases there are no changes to the UI when starting the game.

In essence the schematic world is Swiss cheese and you can see the runtime world where there are holes in the schematic world. Switching to "runtime view" is just like taking away this transparent layer.

Pros:
* No duplicate effort for UI
* More consistent experience
* Get more out of derives
* We're not lying to users. We're not doing "magic". Merging is simple enough to be understood visually.

Cons:
* I might be wrong with anything I just claimed. Please suggest where I could be wrong :sweat_smile: 
* More effort needs to be put into UI design for the editor
* Users might misunderstand this, even with good UI design
* More code is needed, this code needs to be maintained and tested

SamPruden:
I'll take some time to think over your reply as a whole - and possibly not get to it all until morning, But I wanted to specifically respond to this bit.

> I'm still of the opinion though that synchronizing between runtime and schematic worlds during runtime is a thing we should try to achieve. Otherwise we're "just as good as Unity"

I think even without any sync back, we're already much better than Unity. 

Customising the runtime component UI - in a manner than can look identical to the schematic UI - is a huge win for us. The big win here is being able to write a single UI panel that targets multiple components. We can still offer an experience that's almost identical to synchronising - without actually synchronising. And technically, it's **much** simpler.

I'd be interested to hear exactly which parts of Unity's system you have issues with. I've only used Unity's ECS enough to conclude that it's not ready to be used.

MDeiml:
grafik
Something like this. The green things are from the runtime world, the white things are schematic components / entities

SamPruden:
Ah right yeah, so interlacing the two worlds in the UI.

Can you outline what problem exactly you think readback/sync solves? Which of Unity's issues remain unaddressed by forward-only + custom runtime UI?

For me, it was "I want the property panel UX to remain relatively consistent while the game is being played in the editor, to make it intuitive to understand the state during gameplay, and to allow for live manipulation of values for testing purposes - e.g. moving entities around."

The consistency portion particularly was where readback came in, because it allowed the exact same UI used for constructing the components to also be used in play mode. However, if we allow the runtime components to have their UI customised in a way that provides an equivalent UX, then that problem is solved.

I also think that it's right that the runtime UI should be implemented separately, because it's solving a separate problem. The number of possible runtime states is far greater than the number of schematic states. The runtime UI can be written in a more accomodating manner, providing a better experience over a wider range of circumstances.

> Readback can be implemented with enough effort for 100% of sensible schematics

I'm still sceptical of this. I don't currently see what particular value readback provides that justifies the technical complexity - either for us in the engine, or the complexity of implementing the schematic correctly for the user.

> Plugin authors can be trusted to write good implementations for readback. Even in situations where derives don't help

I don't think it's a matter of trust, it's a matter of burden. It will be mentally taxing to implement a readback schematic correctly. This feature is meant to be seamlessly easy to use whenever possible.

MDeiml:
> Which of Unity's issues remain unaddressed by forward-only + custom runtime UI?

Not really any as long as the "runtime UI" is consistent with the schematics UI

> Can you outline what problem exactly you think readback/sync solves?

The same problems as runtime UI, just in a better way. IMO it's the same difficulty to implement. Or is there something I'm not seeing?

I'm gonna explain my case. Imagine a complex schematic for which derives don't work. I don't have a good example for this case. But for now let's just say the schematic component looks like
```rust
struct Complex {
    a_number: u32,
    some_enum: SomeEnum,
    a_vector: Vec3,
}
```
And let's imagine that this relates in a non trivial way with runtime components `ComplexRuntime1`, `ComplexRuntime2`, ...

Now a UI for this would probably look like.
```
+---------------------------------+
| Complex                         |
+---------------------------------+
| A number   [number input field] |
| Some enum  [dropdown field]     |
| A vector   [x] [y] [z]          |
+---------------------------------+
```

Now I'm pretty sure, that something that maps `ComplexRuntime1`, ... to this UI would mean calculating the contents of the number input field, dropdown field, and xyz inputs (which is exactly the data the schematic contains) **plus** define how those variables are shown in the UI **plus** how changes to the UI translate back to `ComplexRuntime1`, ...

Compare this to readback where you just have to define the first of those 3 parts since the other two are already defined for the schematic.

SamPruden:
For the UI I've been imagining a Imgui style API.

Functions to define both the schematic and runtime UIs might look something vaguely like this:

```rust
pub fn complex_schematic_ui (schem: &mut ComplexSchematic, ui: UiContext) {
    ui.field("A number", &mut schem.a_number);
    ui.field("Some enum", &mut schem.some_enum);
    ui.field("A vector", &mut schem.a_vector);
}

pub fn complex_runtime_ui (components: (&mut ComplexRuntime1, &mut ComplexRuntime2), ui: UiContext) {
    let (one, two) = components;
    ui.field("A number", &mut one.a_number);
    ui.field("Some enum", &mut one.some_enum);
    ui.field("A vector", &mut two.a_vector);
}

// ...
Editor.register_component_panel(complex_schematic_ui);
Editor.register_component_panel(complex_runtime_ui);
```

We do some type of "query" for the components we're providing UI for - represented here by a plain tuple out of laziness. We allow things like `Option<&mut ComplexRuntime3>` for flexibility, too.

Effort has been duplicated in this case, but the code is trivial. If we were indeed dealing with `Option<ComplexRuntime3>`, then the runtime UI might branch and optionally draw an extra field, justifying it being a different implementation from the schematic one. (There could also be a mechanism to add new components from the UI function to account for those cases.)

These are simple enough that they can be automated in basic cases.

In the most trivial case - where the runtime component and schematic component are identical - then the default UIs for both will already be identical. It works out of the box with no effort.

If we create a schematic that outputs multiple components or has some other non-trivial mapping to runtime fields, we're encouraged to provide a corresponding runtime UI for consistency.

We're also able to choose somewhere whether this "consumes" the constituent components - meaning their plain representation is hidden in the inspector - or whether the individual components still remain inspectable as well.

SamPruden:
> The same problems as runtime UI, just in a better way. IMO it's the same difficulty to implement. Or is there something I'm not seeing?

In my mind, it's fundamentally impossible to solve readback in the general case because it relies on a faulty assumption - bijectivity. There may not be a unique mapping from runtime data -> schematic.

And it's reasonable for that to be the case!
- A legacy schematic field isn't being used anymore.
- All field values `>=100` simply map to `Size::Large`.
- During gameplay, the component may enter a state that is disallowed in the schematic, i.e. health > 100
- The schematic outputs five components, but one of them has been removed during gameplay. Does that schematic still apply or do we just completely lose our nice UI here? If we still apply the schematic, how do we infer the values of that missing data?
- An integer schematic field maps to a float runtime field.
- Countless more

Some of these cases can be solved by trying to do a merge/patch operation on the schematic, but that also causes problems.

The schematic is intended to still be editable, meaning that data can roundtrip `runtime -> schematic, [edit], schematic -> runtime`. This roundtrip has to have 100% fidelity in all cases, otherwise editing one field in the schematic may corrupt some other bit of game state that the schematic touches.

If we've inferred a missing value (or retained the original value in the schematic by patching) then that omperfectly inferred value (read corrupted state) will be pushed into the runtime data by the interpretation pass that happens after the edit.

All of these problems come from trying to force the assumption that `schematic -> runtime, [gameplay changes state in arbitrary ways], runtime -> schematic, [user edits schematic], schematic -> runtime` is a safe roundtrip. There are definitely cases where this will not be possible to do correctly - although they may be relatively rare. However, even in the cases where this is possible, it's a significant mental burden to put on programmers - they have to think through all of the edge cases and prove that the implementation is correct. And subtle edge cases are very easy to miss!

If this ever does go wrong (which I predict it will often) it will likely be a nasty subtle headscratcher of a bug for the designer who encounters it, and may not even realise that the behaviour they're seeing is a bug.

MDeiml:
> but the code is trivial

In the example you are making the readback would also be trivial. In the the syntax of the current RFC it would be
```rust
fn infer_complex(query: InferenceQuery<(&RuntimeComplex1, &RuntimeComplex2), Complex>) {
    for ((one, two), commands) in query {
        query.infer(Complex {
            a_number: one.a_number,
            some_enum: one.some_enum,
            a_vector: two.a_vector,
        });
    }
}
```

Or alternatively `Complex` could be changed to
```rust
#[derive(Component, Schematic)]
struct ComplexAlt {
    #[schematic_into(RuntimeComplex1)]
    one: (u32, SomeEnum),
    #[schematic_into(RuntimeComplex2)]
    a_vector: Vec3
}
```
To avoid writing any code. If you want to keep `Complex` as your schematic component, instead of the above `infer_complex` you could use
```rust
ComplexAlt::schematic().map(
    |complex_alt| {
        Complex {
            a_number: complext_alt.one.0,
            some_enum: complext_alt.one.1,
            a_vector: complext_alt.a_vector,
        }
    },
    |complex| {
        ComplexAlt {
            one: (complex.a_numer, complex.some_enum),
            a_vector: complex.a_vector,
        }
    }
```



> These are simple enough that they can be automated in basic cases

Same for readback



> In the most trivial case - where the runtime component and schematic component are identical - then the default UIs for both will already be identical. It works out of the box with no effort

With readback the only code you would need to add at all is
```rust
app.add_schematic::<DefaultSchematic<Complex>>();
```
---

Maybe let's talk instead about a case where the mapping between `Complex` and `ComplexRuntime1`, ... is not trivial. Let's say `SomeEnum` has variants `Representation1` and `Representation2`.

For `Representation1` `RuntimeComponen1` contains `a_number` and `RuntimeComponent2` contains `a_vector`. For `Representation2` there is another component `RuntimeCoponent3` which contains a 4 component vector `(a_vector.x, a_vector.y.log(10), a_vector.z * a_number, a_vector.z)`.

MDeiml:
> * A legacy schematic field isn't being used anymore.
> 
> * All field values `>=100` simply map to `Size::Large`.
> 
> * During gameplay, the component may enter a state that is disallowed in the schematic, i.e. health > 100
> 
> * The schematic outputs five components, but one of them has been removed during gameplay. Does that schematic still apply or do we just completely lose our nice UI here? If we still apply the schematic, how do we infer the values of that missing data?
> 
> * An integer schematic field maps to a float runtime field.
> 
> * Countless more

How would those issues be addressed for "custom runtime UI"? Why can't the schematic be changed?

> The schematic is intended to still be editable, meaning that data can roundtrip `runtime -> schematic, [edit], schematic -> runtime`. This roundtrip has to have 100% fidelity in all cases, otherwise editing one field in the schematic may corrupt some other bit of game state that the schematic touches

With custom runtime UI the `runtime -> UI, [edit], UI -> runtime` has to have 100% fidelity.



> If this ever does go wrong (which I predict it will often) it will likely be a nasty subtle headscratcher of a bug for the designer who encounters it, and may not even realise that the behaviour they're seeing is a bug

This can only go wrong when editing schematics manually in play mode, where bugs are not really critical and the result of those bugs goes away once you leave play mode. In this case I don't see why it wouldn't go wrong with the custom runtime UI implementation.

Also why would this happen "often"? I think we agree that most cases are so simple that derives can be used.

MDeiml:
> Maybe let's talk instead about a case where the mapping between `Complex` and `ComplexRuntime1`, ... is not trivial. Let's say `SomeEnum` has variants `Representation1` and `Representation2`.

The readback variant of this would look like
```rust
fn infer_complex(query: InferrableQuery<(Option<(&ComplexRuntime1, &ComplexRuntime2)>, Option<&ComplexRuntime3>)>) {
    for ((representation_1, representation_2), commands) in query {
        match (representation_1, representation_2) {
            (Some((one, two)), _) => commands.infer(Complex {
                a_number: one.a_number,
                a_vector: one.a_vector,
                some_enum: SomeEnum::Representation1,
            }),
            (_, Some(three)) => commands.infer(Complex {
                a_number: three[2] / three[3],
                a_vector: (three[0], three[1].exp(10), three[3]),
            }),
            _ => (),
        }
    }
}
```

SamPruden:
> Why can't the schematic be changed?

Because the main purpose of the schematic schematic is to be a stable representation of the data. It holds the semantic information required to create the runtime world in a way that's independent of how that runtime data is represented. Not having to change when runtime data gets refactored was the original reason for proposing the feature,

A year old asset file built out of schematics should work today in the latest dev build. If the format is stable, it can do that fine. If we tweaked a schematic because of a change to the runtime data, that's gone.

> How would those issues be addressed for "custom runtime UI"? 

The custom runtime UI has no stability requirement and can be changed at any time.

It also doesn't force the data into the shape of the schematic struct. It operates directly on the runtime components - there's no conversion process where information can be lost.

> Also why would this happen "often"? I think we agree that most cases are so simple that derives can be used.

"often" among manually implemented schematics. The 80% of trivial ones would be fine, yes.

It just feels finicky to get right, which means people will often get it wrong if they rush through it. I know I would.

SamPruden:
What would your inference function look like in this case?

```rust
struct FooSchematic(i32);
struct FooComponent(f32);
```

SamPruden:
My absolute favourite thing to come out of this whole design process so far is the fact that "schematic" and "component" have the same number of characters and therefore line up perfectly in a fixed width font. It's beautiful.

MDeiml:
So true

MDeiml:
> Because the main purpose of the schematic is to be a stable representation of the data

I don't really agree, but I guess you invented the term, sooo. In my opinion the main purpose is to be the representation that is used for UI. I guess we just don't agree here, but in the end it's also the code authors decision what they use it for and not ours. We can just make suggestions.

> The custom runtime UI has no stability requirement and can be changed at any time

I see that point. So in the end there would be four representations: schematic, schematic UI (which hopefully is not that different), runtime and runtime UI (which hopefully is the same as schematic UI).

Changing runtime UI without changing schematics means though, that the UI would suddenly behave different when entering play mode. Maybe in a way that is not noticeable to the user. That seems dangerous to me.

> there's no conversion process

There is a conversion process to UI components. Imagine for example that rotation in the UI is given in euler angles. There is a tiny error in floating point arithmetic when converting quaternions to euler angles and back. This error would appear both for readback and custom runtime UI.

> What would the inference function look like in this case?

The author could either choose to round the float (there is an error, but the error would not multiply over time, also obviously the error would only come up if the designer would actually edit the schematic during runtime). Alternatively the author could choose to change the schematic. I would strongly suggest the second option. Otherwise the value couldn't be set to e.g. 1.5 during the actual editing, so outside play mode.

---

Updated pro / contra (just meant to keep this structured):

Pro readback:
* No duplicate effort for UI (I think we agree)
* More consistent experience (Not sure you agree)
* ~get more out of derives~ I guess runtime UI could also be derived with a bit of extra effort
* With the "merging" I suggested, readback is only optional (Not sure if you like that solution)

Pro runtime UI:
* Easier to implement for us (IMO we're just leaving the problem to whoever implements UI)
* Allows schematic representation to drift from runtime representation (I don't think that's a good thing)
* Easier to implement for users (I don't agree)

MDeiml:
I feel like we're at a point where it might be possible that we can't agree. So I'd offer a different way to look at it. The scope of this RFC is schematics, not UI. Obviously we need to make sure that a UI can actually built from what we're creating, but we don't need to think about the specifics.

This means that we shouldn't focus on whether custom runtime UI is useful. I actually agree that it is. The question is rather if readback is useful, if it is implementable and if it can cause unsafety / frustration:

I think we might agree that it is useful. I claimed that it is also implementable, but obviously that question will be answered as soon as I write the "implementation" section in the RFC for this.

So the remaining question is, whether it can cause unsafety or frustration. I take it that your answer to this is "yes"?

SamPruden:
> I don't really agree, but I guess you invented the term, sooo. In my opinion the main purpose is to be the representation that is used for UI. I guess we just don't agree here, but in the end it's also the code authors decision what they use it for and not ours. We can just make suggestions.

TBH I don't understand how instability could ever be acceptable here.

I think we're agreed that schematics are the format in which assets/scenes are serialized. In that case, how can it be acceptable for that format to be unstable? That would break compatibility with assets already built - which is a huge deal!

There's **some** about of tolerance here if we e.g. use Serde for serialization and use its field attributes to handle simple migrations, but that's fairly brittle and we want it to happen as little as possible.

> So in the end there would be four representations: schematic, schematic UI (which hopefully is not that different), runtime and runtime UI (which hopefully is the same as schematic UI).

I don't quite think of the UIs as "representations" because the data is never stored in them - they're more like mapping functions from `(state, user input) -> modified state`, but I suppose that's only an incidental semantic difference.

However, customising the UI is optional in both places, so it's two representations + more if you care about making this one pretty.

The simple case of just combining multiple runtime components into one UI could be automated too - it would just have to inline their default representations next to each other in one panel. A simple attribute or one-liner could take care of that.

> Changing runtime UI without changing schematics means though, that the UI would suddenly behave different when entering play mode. Maybe in a way that is not noticeable to the user. That seems dangerous to me.

I think it's on the programmer not to do anything too silly here.

> There is a conversion process to UI components. Imagine for example that rotation in the UI is given in euler angles. There is a tiny error in floating point arithmetic when converting quaternions to euler angles and back. This error would appear both for readback and custom runtime UI.

This is a great point I'd never thought of - the imgui pattern might accumulate precision error in some cases! However, this is how Unity (and I think Unreal and Godot?) do it today and I've never heard of anybody complaining about it being a problem, so I guess it's not a big deal in practice.

> The author could either choose to round the float (there is an error, but the error would not multiply over time, also obviously the error would only come up if the designer would actually edit the schematic during runtime). Alternatively the author could choose to change the schematic. I would strongly suggest the second option. Otherwise the value couldn't be set to e.g. 1.5 during the actual editing, so outside play mode.

Scenario: My game is on a grid, and it's only allowable to set those positions to grid-aligned values. However, they're floats at runtime because they can be animated between grid positions. 

> Updated pro / contra (just meant to keep this structured):

Good summary!

> No duplicate effort for UI (I think we agree)
> More consistent experience (Not sure you agree)

Agree ish. However, because schematic data and runtime data are sometimes fundamentally different, forced consistency may be bad.

> With the "merging" I suggested, readback is only optional (Not sure if you like that solution)

Equally true for the UI route, where customisation only happens if you don't want to accept the defaults. Mixing defaults and customised is basically the same experience as merging.

SamPruden:
> I feel like we're at a point where it might be possible that we can't agree. 

I fear we may be.

> So the remaining question is, whether it can cause unsafety or frustration. I take it that your answer to this is "yes"?

Yes.

I contest that any solution that you come up with, I will be able to give you plausible real-world scenarios that readback will not be able to handle satisfactorily, but that will be handled by the UI route. You're welcome to take that as a challenge. 😂

SamPruden:
I think I'm going to put together a "rival" RFC - of course intended in the spirit of cooperation not rivalry! That way we can get a clear look at both approaches side by side, and perhaps bring others in to comment at that stage. I may play around with a prototype implementation too, as I find that's usually a good way to be quickly confronted by all of the ways in which I'm an idiot.

MDeiml:
Sure, that will help a lot with discussion.

MDeiml:
> You're welcome to take that as a challenge

I'm going to, but first we need to solve a different question that I forgot just now and that is probably more important. I'm sorry :facepalm:

That is the question of "stable format", which we both agree is one of the motivations of what we're doing. You (if I understand correctly) see this as the most important motivations and interpret stable as "stable, unless stability is impossible". I see usage in editors as the most important motivation and interpret stable as "stable, unless the runtime representation changes enough". I'm gonna go out on a limb here and define my position a bit stricter as "stable, unless schematics are no longer surjective" ("surjective" meaning that every sensible runtime state can at least be reached through schematics).

I'd like to give some context: Unity AFAIK does not have stable formats in your sense. If you change the attributes of a custom `MonoBehaviour`, then data is lost. In the example of changing euler angles to quaternions for rotation for instance, the data would just be lost. Also, if you open a project with a newer version of Unity it warns you that data could be lost.

Do I understand correctly that you would want bevy to behave differently?

MDeiml:
I feel like if we don't agree on what's motivation schematics, then there's no way we'll agree on a design :laughing:

SamPruden:
> That is the question of "stable format", which we both agree is one of the motivations of what we're doing. You (if I understand correctly) see this as the most important motivations and interpret stable as "stable, unless stability is impossible". 

Stability unless stability is impossible, in which can friction should be minimized.

If a game feature gets radically redesigned, then stability from old version assets to new is probably nonsense.

If a feature's **implementation** gets changed (i.e. new/different systems - operating on a different ECS layout) stability should be maintained. I can write a whole new physics engine and keep the same rigidbody schematic, as long as the engine is exposing the same concepts.

Serde style migration attributes on the schematic are acceptable as a last resort.

> every sensible runtime state can at least be reached through schematics

I would say it's every runtime state that the programmers want the designers to be able to create.

I actually think it's perfectly sensible to create a `LockedToGroundTransform` schematic, specifically intended to apply an artificial limitation to the transform. (That could function in a couple of different ways, either by **being** the transform, or overriding its sibling `TransformSchematic`). It would still be dealing with a plain `TransformComponent`.

In programming terms, schematics are a tool for encapsulation. Schematics are the public interface to the state - and programmers can choose what to make public, and what to keep private. Some states can only be reached "privately" - i.e. through progression of game logic. That's a good thing, because many states would be nonsensical - e.g. unnormalized direction vectors.

The only reason you need to relax the stability from full to surjective is to accomodate the problems with the readback approach.

> Unity AFAIK does not have stable formats in your sense.

Sort of.

In the case of Monobehaviours, they don't. That's bad and I want to fix it. However, it's a slightly less bad problem for them. 

Monobehaviours, being managed classes, are not particularly sensitive to data layout. Being stuck with an existing layout of serialized fields for backward compatibility is usually okay - you just do the equivalent of "interpreting" that data in `OnEnable()` to get it into the format you want to consume it in.

In Bevy we're data and performance oriented all the way down. Data layout is very sensitive, and is also something we want to be able to change freely, either as we change our designs or improve performance. Being stuck for backward compatibility reasons is a much bigger problem.

When it comes to their ECS prototype, they sort of do have a stable layer. They use Monobehaviours to serialize their assets, and an interpretation pass to turn that into ECS data. Generally, people will avoid breaking the stability of those Monobehaviours for exactly this reason.

MDeiml:
The reason why this is important, is because this dictates if "you can implement readback for every sensible schematic". If the schematic component covers every combination of components that are meant to come up during play, in mathematical terms if they are surjective on to the set of states encountered during runtime, then (again mathematically speaking) there exists an inverse function.

If on the other hand a situation like example with `f32` and `i32` that you gave is "allowed" / encouraged, then this is not the case. Notice that here the schematic that uses `i32` is not "surjective", i.e. does not cover all cases that are encountered during runtime. (If the `f32` values encountered during runtime are exclusively whole numbers, then it is possible to implement readback).

I hope I'm not misinterpreting your answer, but our difference of opinion boils down to:

* "Schematics should be able to represent everything encountered during runtime" (I actually don't think this has to be true for every schematic, but I'm gonna ignore that for simplicity)

vs

* "Schematics should be able to represent everything encountered when editing levels"

I feel like both are very much valid. It's probably not surprising then, that I see schematics as something that can also be used during runtime while you disagree.

SamPruden:
Excellent summary!

My stance is that surjectivity is an extra restriction, and we shouldn't make decisions that constrain our users without very clear justification.

The fact that we disagree over whether the `i32`/`f32` example makes sense illustrates that some people may want to be able to do that and IMO we should cater to them. 

I'm not persuaded that there's anything that the readback approach does so dramatically better than the UI approach as to justify these restrictions.

I think the path forward is that we both write up and/or prototype our respective versions, then come back and compare how they look in practice - perhaps challenging each other to show how we'd deal with some challenging scenarios.

MDeiml:
Now since we won't come to agree on the topic of "stable data", I'm gonna adopt your stance on that for discussions sake.



> I contest that any solution that you come up with, I will be able to give you plausible real-world scenarios that readback will not be able to handle satisfactorily, but that will be handled by the UI route. You're welcome to take that as a challenge.

This of course means that I agree that there are schematics where you couldn't implement readback in a satisfactory way. But I'd still like to have readback as an optional feature of schematics, since I believe that the situation where it can't be implemented are very rare. A editor UI would find a way to deal with these rare cases, whether through "custom runtime UI", "merging" or some solution we haven't thought of yet.

I'd expect, that if someone learns of a feature in bevy, which helps in converting one representation into another, then a natural question would be "Is there also a way to convert back?". I'd like the answer to be "yes". Even if not, it makes sense to ensure that our design at least **has the possibility** for conversions the other way.

MDeiml:
> challenging each other to show how we'd deal with some challenging scenarios

Haha, not sure how productive that would be, but sounds fun. I imagine that - having different goals as well as different scopes (I guess you're also going to talk about UI in your RFC) - we just wouldn't agree on which is better xD

SamPruden:
Github says our discussion is too robust. I got locked out for 10 minutes by a rate limit! 😂

Yep I agree with all of that. My challenge to myself is to make the UI proposal good enough to persuade you that readback is obsolete. I reckon I've got a 50% shot at that.

I'm going to take some time to collect my thoughts and come back with a draft RFC and/or prototype - probably in a couple of days if I had to guess.

MDeiml:
Didn't even know github had a rate limit for discussions xD

Sure, take your time.

Diddykonga:
I'm very unknowledgeable about the whole of the conversation/discussion, but I figured I would give my view on the subject anyways. 

Schematics - Editor or Artist representation of game data/systems.
Runtime - Runtime or Programmer representation of game data/systems.

Two easy examples brought up so far was:

1.     Schematic - Integer grid-based positions for the Editor/Artists.
       Runtime - Float animation-based positions for the Runtime/Graphics.
2.     Schematic - Vector3 euler-based rotations for the Editor/Artists.
       Runtime - Quaternion imaginary-based rotations for the Runtime/Graphics.

    
The idea of stability was brought up for Schematics and while they would/should provide stability from changes to the Runtime representation, Schematics themselves would/should not be stable because an Artist/Editor-user may need to change that representation as they see fit.

How Archetype Invariants + Entity Kinds would work with Schematics I am unaware, perhaps these are Runtime constraints only.

SamPruden:
> Schematics themselves would/should not be stable because an Artist/Editor-user may need to change that representation as they see fit.

Just to check that I'm understanding you @Diddykonga, do you mean that a designer should be able to choose e.g. whether to input a rotation as Euler angles or as a quaternion?

If so, I think that's best dealt with one of two ways:

1. The UI lets you type the numbers in either format, but it always serializes is one canonical format.
2. Represent the rotation as a `enum Rotation { Euler(Vec3), Quaternion(Vec4) }`.

Both of these methods could be used with either myself or @MDeiml's preferred approaches.

Diddykonga:
> Just to check that I'm understanding you @Diddykonga, do you mean that a designer should be able to choose e.g. whether to input a rotation as Euler angles or as a quaternion?

My apologies, what I meant is that there would always be one way for editor and one for runtime, but that both should have mutability/migration-ability, which was why I said it is unstable. Although I may be misunderstanding what stable is in this context ¯\_(ツ)_/¯

So as long as the Editor Schematic is Euler Angles, that's what a designer/artist would use, but should also be able to change it without changing the Runtime Schematic, to support designer/artist workflows more precisely.

MDeiml:
I think you understood stability right. "Stability" would mean that the actual data structures used for the schematic representation don't change at all under any circumstances.

If I understand correctly you're actually mentioning another point that we haven't yet talked about. Which is that artists / designers might want to **intentionally** change the data representation they see in the UI.

An example would be if the developer / non-designer initially adds rotation to the UI as quaternions. In this situation a designer would probably complain that they don't understand quaternions and would rather like to input euler angles. The developer would then need to change the schematics to meet that request.

So I guess optimizing for designer workflow is contrary to the schematic representation being stable...

SamPruden:
In practice, what matters is long term stability. I want to be able to make runtime implementation changes without worrying about breaking assets that were built months ago. One would hope that the way that example would resolve itself is that the arist would complain as soon as the schematic was given to them, and changes would be made before a backlog of content was built on the bad version. Stability only matters once the team start relying on it.

Technically stable doesn't have to mean that the schematic struct layout can't change at all - the real requirement is that any previous version can sensibly deserialize into the current version.

If we assume that serialization is being done by `serde` or something with similar capabilities, then I believe that in the absolute worst case it should be possible to provide a manual implementation of `Deserialize` for the schematic type that deals with arbitrary migrations - we should probably include a version tag in the serialization format to facilitate this.

Having to do that is tedious. Trying to support deserialization from multiple prior versions of the same struct is error prone, and probably nescessitates automated tests too. Nobody wants to do that for their supposedly quick & seamless schematics, but the fact that this is possible means that even the worst case isn't a catastrophe.

PR #65, opened 2022-11-01
Topics

mxgrey:
This RFC propose a reactive pub-sub framework for Bevy, organized by topics.

I anticipate this RFC will need a considerable amount of feedback and iteration. I'm opening this to help facilitate technical discussion and not to declare that I have a firm solution for this feature.

If this concept is of significant interest, should we go ahead and reserve the `bevy_topics` crate?

### DIFF
--- a/rfcs/xx-topics.md
+++ b/rfcs/xx-topics.md
@@ -1,0 +1,295 @@
+# Feature Name: Topics
+
+:warning: This proposal depends on [one-shot systems](https://github.com/bevyengine/bevy/pull/4090), which is not yet a merged feature.
+
+## Summary
+
+This RFC introduces a feature to help with developing reactive applications.
+Topics are data channels for moving asynchronous messages through many-to-many relationships between publishers and subscribers.
+Each subscriber can listen to messages from many publishers, and many subscribers can listen to each publisher.
+
+## Motivation
+
+Topics are a common mechanism for middlewares to pass data between distributed systems.
+Topics are also called pub-sub (publish-subscribe) and are supported by middlewares like ROS, MQTT, ZeroMQ, Kafka, RabbitMQ, and many others.
+In these systems, anonymous publishers are able to push out messages to "topics" (data channels) which anonymous subscribers can listen to.
+This allows the development of distributed reactive systems, organized by the abstraction of "topics" so that publishers can be easily discovered by relevant subscribers without needing to know details of how the overall distributed system is implemented.
+
+This RFC proposes implementing topics inside of Bevy as ergonomic Systems, Resources, and Components.
+
+Topics can be used inside of Bevy to help Bevy users implement their application using a "distributed" architecture to decouple and organize event flows from various agents/entities.
+This is similar to a previous suggestion about [Entity Events](https://github.com/bevyengine/bevy/issues/2070) and has similar motivations.
+Bevy's current resource-based Event mechanism is not convenient for representing more complex flows of information that may be entity/agent-driven or entity/agent-sensitive.
+
+At the same time, defining this generic ergonomic pipeline for topics within Bevy should help facilitate the integration of various middlewares into Bevy.
+For example, a `bevy_mqtt` plugin could simply subscribe to bevy topics and then effortlessly publish the received messages to MQTT topics, and vice-versa for two-way communication.
+
+Bevy users who want their application to have networking or inter-process communication can implement their systems in terms of the bevy topics and services, then plug in whichever middleware suits their overall networked / multi-process system.
+They could also change their choice of middleware later without changing how the systems of their application are implemented.
+Or multiple middlewares can be used by one application without affecting how the systems are implemented.
+
+## User-facing explanation
+
+### Topics
+
+These are the key structs for topics:
+* `Publisher<T>`: a component struct that provides a `publish(data: T)` method
+* `Subscriber<T>`: a component struct that provides a `subscribe<T, P: SystemParam, F: Fn(T, P)>(publisher: Entity, callback: F)` method
+* `Domain`: a component that maintains a dictionary of topic names along with the publishers and subscribers that are interested in each topic
+* `GlobalDomain<T>`: a Domain that can be used as a resource instead of as a component
+
+To send out messages, spawn a `Publisher<T>` component for an entity, optionally giving some initial messages:
+
+```rust
+struct MyMessage {
+    data: String,
+}
+
+fn make_a_publisher(
+    commands: Commands,
+) {
+    let mut publisher = Publisher::<MyMessage>::new();
+    publisher.publish(MyMessage{data: "Hello, world!".to_string()});
+    publisher.publish(MyMessage{data: "I am being spawned!".to_string()});
+    commands.spawn().insert(publisher);
+}
+```
+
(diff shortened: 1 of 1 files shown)


james7132:
Haven't had the time to fully read through the RFC, but a quick skim through raises the following questions:

 - Why can't this be an ecosystem crate? Why must this be core to the engine itself?
 - What does this offer that `Events` do not? Besides the fact that events are global resources, events are already MPMC by design. What benefits and tradeoffs are we making by architecting it this way over existing events?
 - The finer implementation details here are important and are largely hand-waved away. If we're looking at topics as distributed message queues in the traditional sense, they must be guarded from aliasing access. This will require at the very minimum guarding atomic accesses which can be detrimental to performance considering bevy's ECS is currently requires zero atomic accesses on a per-entity basis.
 - *When* in the schedule do subscriber's handlers run? Immediately on publish? In a different stage? During a pre-created system? If so, how do we schedule other systems relative to it?
 - Why take this approach to networking over established game networking flows? Historically, games have been bottlenecked more by latency than distributed consistency. Simply using a distributed message broker (i.e. MQTT) may counter these goals.

alice-i-cecile (commented):
Seconding James's comments. Seems interesting, but definitely feels like something that can and should be explored extensively in a 3rd party crate.

mxgrey:
Thanks for the incredibly quick feedback!

> Why can't this be an ecosystem crate? Why must this be core to the engine itself?

Totally agree that this can/should be an ecosystem crate. I opened the PR here to seek inputs from anyone else who may be interested in this idea and to avoid duplicating effort in case someone else is already working on this concept. If that's an inappropriate use of RFCs then feel free to close this PR and I'll seek out a new home for the document.

All the same applies to #66 .

> What does this offer that Events do not?

I think there are several key benefits, although the value of these benefits is situational, so I expect topics to complement Events and not replace them:
* Messages are scoped to publishers and topics, so subscribers can be set up to react to messages differently based on their source, without needing to embed extra information in the message data
* Subscription callback systems will only be run when one or more new subscribed messages are available instead of being run on every cycle the way an Event reaction system would be
  * Admittedly it's likely possible to implement a subscription+callback system for Events to achieve this same benefit

> The finer implementation details here are important and are largely hand-waved away.

Very much acknowledged at the top of [Implementation Strategy](https://github.com/mxgrey/bevy_rfcs/blob/topics/rfcs/65-topics.md#implementation-strategy) :sweat_smile: 

> If we're looking at topics as distributed message queues in the traditional sense, they must be guarded from aliasing access.

Acknowledged in the [parallel publishing](https://github.com/mxgrey/bevy_rfcs/blob/topics/rfcs/65-topics.md#parallel-publishing) section.

> When in the schedule do subscriber's handlers run?

Acknowledged in the [Callback drive](https://github.com/mxgrey/bevy_rfcs/blob/topics/rfcs/65-topics.md#callback-driven) and [Callback timing](https://github.com/mxgrey/bevy_rfcs/blob/topics/rfcs/65-topics.md#callback-timing) sections.

> Why take this approach to networking over established game networking flows?

Truthfully I'm interested in using bevy in robotics, simulation, and IoT contexts rather than for conventional video game development, so I won't have a good answer for this since I'm not familiar with how client-server communication is done for conventional video games. But for the context that I'm coming from, distributed pub-sub architecture is a must-have, so making that accessible in bevy would open bevy up to whole new industries outside of video game development. Rising tide lifts all boats and whatnot.

### But really, why not Events??

As I think more about it, I wonder if it would be possible to piggyback directly on the existing Events implementation to make scoped pub-sub work. That may also help to solve the concurrency concerns.

Maybe instead of `Publisher<T>` being a component it could be a `SystemParam` that wraps around `EventWriter<Message<T>>` where:

```rust
pub struct Message<T> {
  for_topic: Topic,
  data: T,
}
```

Then each time you use `Publisher<T>` you just specify the entity or topic that's being published for. `Entity` could be convertible into a valid `Topic` object.

Each message type `T` would need to be registered so that an `Events<Message<T>>` can be registered for it so that `Publisher<T>` can be queried. Registering the message type `T` would make it much easier to generate systems that can query for `Subscription<T>` components that contain callbacks which need to be triggered. Then since Bevy can already recognize when an event was not registered correctly, it will transitively recognize when a message type was not registered correctly.

Does that seem like a more feasible approach?

Scheduling callbacks still presents some challenges, especially since this feature would need to support stageless systems and I'm not familiar with those yet. I'll need to get up to speed on stageless before I can reason about that.

alice-i-cecile:
> Maybe instead of Publisher<T> being a component it could be a SystemParam that wraps around EventWriter<Message<T>>

Unsure if this will be easier than making it a custom `SystemParam` :)

> Scheduling callbacks still presents some challenges, especially since this feature would need to support stageless systems and I'm not familiar with those yet. I'll need to get up to speed on stageless before I can reason about that.

Scheduling callbacks should be pretty straightforward post-Stageless: they'd just be emitted as commands, which can be evaluated whenever the user inserts a system that processes them ( a "command flush point"). You may end up wanting some performance improvements here; commands could be faster and the callbacks will be evaluated completely sequentially, but eh, for tiny callbacks that may actually be faster.

mxgrey:
> Unsure if this will be easier than making it a custom SystemParam :)

That's what I was thinking! :grin: 

Something like:
```rust
#[derive(SystemParam)]
pub struct Publisher<'w, 's, T> {
    writer: EventWriter<'w, 's, Message<T>>,
}

impl<'w, 's, T> Publisher<'w, 's, T> {
    pub fn publish(&mut self, to_topic: Topic, msg: T) {
        self.writer.send(Message::new(to_topic, msg));
    }
}
```

Then there'd be a `Subscriber` custom system param that can generate `Subscription<T>`:
```rust
#[derive(SystemParam)]
pub struct Subscriber<'w, 's> {
    commands: Commands<'w, 's>,
}

impl<'w, 's> Subscriber<'w, 's> {
    pub fn subscribe<T, F>(&mut self, to_topic: Topic, callback: F) -> Result<Entity> {
        // TODO: Do something to make sure that T is registered as a message type
        // and return Err if it is not.
        Ok(self.commands.spawn().insert(Subscription::new(to_topic, callback)).id())
    }
}
```

Then there'd be a system inserted for each `T` message type to read `Message<T>` events and query for `Subscription<T>` components that are storing callbacks. When a message is published to a topic that matches the topic a subscription is listening to, the callback gets triggered, ideally with a way for the subscription to specify exactly when to trigger it.

I'm still hand waiving a lot when it comes to callback scheduling, but I think the approach to connecting publishers and subscriptions is feeling a lot more concrete. It seems like the `Domain` concept from the proposal might be thrown out, or perhaps just incorporated into the `Topic` concept.

alice-i-cecile:
You may also be interested in #17 and #32, which tackle related problem spaces :)

huang12zheng:
I'm new to rust and I'm just saying what I think.

1. ancestor can create a Container::<Mission>;
2. What can posterity do?
let mut publisher = Publisher::&lt; Mission&gt; ::new();
let mut subscriber = Subscriber::&lt; Mission&gt; ::new();


1. If ancestor is no Container::<Mission>, posterity can be shared with everyone or subscribed to by everyone
2. If ancestor has a Container::<Mission>, all the posterity would only see it, they would ignore higher Container::<Mission>
> root can have a Container, and None won't happen

PR #71, opened 2023-06-25
Write only world query

nicopap:
#### Summary

Add a `WriteOnly<T>` `WorldQuery`. `WriteOnly` acts like `&mut`, but tells bevy
that the value of component `T` is overwritten unconditionally by the system.
Add the `B0006` error. This tells users that they are modifying a component
that is later overwritten unconditionally.

### [RENDERED](https://github.com/nicopap/rfcs/blob/write-only-world-query/rfcs/71-write-only-world-query.md)

### DIFF
--- a/rfcs/71-write-only-world-query.md
+++ b/rfcs/71-write-only-world-query.md
@@ -72,6 +72,25 @@ Consider `ui_layout_system`:
 
 https://github.com/bevyengine/bevy/blob/469a19c290861049ad121aa83d72f830d1ec9b40/crates/bevy_ui/src/layout/mod.rs#L215-L230
 
+```rust
+pub fn ui_layout_system(
+    primary_window: Query<(Entity, &Window), With<PrimaryWindow>>,
+    windows: Query<(Entity, &Window)>,
+    ui_scale: Res<UiScale>,
+    mut scale_factor_events: EventReader<WindowScaleFactorChanged>,
+    mut resize_events: EventReader<bevy_window::WindowResized>,
+    mut ui_surface: ResMut<UiSurface>,
+    root_node_query: Query<Entity, (With<Node>, Without<Parent>)>,
+    style_query: Query<(Entity, Ref<Style>), With<Node>>,
+    mut measure_query: Query<(Entity, &mut ContentSize)>,
+    children_query: Query<(Entity, Ref<Children>), With<Node>>,
+    mut removed_children: RemovedComponents<Children>,
+    mut removed_content_sizes: RemovedComponents<ContentSize>,
+    mut node_transform_query: Query<(Entity, &mut Node, &mut Transform, Option<&Parent>)>,
+    mut removed_nodes: RemovedComponents<Node>,
+) {
+```
+
 We would replace
 
 ```rust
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
Hmm. I'd really like to not use `Transform` for UI, solving one of these cases at least.

rfcs/71-write-only-world-query.md
```diff
@@ -0,0 +1,385 @@
...
+
+Daniel reads the error message, goes to <https://bevyengine.org/learn/errors/#b0006>
+and learns that they should — I quote — "schedule `shake_ui` in `PostUpdate`,
+and label it as `shake_ui.after(ui_layout_system)` to effectively update `Transform`".
+
+Daniel updates their app schedule code:
+
+```rust
+app.add_plugins(DefaultPlugins::new())
+  .add_systems(PostUpdate, shake_ui.after(ui_layout_system));
```
cart, line 145: This isn't actually "fully correct" for this scenario. Making changes to the `Transform` after `ui_layout_system` does ensure it won't get stomped on by that system. And for the _current Bevy systems written_ I do think it would work. But the computed node Transform is considered an output of the `UiSystem::Layout` set. Anything trying to consume the "final computed layout " of a node should schedule relative to that. Therefore if you are modifying Transform (and affecting final layout), you should put your system in `UiSystem::Layout`.

rfcs/71-write-only-world-query.md
```diff
@@ -0,0 +1,385 @@
...
+    self.0.reborrow()
+  }
+  pub fn into_inner(self) -> Mut<'a, T> {
+    self.0
+  }
+}
+
+```
+
+`WriteOnly<T>` is very much nothing more than a thin wrapper around `Mut`.
```
cart, line 48: `WriteOnly` seems like the wrong name for this:
* We already have the concept of "read only queries". Which means "you statically do not have permission to do anything but read". This is not the "write" equivalent of that.
* The goal of this feature _is not_ to assert that we only write to a thing (and not read from it). It it is to assert that _all_ values will be written to, with none skipped. Reading before or after the write is completely ok, provided you do the write. `WriteOnly` not only isn't accurate ... it encourages the user to think about it in the wrong way.  If I saw WriteOnly in a query, I wouldn't think "i need to write every item in the query". I would think "i cannot read items in this query but I can still skip items I don't want to write".

nicopap: The no-read semantic is important. What makes `ui_layout_system` handling of `Transform` different from other systems is indeed that it overwrites the `Transform` without leaving a trace of the existing value. Whether it does this to all the entities it access or only a subset is only partially relevant.

```rust
fn sys(mut q: Query<&mut Transform>) {
 for t in &mut q {
    t.translation.x += 1.0;
  }
}
```

Here, a previous system that changes `Transform` still has a visible effect after `sys` runs. So we don't want to say that translation was overwritten.

While for the following:

```rust
fn sys(mut q: Query<&mut Transform>) {
 for t in &mut q {
    if t.translation.x > 10.0 {
      continue;
    }
    *t = Transform::IDENTITY;
  }
}
```

The previous system's value will be overwritten (for a subset of `Transform`) If the user wrote a system that sets a `Transform` that has a `translation.x` smaller than 10.0, that value will never be read afterward.

Our goal here is an error message similar to rust's [`unused-assignments`](https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html#unused-assignments) lint.

nicopap:
> Hmm. I'd really like to not use `Transform` for UI, solving one of these cases at least.

A lot of games requires slight updates to UI positions after layouting.

A lot of game animate UI elements (say: shift right the currently focused element, or give a very slight shifting effect for a camera movement effect, or even animated text). It's a requirement to be able to manipulate the position after the layouting output.

Switching to a different component will still require a layout system to overwrite an existing positional component. This means that whether it's a `Transform`, or `UiPosition`, or whatever it is named, telling the user they are doing it wrong can  still be valuable.

Also, enabling back writes to `GlobalTransform` and `Visibility` is still useful.

An alternative is to define an additional component (say: `UiOffset`) for and handle yourself the offsetting based on the value of that additional component. [It would look like this](https://github.com/nicopap/cuicui/tree/4a54e1fc02609724249b61bf55f5f8ab36f6c7a1/bevy_layout_offset)

But maybe `UiOffset` only exists because bevy is incapable of telling users what went wrong with system ordering. And adding an additional layer of abstraction is not worth it if bevy has this capability.

nicopap:
Note that work on this was inspired by `cuicui_fab`.

`cuicui_fab` is fairly complex and I could write a small book about it. But write-only is extremely useful in it. And I've spent a couple weeks trying to migrate it to use bevy systems, and this is what spurred this proposal. Not that my work on `cuicui_fab` is dependent on this RFC. Quite the opposite in fact.

PR #72, opened 2023-11-12
Interactive Editor

LeifAndersen:


rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
+# Feature Name: `editor-foundation`
+
+## Summary
+
+This RFC outlines the foundation for Bevy's editor. It describes its basic
+components and how they will interact with Bevy's engine. Following this RFC
+will result in a minimum viable product Bevy Editor.
```
alice-i-cecile, line 7:
```suggestion
will result in a minimum viable product for the Bevy Editor.
```

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+
+This RFC does NOT cover the particular user interface (UI) design for the
+editor. Instead, it provides the building blocks for a proper UI design, with a
+minimal usable interface to place these components in context.
```
alice-i-cecile, line 11: This sentence isn't clear to me.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+
+## Motivation
+
+Unlike most major game engines, Bevy lacks an interactive editor. This lack has
+three major downsides:
+
+1. it increases the barrier to entry for potential new developers;
+2. this forces scene designers to learn to create scenes with code; and
+3. developers must rebuild their project before seeing any changes they make.
```
alice-i-cecile, line 20: The ";" make this really hard to read.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+editor. Instead, it provides the building blocks for a proper UI design, with a
+minimal usable interface to place these components in context.
+
+## Motivation
+
+Unlike most major game engines, Bevy lacks an interactive editor. This lack has
+three major downsides:
+
+1. it increases the barrier to entry for potential new developers;
+2. this forces scene designers to learn to create scenes with code; and
```
alice-i-cecile, line 19: I don't think this is the right way to frame it: it's not a learning curve problem. It's a "dramatically slower and more frustrating" problem.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+This RFC does NOT cover the particular user interface (UI) design for the
+editor. Instead, it provides the building blocks for a proper UI design, with a
+minimal usable interface to place these components in context.
+
+## Motivation
+
+Unlike most major game engines, Bevy lacks an interactive editor. This lack has
+three major downsides:
+
+1. it increases the barrier to entry for potential new developers;
```
alice-i-cecile, line 18: I don't think that an editor will meaningfully improve this.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+components and how they will interact with Bevy's engine. Following this RFC
+will result in a minimum viable product Bevy Editor.
+
+This RFC does NOT cover the particular user interface (UI) design for the
+editor. Instead, it provides the building blocks for a proper UI design, with a
+minimal usable interface to place these components in context.
+
+## Motivation
+
+Unlike most major game engines, Bevy lacks an interactive editor. This lack has
```
alice-i-cecile, line 15:
```suggestion
Unlike most major game engines, Bevy lacks an interactive editor. This has
```

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+three major downsides:
```
alice-i-cecile, line 16: I think that "no good way to display visual elements like a schedule visualizer" is a key downside.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+could have spent building the game.
+
+Bevy's plugin-heavy design poses several challenges to making a traditional
+editor. Primarily, unlike other game engines, Bevy does limit developers access
+to the core game loop, similar instead to many game frameworks. While an
+interactive editor could be built on top of Bevy that takes away this freedom;
+doing so violates the spirit of Bevy's plugin-heavy design, thus robbing it of
+Bevy's expressive nature.
+
+**Bevy's editor must acknowledge the duel nature of Bevy as both a game engine
```
alice-i-cecile, line 33:
```suggestion
**Bevy's editor must acknowledge the dual nature of Bevy as both a game engine
```

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+
+1. it increases the barrier to entry for potential new developers;
+2. this forces scene designers to learn to create scenes with code; and
+3. developers must rebuild their project before seeing any changes they make.
+
+While teams can mitigate the last two points by creating project specific
+tools, this does require them to divert resources building those tools that they
+could have spent building the game.
+
+Bevy's plugin-heavy design poses several challenges to making a traditional
```
alice-i-cecile, line 26: I don't think the following paragraphs in this section are helpful. The points that they're trying to make aren't clear.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+
+## User-facing explanation
+
+The easiest way to get started with Bevy's editor by downloading it with Cargo:
+
+```
+$ cargo install bevy
+```
+
+(Note the use of `install`, not `add`!)
```
alice-i-cecile, line 48: Explain the implication here.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+This will add `bevy` to our path:
+
+```
+$ bevy
+Bevy's Command Line Tool
+
+Usage: bevy [OPTIONS] [COMMAND]
+
+...
+```
```
alice-i-cecile, line 59: I think that the code block is imbalanced here.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+
+This tool allows us to create new projects, manage assets, edit projects, and
+even publish games. Note that `bevy` is not meant to replace `cargo`, but to
+work alongside it.
+
+Also, just like Cargo, the Bevy version used in a project does not need to match
```
alice-i-cecile, line 65:
```suggestion
Also, just like Cargo, the Bevy library version used in a project does not need to match
```

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+the version of `bevy` you have installed. You can upgrade and downgrade them
```
alice-i-cecile, line 66:
```suggestion
the version of the `bevy` CLI you have installed. You can upgrade and downgrade them
```

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+independently. Therefore, we recommend you always use the latest version of the
+`bevy` tool.
+
+Bevy's editor can be used graphically, or as a command line tool. This document
+covers the command line usage, but running `bevy edit` in a project will open
+the graphical view.
+
+**The `bevy` command line interface is not stable. Do not use it in build
+scripts!**
```
alice-i-cecile, line 75:
```suggestion
**The `bevy` command line interface follows semver. Be careful when using it in build scripts!**
```

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+project can start from any number of templates you can get from the Bevy asset
+repository.
+
+Next, let's take a look at the project structure:
+
+```
+flappers/
+    README.md
+    Cargo.toml
+    LICENSE-Apache2.txt
```
alice-i-cecile, line 103: We should default to dual MIT + Apache.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+first three files: `RREADME.md`, `Cargo.toml`, and `LICENSE-Apache2.txt` are
+standard meta-information that are in most crates. The next folder: `assets` is
+the default location to place project non-code assets. The bulk of the source
+code goes in the `src` directory, likewise tests go in `tests`. Project builds
+go into the `target` directory.
+
+The `editor` directory provides a sub-crate for the Bevy editor. Associating
+Bevy's editor with the project allows you to have multiple projects using
+different versions of Bevy in the same environment. The `bevy` command line tool
+delegates to this editor for any project specific actions. It also has the side
```
alice-i-cecile, line 135: I think this last sentence would be clearer if split out into its own paragraph.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+benefit of allowing you to more heavily tune the editor for your specific game
+while still having access to the full suite of tools Bevy provides.
+
+The final directories are for version control management and game publishing.
```
alice-i-cecile, line 139: We should just name these directly. Easier to follow, and very useful for beginners.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+```
+
+Likewise, you can remove the asset with:
+
+```
+bevy assets remove <asset-name>
+```
+
+In general, assets come in three varieties: *code*, *data*, and *template*. Code
+assets are little more than Cargo crates, many can even be installed with the
```
alice-i-cecile, line 158:
```suggestion
assets are little more than Cargo crates, and many can even be installed with the
```

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+`cargo` command.
```
alice-i-cecile, line 159:
```suggestion
`cargo add` command.
```

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+
+Data assets are overlaid onto the existing project, and usually go into the
+`assets` directory. Frequently these assets are large, should not be added to
+the repository directly, or should be loaded at build time from an external
+source. Therefore, in addition to copying the assets into the project, data
+assets can also be references to an asset at an external location. **Note that
```
alice-i-cecile, line 165: I don't understand this last sentence.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+bevy assets remove <asset-name>
+```
+
+In general, assets come in three varieties: *code*, *data*, and *template*. Code
+assets are little more than Cargo crates, many can even be installed with the
+`cargo` command.
+
+Data assets are overlaid onto the existing project, and usually go into the
+`assets` directory. Frequently these assets are large, should not be added to
+the repository directly, or should be loaded at build time from an external
```
alice-i-cecile, line 163: I think we should consider just setting up git LFS by default?

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+
+### Publishing Tools
+
+Bevy also helps you publish your games, either by building distributables, or
+even having CI distribute builds each time its pushed.
+
+Use `bevy build` to build your game:
+
+```
+bevy build <target>
```
alice-i-cecile, line 234: I think `target` should be optional: the default should build for Mac / Linux / Windows. And then allow configuration of default build targets.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+
+### Scene and Entity Manipulator
+
+Bevy comes with an entity and scene manipulator. The easiest way to use these
+tools is with the graphical editor, run `bevy editor` in the project directory.
+
+The entity manipulator lists all of the entities in a given world, as well as
+the components that are associated with those entities. You can add new entities
+from this view, and attach new components to them. You can also edit the
+entities that were added using the manipulator.
```
alice-i-cecile, line 262: Only the ones added using the manipulator?

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+
+You can view all entities in your world, not just ones created with the
+manipulator. However, you are only guaranteed to be able to edit static
+entities, and entities created with the manipulator.
```
alice-i-cecile, line 266: I don't understand the motivation here.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+
+In addition to the entity manipulator, Bevy also provides developers with a
+scene manipulator. This allows developers to place entities in a larger level,
+and get a real-time view of how they interact with each other.
+
+### Systems Manager
+
+In addition to the manipulator, Bevy provides developers with a view of all the
+systems their games uses. In addition to creating new systems, developers can
```
alice-i-cecile, line 275: I don't think "add new systems from the editor" makes sense for an MVP.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+use this manager to find and open the code for existing systems. The systems
+manager can also be viewed in the graphical editor, run `bevy editor` in the
+project directory.
+
+### Localization, Accessibility, Porting, and Plugins
+
+Just like the rest of the Bevy engine, the editor can be extended using plugins.
+Some plugins are installed to the system, while others are added only to a
```
alice-i-cecile, line 283: I don't understand this sentence.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+editor:
+
+* Project/System separation
+* Version dispatching
+* CLI Interface
+* Window Interface
+* New project construction
+* Asset management
+* Distributable packaging
+* Credential/secrets management
```
alice-i-cecile, line 304: IMO publishing can be cut from an MVP.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+
+The following is a list of features that are needed for the minimum viable
+editor:
+
+* Project/System separation
+* Version dispatching
+* CLI Interface
+* Window Interface
+* New project construction
+* Asset management
```
alice-i-cecile, line 302: I don't think this is required for an MVP.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+* Distributable packaging
+* Credential/secrets management
+* Scene/Entity manipulator (with possible component editing dispatch).
+* Systems list
+* Plugin architecture
```
alice-i-cecile, line 307: It's not clear what you mean here.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+* Project/System separation
+* Version dispatching
+* CLI Interface
+* Window Interface
+* New project construction
+* Asset management
+* Distributable packaging
+* Credential/secrets management
+* Scene/Entity manipulator (with possible component editing dispatch).
+* Systems list
```
alice-i-cecile, line 306: Wildly useful, but I'm not sure it's particularly required for an MVP. I don't think it's likely to impact architecture.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+
+* Project/System separation
+* Version dispatching
+* CLI Interface
+* Window Interface
+* New project construction
+* Asset management
+* Distributable packaging
+* Credential/secrets management
+* Scene/Entity manipulator (with possible component editing dispatch).
```
alice-i-cecile, line 305: I don't understand the parenthetical.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+While it is possible to make a prototype with a subset of these features, each
+feature implementation must (1) makes space for every other feature, and can (2)
+assumes the other features exist.
+
+The rest of this section will be a sketch out the rough implementation for each
+of these features.
+
+### Project/System separation
+
+1. Create two crates: `bevy_editor` and `bevy_editor_command`, the first will be
```
alice-i-cecile, line 318: Bikeshed: `bevy_editor_command` doesn't clearly explain what the crate does to me.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+   rely on the editor, developers that _do_ want the deliverable to rely on the editor can call it themselves.
+
+### CLI Interface
+
+1. Simple command line argument parsing can be done with clap.
+2. Simple error handling can be provided with anyhow.
+3. There does not seem to be a standard pretty-output library for rust,
+   prettytable seems to be a start.
+   1. Options like crossterm might also fit the bill, but also seem like overkill.
+4. The manipulator/live viewing components can delegate to the window interface.
```
alice-i-cecile, line 354:
```suggestion
1. Simple command line argument parsing can be done with `clap`.
2. Simple error handling can be provided with `anyhow`.
3. There does not seem to be a standard pretty-output library for rust,
   `prettytable` seems to be a start.
   1. Options like `crossterm` might also fit the bill, but also seem like overkill.
4. The manipulator/live viewing components can delegate to the window interface.
```

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+
+1. Extend from the base crate's world.
+1. Ideally, everything should be do-able through the graphical interface, but
+   some things can be pushed to later releases.
+2. The `bevy_editor_command` window only needs to handle: (1)Credential Storage,
+   (2) Template Downloads, and (3) Project Selection, everything else is handled
+   by the a separate `bevy_editor` window.
+3. The `bevy_editor_command` window must seamlessly transition to the
+   `bevy_editor` window, as if opening a project.
+4. The ui should be made using bevy components.
```
alice-i-cecile, line 366:
```suggestion
4. The editor's UI should be made using `bevy_ui`.
```

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+* Plugin architecture
+
+While it is possible to make a prototype with a subset of these features, each
+feature implementation must (1) makes space for every other feature, and can (2)
+assumes the other features exist.
+
+The rest of this section will be a sketch out the rough implementation for each
+of these features.
+
+### Project/System separation
```
alice-i-cecile, line 316: We should probably mention that "these crates live in the bevyengine/bevy repo" somewhere in this section.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+3. Build template object mapping template languages to their result.
+    1. For example foo.rs.tera should become foo.rs.
+4. Fill in appropriate template fields from the user's environment.
+5. Add CI workflows.
+6. Add licenses.
+7. Initialize as Git repo.
+
+### Asset management
+
+1. Template management can be done using terra.
```
alice-i-cecile, line 381:
```suggestion
1. Template management can be done using `terra`.
```

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+2. Downloading static assets can be managed by reqwest.
```
alice-i-cecile, line 382:
```suggestion
2. Downloading static assets can be managed by` reqwest`.
```

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+3. Downloading code assets can be done with cargo.
```
alice-i-cecile, line 383:
```suggestion
3. Downloading code assets can be done with `cargo`.
```

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+4. Downloading template assets can be done with libgit2.
```
alice-i-cecile, line 384:
```suggestion
4. Downloading template assets can be done with `libgit2`.
```

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+5. Template assets should use `directories` to store system wide.
```
alice-i-cecile, line 385: Not fully clear what this sentence means.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+2. Let developers create per-target custom build scripts.
+3. Should ideally support WASM out of the box.
+
+### Credential/secrets management
+
+1. Should use the operating system's keychain/credential storage for storing
+   secrets.
+2. For now, only for uploading to itch.io in CI builds.
+
+### Scene/Entity manipulator (with possible component editing dispatch).
```
alice-i-cecile, line 399: Paranthetical is unclear.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+
+1. Can use rfc 62 as its data model.
```
alice-i-cecile, line 401: Definitely needs a link.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+2. Should provide an editor for adding and removing components to entities.
+3. Should also contain a basic editor for setting the initial values of an
+   entity's components.
+   1. As a bonus, plugins can allow for domain-specific editor for given
+      component types.
+4. The scene viewer should place all objects in a scene at their initial
+   position.
+5. A play button lets users run any given scene with the world's systems.
```
alice-i-cecile, line 409: This is a major feature: I would like to see this split out into its own section.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+This is a lot of things for a single RFC. It would be possible to make a minimum
+editor with fewer features, and would lead to an editor quicker.
+
+However, the size of this RFC is merited for these reasons:
+
+1. Most of these features are already required for making the editor usable.
+2. Implementing most of these features can go faster when it assumes the other
+   features are build along side it. For example, the metadata used for the
+   entity manipulator can also be used for systems browser.
+3. Making these features as plug-ins helps us make sure the plugin system is
```
alice-i-cecile, line 439: Generally speaking, there's some ambiguity between "plugins in the sense of the `Plugin` trait" and "plugins in the sense of editor extensions". Unless these are intended to be strictly equivalent (which should be clarified), we should pick a different name.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+   appropriately powerful.
+
+## Rationale and alternatives
```
alice-i-cecile, line 442: I think that each of these questions should get their own subsection with FAQ style headings. Currently, they're a bit too terse to properly understand.

alice-i-cecile, line 442: "Why should we use bevy_ui" is an important rationale

afonsolage (commented):
I did a first pass and overall it seems to be focusing more on Bevy Editor command line aspect than Bevy Editor itself.

While I agree with most of the features here, it still require users to know and feel comfortable with command line tools (which most designers aren't).

So this RFC should focus on bevy command line tool or expand and give more details on visual editor itself, which can become pretty huge and conflict with others non-approved RFCs

rfcs/72-editor.md
jnhyatt: I agree we shouldn't add license files in the template. I can see this turning into complaints when commits accidentally include unwanted licenses.

Weasy666: If you make the project creation a bit more interactive, then you could ask the user which kind of license, if any, should be added.
A bit like Rustup asks you which Rust version and modules should be installed.

rfcs/72-editor.md
Weasy666: Maybe [gitoxide](https://github.com/Byron/gitoxide) is a better fit for this, one less wrapped C library. Makes builds more predictable and stable on different platforms.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+    target/
+        ...
+    editor/
+        Cargo.toml
+        src/
+            main.rs
+    .git/
+        ...
+    .github/
+        workflows/
```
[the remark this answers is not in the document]
Weasy666: We could make that part of one of the templates. Then user can get a default workflow for their CI, if a template for that exists.

rfcs/72-editor.md
LeifAndersen: Mmm...perhaps this is a bit unclear on my part. What I imagined was that licensing is separate from the template, but still provided as a (changeable) default. The process for new project construction that I'm imagining is:

1. Fill out template.
2. Add chosen license
3. Add version control
4. Add CI template

This is because the license you choose to release your project under is (imo, usually) a separate concern from which template you use to start your project. And since the cross product of all licenses and templates makes sense, they should be seperate flags.

So something like:

```
bevy new -l mit -l gplv2 -t cardgame my-project
```

I do think we should by default add a license (unless the user explicitly passes something like `-l nolicense`), even if its an 'all rights reserved' statement.

rfcs/72-editor.md
LeifAndersen: How about:

Instead, it provides the minimum set of features required for the Bevy editor, and discusses basic ways these features compose. The full composition of these features, however, is left for a future RFC.

?

rfcs/72-editor.md
LeifAndersen: oops, those used to be multi-part clauses, I can change them to be independent sentences.

rfcs/72-editor.md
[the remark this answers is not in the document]
LeifAndersen: See: https://github.com/bevyengine/rfcs/pull/72#discussion_r1390488808

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+flappers/
+    README.md
+    Cargo.toml
+    LICENSE-Apache2.txt
+    assets/
+        ...
+    src/
+        main.rs
+    tests/
+        ...
```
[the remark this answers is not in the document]
LeifAndersen: Shrug, this seems like a thing that is decided by the template, not this tool. Although we can debate about if it should be in the default template, I could go either way.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
+# Feature Name: `editor-foundation`
```
cart, line 1: First: I'm stoked that you're thinking about these things and there are a lot of good ideas here!

As a general comment, I think this RFC bites off more complexity than it should, in too many disparate areas, in not enough detail. So many of these things will be dictated by details that haven't yet been established. It feels like putting the cart before the horse. I think each of these categories deserves one or more RFCs, and each RFC should cover more details and more "design thoughts / justifications". I think the path forward is picking one of these areas that interests you, diving deep, and driving the scenario to completion. Another path forward is to omit any prescriptive details and we have an RFC whose purpose is just "agreeing on a small core list of things that need to be done for an editor MVP", but thats not really our design bottleneck right now. We need to sort out what specific scenarios look like.

1. **Default Project Structure**: this will largely be determined by the editor data model (which has not yet been established). If we roll with hot-reloading / dynamic linking, that will require a very different project structure than statically referencing. This deserves its own RFC, and should be covered only after contending with these details.
2. **Bevy Assets**:  This definitely deserves to be broken into one or more RFC design that (together) cover in detail the end-to-end "asset repository" scenario.  What will a repository "be" functionally (what protocol will it speak)? What will the manifest format look like and how will it be consumed (aka what _is_ an asset)? "Code assets" will need to play into the **Default Project Structure** scenario (either cover this after (1), or make sure it covers the range of possible paths).  Where will "downloaded code asset" code go? How will it be built? Do we distinguish between editor plugins, app plugins, and libraries? Is adding assets repeatable? Are users allowed to edit them? Are they stored adjacent to user assets (also plays into (1))? 
3. **Bevy Build**: Introducing a separate build command has massive implications for how people develop games in Bevy (lots of pros and lots of cons). This deserves way more than a couple of lines, both to describe what it does internally, why it does it that way, and plenty of rationale justifying its existence relative to `cargo`. One of the justifications for it is "bundling with all needed assets" (which I do agree with). But there are a ton of details there that have not yet been established. Will this be used for "dev" scenarios or just "final deployments"? If it is used for dev scenarios, will it try to avoid "re-doing" work (ex: copying expensive folders, rebuilding assets). Bevy Asset V2 doesn't yet have a CLI (but it will). How will this command relate to it? Do we need both apis? Why? One major justification for a separate build command is automatic cargo feature management. That deserves coverage and mentioning.
4. **Scene and Entity Manipulator**: We need something shaped roughly like this, but this deserves way more coverage than it is given before we can decide on what these systems are and how they relate to each other (ex: what is editable and when).

LeifAndersen: > Another path forward is to omit any prescriptive details and we have an RFC whose purpose is just "agreeing on a small core list of things that need to be done for an editor MVP",

That's much closer to what I was going for.

> but thats not really our design bottleneck right now. We need to sort out what specific scenarios look like.

Ah, I see.

By scenarios, do you mean user stories?

cart: > That's much closer to what I was going for.

I'm happy to roll with this path, but I'm relatively certain that as soon as we eliminate the "design details" we will trivially be in consensus. Ex: "yes we definitely want a bevy cli, an asset repo, a visual editor with a scene and entity inspector, a standard project structure, and plugin support". I think our time is probably better spent discussing what each of these things looks like in detail.

> By scenarios, do you mean user stories?

Kind of. From my perspective, "scenario" is a nebulous term that applies to either user stories or "feature areas". In this case, I used "scenario" to mean "specific implementable feature areas" like "bevy asset repos", "default project structure", "bevy build", "scene and entity manipulator", etc. "User stories" can cross cut through many of these "scenarios".

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+A secondary, smaller challenge, is that Bevy relies on the Cargo build system
+for modularity. An interactive-editor that is core to Bevy must both respect
+this design, and work with the limitations and abilities of Cargo.
+
+## User-facing explanation
+
+The easiest way to get started with Bevy's editor by downloading it with Cargo:
+
+```
+$ cargo install bevy
```
devil-ira, line 45: Using `bevy` here is not possible because the `bevy` crate already exists.
```suggestion
$ cargo install bevy_cli
```
The actual binary being named `bevy` is good though :)

rfcs/72-editor.md
devil-ira: It should definitely not add any license by default. That is unacceptable.

Licensing a Bevy project is niche enough that I'd leave it out of `bevy new` entirely.
`cargo new` doesn't provide this functionality either.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+    Cargo.toml
+    LICENSE-Apache2.txt
+    assets/
+        ...
+    src/
+        main.rs
+    tests/
+        ...
+    target/
+        ...
```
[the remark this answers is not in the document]
alice-i-cecile: Yeah, I think we can cut this.

rfcs/72-editor.md
LeifAndersen: Cargo doesn't, but these do:

1. Npm (Javascript) does.
2. Raco (Racket) does.
3. Leiningen (Clojure) does.
4. Bevy's existing template on github

Although notably these don't:
1. Sbt (Scala)
2. Pkg (Julia)
3. Godot

Another option is to create a LICENSE file and put text in it saying `<INSERT LICENSE>` as a default.

If we don't want to pick any particular license as a default, we can also just put something like `<Insert License Here>` as a default.

afonsolage: I would start without license and later on add a option to include license, because I still feel like the user should opt in for specifying the license, not the tool.

Also the vast majority of Bevy users will make closed source games, which normally is copyright, so no license by default fits the majority of use cases.

Godot, Unity and Unreal doesn't do that, and they should be our benchmark, not package managers, since Bevy is aiming to be a game editor.

rfcs/72-editor.md
LeifAndersen: Crates.io seems to let you have a binary and library crate have the same name. For example I just tested:

https://crates.io/crates/leiftestlibbinsplit

You can run `cargo install leiftestlibbinsplit` and run the executable and it runs just fine. And you can also make a new project and run `cargo add leiftestlibbinsplit`, and now you can `use leiftestlibbinsplit::*` in your project.

Am I missing something?

rfcs/72-editor.md
LeifAndersen:
```suggestion
*Note the use of `install`, not `add`!* This means Bevy's editor for your entire rust environment. In the future you may also be able to download it from external sources such as Steam.
```

rfcs/72-editor.md
eidloi: I think the point here should be to help new developers visualize the running content, something that bevy_inspector_egui already does well for instance

MiniaczQ: I think the target here should be not the coders, but art designers.
Allowing them to preview, and in the future edit the graphics in an interactive way is the way.

rfcs/72-editor.md
MiniaczQ: This should be what `bevy_inspector_egui` does currently, so exploration, filtering and editing component values.
By "dispatch" possibly suggesting a trait (object) with API for exposing the mutable values.
Does Bevy expose all the necessary interfaces for that or are they defined in the plugin crate?
Defining an API in Bevy would be a nice first step.

rfcs/72-editor.md
MiniaczQ: Would this require defining how states and scenes are bound together?

rfcs/72-editor.md
MiniaczQ: What about `build.rs` and any sort of preprocessing/baking.
Would this be the command that does it?

As of right now it seems like a light wrapper around `cargo build`.
Worth noting, it should allow all of the arguments `cargo build` takes.
Possibly making default `cargo build` arguments a project configuration.

rfcs/72-editor.md
MiniaczQ: #62 probably related to this

rfcs/72-editor.md
irate-devil: Sorry for the late reply and not being more specific.
It's possible to have a library and binary in the same crate but I don't think that's an option for Bevy. It would require syncing the release of a new version of Bevy's cli to a release of Bevy as a whole. We technically could do this but I figure this will cause us trouble sooner or later.

rfcs/72-editor.md
```diff
@@ -0,0 +1,508 @@
...
+The default template provides a lot more than a standard new cargo project. Much
+of this is not strictly required, but will be wanted for the average project.
+Because this is mostly convention, feel free to change the layout as needed. The
+first three files: `RREADME.md`, `Cargo.toml`, and `LICENSE-Apache2.txt` are
+standard meta-information that are in most crates. The next folder: `assets` is
+the default location to place project non-code assets. The bulk of the source
+code goes in the `src` directory, likewise tests go in `tests`. Project builds
+go into the `target` directory.
+
+The `editor` directory provides a sub-crate for the Bevy editor. Associating
```
djeedai, line 132: I don't understand this entire paragraph, sorry. What does `associating` mean here, how is that related with Bevy versions, and why is there a crate involved? I'm clueless. 🤔

rfcs/72-editor.md
djeedai: In any case I don't think having 2 things with the same name is good. This will create unnecessary confusion.

PR #74, opened 2023-12-27
Sub-states through `States` derive macro

MiniaczQ:
__[RENDERED](https://github.com/MiniaczQ/rfcs/blob/macro-derived-substates/rfcs/74-macro-derived-substates.md)__


MiniaczQ:
This RFC may change significantly or become unimplementable depending on https://github.com/bevyengine/bevy/issues/9130

I never accounted for "forced" schedules here.
In that case [physically separating states](https://github.com/MiniaczQ/bevy-design-patterns/blob/better-states/patterns/instant-states-hierarchy/README.md) seems like a better choice.

MiniaczQ:
Closed in favor of https://github.com/bevyengine/bevy/pull/11426

PR #75, opened 2024-01-01
Fragmented Levels

codercommand:
**[RENDERED](https://github.com/codercommand/bevy_level_rfc/blob/main/rfcs/75-fragmented-levels.md)**

An API and file format for managing levels composed of multiple scenes. This feature would be used by the future bevy editor to provide a new innovate way for multiple people to work in the same level without: merge conflicts, file locking, or requiring a live multiplayer editor.


codercommand:
This is an idea I had a few months ago, and I talked about briefly on the Bevy discord (https://discord.com/channels/691052431525675048/745805740274614303/1191144487825715220)

Decided to make an RFC despite not making a prototype yet because I decided it would be better to get discussion started instead of waiting.

andriyDev:
I love the idea of having people able to work on separate parts of a level easily. Scenes as "layers" of a level is a great idea imo.

One thing I would like is for levels to be more generic. The same way "scene" isn't a scene (a scene can be a level, but also a player character, or an item pickup, or whatever), a "level" doesn't need to be just a level. A concrete example could be a player character. The player character level could be composed of the gameplay scene (colliders, markers, etc), a production art scene (the actual production mesh, armature, etc), and a debug art scene (a capsule or something for development). The same way a game world level can load chunks of its world separately, a player character level can conditionally load the production or debug art scenes.

This means that the `change_level` part of the API should not be part of the main API. If the player itself is a level, it needs to be able to be spawned (and loaded/unloaded) independently. Levels can still easily be made with a simple `change_level` system though, in a similar way to how states work:

```rust
fn change_level_system(mut current_level: ResMut<CurrentLevel>, next_level: Res<NextLevel>) {
  if current_level.level == next_level.level {
    return;
  }
  current_level.unload_all_scenes();
  current_level.level = next_level.level;
  for scene_to_load in next_level.scenes.iter() {
    current_level.level.load_scenes(scene_to_load);
  }
}
```

Note if this is all it takes, it might be better to just have users make this and not provide it from Bevy. Likely a proper level transition will require loading assets and such (so you need a loading screen), so providing a "convenient" level transition may not be worth the maintenance. I'm not confident about that though.

---

If we go with this more generic approach, I would like to not use the word "level". It's a word that heavily implies a world, but if this is extended to allow for players to be levels, then that seems like a confusing word to use ("the player character level" could mean a level that *only* contains the player, or perhaps it means a game world that's owned by the player or something?). I also have similar concerns with the word "scene", but Godot has established "scene" already as not-just-a-world, so I'm more fine with keeping that word.

I don't have a good candidate for another name. Maybe `SceneSet`, but that word seems a little too broad. At least it doesn't make you immediately think of levels, it makes you think of a "collection of scenes" which is all this is really.

codercommand:
If multiple scenes can be loaded as suggested and people can make their own resource handles for their scene, then it gives alternatives for naming.

For the file format name it could be something like `.cos` (Collection of Scenes), or `.sc` (Scene Composition).

Then the trait/struct that has the API for managing `.cos`/`.sc` files could be called something like: `SceneManager` or `SceneComposer`.

If there are multiple levels then `change_scene` probably won't exist because you would load the level via the asset loader, to bring it inline with existing bevy systems, then add it to a resource yourself.

Registering new level handles would look something like this:
```rust
// SceneComposer adds methods for managing the scene.
// SceneComposition is the resource type made from the `.cos`/`.sc` file.
#[derive(Resource, SceneComposer)]
struct GameplayLevel(Handle<SceneComposition>);

// Startup system
fn setup_gameplaylevel(
    mut commands: Commands,
    server: Res<AssetServer>
) {
    let handle: Handle<SceneComposition> = server.load("level/maze.cos");
    commands.insert_resource(GameplayLevel(handle));
}
```

Then to change levels you could do something like this:
```rust
fn change_level(
    events: EventReader<ChangeLevel>,
    mut level1: ResMut<Level1>,
    mut level2: ResMut<level2>,
) {
    for event in events.iter() {
        match event {
            ChangeLevel::Lobby => {
                level2.unload();
                level1.load();
            },
            ChangeLevel::MainLevel => {
                level1.unload();
                level2.load();
            }
        }
    }
}
```

The primary purpose of `Fragmented-Levels` is to provide an API for fragmenting, merging, composing ([level-concepts](https://github.com/codercommand/bevy_level_rfc/blob/main/rfcs/75-fragmented-levels.md#level-concepts)), and managing levels.

Stuff like level changes via state or events should be implemented by whoever is using the API and shouldn't be part of the RFC since it makes the API more complicated than it needs to be, and it's better to encourage people to write code for their use case instead of providing strange defaults. Some helper functions or premade systems that 90% of people would make themselves might be nice to include as optional extras though.

rfcs/75-fragmented-levels.md
```diff
@@ -0,0 +1,141 @@
...
+todo
+
+
+
+## Drawbacks
+
+- It's an extra crate to maintain.
+
+
+## Rationale and alternatives
```
MrGVSV, line 119: Could add to this section what differentiates the strategy proposed in this RFC from something like nested scenes (which, to my knowledge, is a [planned feature](https://github.com/bevyengine/bevy/discussions/9538))? It feels like a level could also be created as a scene of scenes.

tbillington:
This sounds quite similar to additive scenes in Unity. I've used them like you describe and heard the same from others too.

Though your concept differs with layers being restricted to being used in the 1 "parent" scene, and with some UX niceties.

Have you tried them in Unity? It's definitely useful prior art.

rfcs/75-fragmented-levels.md
codercommand: I read about that a week or two ago and completely forgot where I had heard about scene nesting. Then for some dumb reason assumed it was a UI only thing.

I'll amend the RFC at some point to use nested scenes to replace level files. Using nested scenes is also an improvement since you can have multiple nested scenes with layers without making the API more complicated because multiple nested scenes are planned on being supported already.

codercommand:
> This sounds quite similar to additive scenes in Unity. I've used them like you describe and heard the same from others too.
> 
> Though your concept differs with layers being restricted to being used in the 1 "parent" scene, and with some UX niceties.
> 
> Have you tried them in Unity? It's definitely useful prior art.

Used Unity for almost a year in my first year of University about 2 years ago, haven't used it since. Never learnt about additive scenes. Would probably be good to put in prior art I agree. I should probably also do the same for Godot and Unreal since they have similar.

In Godot you can do nested scenes, and Unreal you can do have streamed levels or make blueprints out of levels.

My idea essentially takes the nested scenes idea and expands it with functionality that improves it. This RFC is approaching it from the API perspective so it's easy to implement in the bevy editor, and proposing the concept. A 2nd RFC detailing exactly how it looks and you interact with it might be needed or not, don't know yet.

As for limited to the parent scene, that'll probably change. See the comment above yours.

Yes, mine has some UX niceties, but I believe to make it easier to implement it should be supported from the start and will require a specific API - I could be wrong, maybe no specific API is needed.

-----------------------------------------------------------------------

What makes my idea different is that when you traditionally edit a nested scene or additive level, it's on it own in a separate scene and you can't see the larger scene it's used in. Instead with this is when you edit a scene (formally level), the nested scenes with a component of `Layer` would all be treated the same way layers are treated in Photoshop or paint.net. This means you can edit some part of the map with a semi transparent hologram overlapping your work, and make certain scenes/layers invisible and other visible.

The APIs that are needed to be made for the editor, are also useful for stuff like chunk loading, level loading, or doing something like multiple dimension levels for gameplay, so it's considered a side benefit for people writing game code and should be exposed as a separate module/crate, not editor exclusive.

------------------------------------------------------------------------------------------------------------

If you need to create delegates, signals, or other types of events between scenes in something like Godot the intuitive way requires editing a scene normally. This means if you are making a level and want to test it, you can't add any functionality you need until whoever is working on the level is finished or you possibly resolve a merge conflict.

From the way I've seen Bevy work, this isn't an issue. You could make a new scene with event dependencies, and other type based dependencies, and it would work without having to edit another level.

Because Bevy doesn't have other dependencies in scenes besides other scenes, it makes implementing the concept of layers for the bevy editor easier than other engines in my opinion.

rfcs/75-fragmented-levels.md
codercommand: Just realized one side affect of used the new bevy scenes file instead of the level file, it will make the new scene API and file format more complicated.

When you load a level, you don't want to always load all the attached scenes. If I'm not mistaken the current plan is for a scene when opened is to load itself and all nested scenes recursively. The level API doesn't do this. Instead it gives a handle, and you specify which scenes you want loaded along with their visibility and/or transparency.

If this uses the new bevy scene, then changes may need to be made to how bevy scene is intended to work. Which would make this an extension RFC for bevy scenes.

Personally I think keeping `Fragmented-Level` and bevy scenes disconnect is better than making it a feature of bevy scenes. But not using scenes since they support nesting seems like a shame, so I'm not sure which is better. Want some more opinions on this.

codercommand:
Huh, never seen that before. Definitely useful prior art.

Sounds like what I'm prosing has been done before (which I didn't know), but I'm changing some of the terminology, and UI.

------------------------------------------------------------------------------------------------

Now I know this, the question is does the bevy editor and scene format already plan on doing this? If so, this RFC is irrelevant. If not, then does bevy want to have a feature like this? If so, then this RFC needs more iteration to figure out how bevy should support this concept which unity already has called additive scenes.

Personally I prefer changing the terminology and thinking of how to do scene composition in a way that makes sense for Bevy. I'm not a fan of copying another engine directly, but instead taking the concept and changing parts where needed to make suitable and intuitive in the new context (this case bevy).

codercommand:
To keep everyone up to date, there was further discussion with Alice on [discord](https://discord.com/channels/691052431525675048/745805740274614303/1191799929501401118).

In summary the RFC is being closed and split in two:
- The concept of levels is better suited to large open worlds/level streaming, and would be a 3rd party crate, that may or may not be unstreamed into bevy itself.
- The idea of layers and being able to edit different layers in a scene kind of already exists in the new bevy scene format. What's more important is the UI, UX, and promotion in bevy learning materials on how to use layers/nested scenes to improve their workflow. An issue for this will be made on the main bevy repo at some point.

I will also be retaining the `bevy_level` crate name for now.

PR #77, opened 2024-03-19
Dev tools abstraction

alice-i-cecile:
> Dev tools are an essential but poorly-defined part of the game development experience. Designed to allow developers (but generally not end users) to quickly inspect and manipulate the game world, they accelerate debugging and testing by allowing common operations to be done at runtime, without having to write and then delete code. To support a robust ecosystem of dev tools, `bevy_dev_tools` comes with two core traits: `ModalDevTool` (for tools that can be toggled) and `DevCommand` (for operations with an immediate effect on the game world). These are registered in the `DevToolsRegistry` via methods on `App`, allowing tool boxes (such as a dev console) to easily identify the available options and interact with them without the need for ad-hoc user glue code.

Co-authored with @matiqo15 and @rewin123.

Read it [**RENDERED**](https://github.com/alice-i-cecile/rfcs/blob/dev-tools/rfcs/77-dev-tools-abstraction.md) here.


rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,572 @@
...
+
+- unidiomatic: these are currently configured via stand-alone resources, and changing that would require either synchronization or the imposition of this pattern on every tool that could be used as a dev tool
+- breaks the granularity of change detection: we can't determine *which* dev tool's config has changed
+- accessing information about the current config of a dev tool requires an added layer of indirection, and will simply fail if the dev tool was not registered, even in cases where no toolbox is being used
+
+While this makes our internals more complex, that's a trade-off we're willing to accept.
+
+### Why not use trait queries?
+
+If we need to access a set of objects, all of which implement the same ttrait, why not simple use [trait queries](https://github.com/bevyengine/rfcs/pull/39).
```
pablo-lua, line 524:
```suggestion
If we need to access a set of objects, all of which implement the same trait, why not simple use [trait queries](https://github.com/bevyengine/rfcs/pull/39).
```

pablo-lua (commented):
Looks good! Really liked this RFC, just a couple comments with things that I noticed here and there.

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,572 @@
...
+To model this, we leverage Bevy's existing `Command` trait, which exists to perform complex one-off modifications to the world.
+
+```rust
+/// Dev commands are used by developers to modify the `World` in order to easily debug and test their application.
+/// 
+/// Dev commands can be called with arguments to specify the exact behavior: if you are creating a toolbox, parse the provided arguments 
+/// to construct an instance of the type that implements this type, and then send it as a `Command` to execute it.
+/// 
+/// The documentation on this struct is reflected, and can be read by toolboxes to provide help text to users.
+trait DevCommand: Command + Reflect + FromReflect + Debug + 'static {
```
bushrat011899, line 225: Instead of the `DevCommand` trait including a `parse_from_str` method, it would make more sense to require `DevCommand: FromStr`.

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,572 @@
...
+
+    /// Enables this dev tool.
+    fn enable(&mut self) {
+        self.toggle(true);
+    }
+
+    /// Disables this dev tool.
+    fn disable(&mut self) {
+        self.toggle(false);
+    }
```
pablo-lua, line 132:
```suggestion
    fn enable(&mut self) {
        self.set_enabled(true);
    }

    /// Disables this dev tool.
    fn disable(&mut self) {
        self.set_enabled(false);
    }
```

pablo-lua: misunderstood in early comment, now I think its okay

pablo-lua (commented):
Noticed a error in my previous comment, sending again

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,572 @@
...
+        }
+    })
+}
+```
+
+Finally, we want to be able to pass in a user supplied string, parse it into a dev command and then evaluate it on the world.
+
+```rust
+#[derive(Event)]
+struct DevCommandInput(String);
```
makspll, line 366: This doesn't seem to be used in this example:
```suggestion
```
I do wonder what the motivation behind adding 
```rust
    fn parse_from_str(string: &str) -> Result<Self, DevToolParseError> {
        Err(DevToolParseError::NotImplemented)
    }
```
to the modal dev tool interface is

bushrat011899 (commented):
This is an interesting problem to try and solve. I'm not fully sold on the current distinction between commands and modals, and feel like it needs a bit more refinement. To me, the biggest difference between the two is one has *state*, while the other is *stateless*. Taking the `no-clip` camera as an example, that is definitely stateful, but you still want commands to interact with it (e.g., `/noclip speed 10`, `/noclip on`, etc.).

I think it would be useful to think of the `ModalDevTool` trait as describing a tool state which can be toggled (omitting the `parse_from_str` entirely). This lends itself well to adding extra traits for visualising these tool states (e.g., in a log, in a toolbox, etc.). Rich interactions with the dev tool states would then be the job of dev commands.

So for a no-clip camera dev tool, you'd define the `ModalDevTool` as the state of the tool, and a `DevCommand` for adjusting parameters. It would make sense for a default `DevCommand` for toggling an arbitrary `ModalDevTool` to be provided (e.g. `/toggle noclip true` ).

makspll:
This is really cool! A few things:
### parse_from_str
 first of all why have `parse_from_str` on modals at all? For commands this is clear, a command has a one-to-one relationship with *some* string representation, for a modal this is less clear cut, and a parse_from_str feels more like deserialisation. If anything, instead of returning a Self instance, modals could have a `react_to_str` method but this would step on DevCommand territory.

Also `parse_from_str` seems to be very un-opinionted, which may be on purpose, but I can see different standards arising, i.e. do we include the dev tool "executable" name as first argument, string separators, how do we interpret quoted strings (i.e. does `"arg"` ==  `arg`), I don't think we necessarily should depend on clap directly, but we could have some more abstractions here, I am thinking something along the lines of a smaller `https://docs.rs/clap/latest/clap/trait.FromArgMatches.html`

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,591 @@
...
+To model this, we leverage Bevy's existing `Command` trait, which exists to perform complex one-off modifications to the world.
+
+```rust
+/// Dev commands are used by developers to modify the `World` in order to easily debug and test their application.
+/// 
+/// Dev commands can be called with arguments to specify the exact behavior: if you are creating a toolbox, parse the provided arguments 
+/// to construct an instance of the type that implements this type, and then send it as a `Command` to execute it.
+/// 
+/// The documentation on this struct is reflected, and can be read by toolboxes to provide help text to users.
+trait DevCommand: Command + Reflect + FromReflect + FromStr<Error=DevToolParseError> + Debug + 'static {
```
pablo-lua, line 219:
```suggestion
trait DevCommand: Command + Reflect + FromReflect + FromStr<Err=DevToolParseError> + Debug + 'static {
```

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,591 @@
...
+
+```rust
+/// Modal dev tools are used by developers to inspect their application in a toggleable way,
+/// such as an FPS meter or a fly camera.
+/// 
+/// Their configuration is stored as a resource (the type that this trait is implemented for),
+/// and they can be enabled, disabled and reconfigured at runtime.
+/// 
+/// The documentation on this struct is reflected, and can be read by toolboxes to provide help text to users.
+trait ModalDevTool: Resource + Reflect + FromReflect + FromStr<DevToolParseError> + Debug {
```
pablo-lua, line 106:
```suggestion
trait ModalDevTool: Resource + Reflect + FromReflect + FromStr<Err=DevToolParseError> + Debug {
```

alice-i-cecile:
> First of all why have parse_from_str on modals at all

I've expanded the toolkit example to try and clarify this: you need it to be able to allow for free-form runtime config of these resources. Obviously you *could* make dev commands for each way you might want to configure the resource, but I was thinking it would be more user-friendly to allow them to just set the struct's values directly.

> Also parse_from_str seems to be very un-opinionted, which may be on purpose, but I can see different standards arising

Yeah, I'm not sure of the best way to handle this. I'd like to interoperate with `std`'s `FromStr` trait, but it really doesn't impose any structure at all.

> however if the modal has 20 properties, your console might as well support multi line json haha :P

Yeah, I actually really want to support parsing these objects from the new `bsn` format at some point!

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,634 @@
...
+    ///
+    /// For a typed equivalent, simply use the `FromStr` trait that this method relies on directly.
+    fn parse_and_insert_tool(&self, world: &mut World, s: &str) -> Result<(), DevToolParseError>{
+        // Parse out the name
+        let name = s.clone().split_whitespace.next()?;
+        
+        // Get the associated `ComponentId`, so we can use it to insert a resource of a dynamic type
+        let component_id = self.lookup_tool_component_id(name);
+        // Look-up the existing resource to get access to get access to the metadata we need
+        let tool_metadata: &dyn DevTool = self.get_tool_metadata(component_d)?;
```
pablo-lua, line 544:
```suggestion
        let tool_metadata: &dyn DevTool = self.get_tool_metadata(component_id)?;   
```

rewin123:
The RFC looks really good! I really like the good analysis done on usage and what should be classified as dev tools and will be used and what will not. From this analysis comes for me a reasonable and good trait structure. 

However, I think the RFC did not address one important issue in my opinion.

Part of the dev tools may want to display debugging information on the screen or get some input from the developer. For example, the fps overlay displays the number of fps in the bottom right corner, and the entity inspector may want to show some editable fields and buttons near the right edge of the screen. Or the realtime statistics render graph will want to display rendering times information somewhere in the static area. And if multiple dev_tools claim the same area of the screen to display, it will be a mess. And obviously dev_tools can and will conflict over displaying information. Accordingly, the question not solved and not asked in my opinion in RFC is "how is it planned to organise the location of the displayed information of dev modules?". (Or how do we start building editor UI from dev tools xDDDD, sorry).

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,634 @@
...
+
+One method on `DevToolsRegistry` is worth special attention: `parse_and_insert_tool`.
+This tool takes a common pattern, parsing the configuration for a dev tool from a provided string, and provides a user friendly, safe API over it.
+
+```rust
+impl DevToolsRegistry {
+    /// Parses the given str `s` into a modal dev tool corresponding to its name if possible.
+    ///
+    /// For a typed equivalent, simply use the `FromStr` trait that this method relies on directly.
+    fn parse_and_insert_tool(&self, world: &mut World, s: &str) -> Result<(), DevToolParseError>{
```
pablo-lua, line 537: First: What are we parsing here? I was having a look and ModalDevTool trait doesn't have a metadata at all. If we expect to not only parse commands here, but also tools themselves, we can store in the Metadata some kind of identifier to it: Maybe a enum of some kind?

I Think we need to clarify how are we going to register commands so this can be an easy way forward. 
For example, lets say we store in the metadata an enum 
```rust
enum DevToolType {
   ModalDevTool,
   DevCommand
}
```
Now, something that maybe can be strange: To do everything the way we currently do with strings, we must ensure that all the tools have different names.
And so, we get the tool metadata, with the enum stored, and we can make a command or insert the tool, as simple as possible.

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,634 @@
...
+#[derive(Resource)]
+struct DevToolsRegistry {
+    /// The stored collection of modal dev tools, tracked in a type-erased way using [`ComponentId`]
+    /// 
+    /// The key is the `name()` provided by the `ModalDevTool` trait.
+    modal_dev_tools: HashMap<String, ToolMetaData>,
+    /// The metadata for all registered dev commands.
+    /// 
+    /// The key is the `name()` provided by the `DevCommand` trait.
+    dev_commands: HashMap<String, DevCommandMetaData>
```
pablo-lua, line 452: For starters, can we unify this two metadatas into a single metadata?

alice-i-cecile:
@rewin123 great question! I've added more thoughts on this to both "Rationale" and "Future work" justifying why it's not included currently: https://github.com/bevyengine/rfcs/pull/77/commits/45506d2e923e6ce261790e2f1c69476b6c158c87

rfcs/77-dev-tools-abstraction.md
pablo-lua: Lets say we return Metadata here, instead of &dyn DevTool. If we go with the enum approach, we can match arms and one or another, depending on the tool type.

pablo-lua (commented):
The DevToolsRegistry was a bit confusing, but I tried my best to organize my thoughts here

codercommand:
On the topic of commands, personally I've come to really like menus like F1 on VScode and ctrl+k on Github.

Personally I like the idea of commands being a feature that has ui inspired from Githubs ctrl+k menu
image

Then as for features it works like traditional console and have take commands, but it can also provide context. This allows it to provide context relevant info like the file explorer right click context menu or you can search through sub commands/directories.

I'm not sure on the best way it should work. Since it's a hybrid between a console and a context menu, maybe have the user right click and if they have something selected in the scene it can be used by the command.

Imagine your workflow looking something like click on thing, right click to get menu, type command or optionally click on a proposed option. And if you click on an option, it could open more suboptions, etc.

rfcs/77-dev-tools-abstraction.md
alice-i-cecile: For now, maybe. Down the line, I think their implementations will diverge, so I'd like to keep them separate and eat the code duplication.

alice-i-cecile:
Great suggestion: I love command palettes too, and I've added them as an interesting example of a potential toolbox to consider when designing.

rfcs/77-dev-tools-abstraction.md
alice-i-cecile: > First: What are we parsing here? I was having a look and ModalDevTool trait doesn't have a metadata at all

This is an omission! Lemme fix that...

codercommand:
Something else to propose. I've noticed some engines like Godot and Defold don't support multi monitor support.

Since this system will need to define what takes screen space and bevy already comes with multi window support. Maybe make it that there is a tools window and game window?

This would allow for x2 monitor workflows.

You'd have grid slots on the game/main window that can be enabled and disabled, and you can choose what displays where the same as nividias hud layout thing.
image

Then the second window would be used by other tools that require more screen space. These ones would take the entire window except for a strip at the top which would be a tab browser. This would allow for full window tools, and you can click between their tabs to swap between them like in a browser.

With this setup you now have dual monitor support.

If you wanted single monitor support, you can make the game window a tab. Now it shares the same window as the other tools.

codercommand:
I can imagine a plugin providing a feature like scene editing or animation graph editing.

You'd add the crate to the toml, add the plugin, and via the APIs it would add stuff to the command palette, and it would create a tab in the 2nd window that it uses to display it's stuff.

Something like this would mean the editor runs in game runtime. This makes it very easy for the editor to interact with the game world with little to no new APIs needed.

alice-i-cecile:
Yep, definitely an interesting topic to explore (and I'd like multimonitor support too!). I think that's more in the realm of a potential toolkit, rather than the MVP abstraction however.

codercommand:
Thought it may cover the abstraction since the dev tool might need to specify if it uses a tab or grid slot on the game window.

naomijub:
1. I have the same concern as @rewin123, being someone that has dedicated my time to develop space editor UI, I feel that this solution is great and quite versatile, but doesn't address my main area of work **"How this dev tools will be represented in the ui?"**. Besides what Rewin already elaborated on shared ui locality, I find that accessing dev_tool information and manipulating it in runtime is quite important, but it seems the focus was parsing and enabling, while I would like something more potent on the Reflection side, like hooks to events, systems, etc. 

2. I also feel that storing keys as String is quite bothersome, specially for debugging purposes (`HashMap<String, ...>` ), this because we cannot easily preview or understand what are the contents, on the other hand, if we stored as `TypeId`, or something similar, the amount of options we can consider is vastly reduces. For example, String could be anything, but if we have a well organized project, `TypeId`, would be limited to modules named `dev_tools`  or `settings`.

3. I feels that modals are equivalent to Resources and Command to Events, which is basically how we have implemented dev tools in space editor. I kind of expect a bit more from this kind of thing than what we already have with a tweak., but I dont have any constructive suggestions here.

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,658 @@
...
+            self.disable();
+        } else {
+            self.enable();
+        }
+    }
+}
+```
+
+Modal dev tools are registered via `app.init_modal_dev_tool::<D>()` (to use default config based on the `FromWorld` implementation) or via `app.insert_modal_dev_tool(config: D)`.
+This adds them as a resource (just like ), but also registers their `ComponentId` with the central `DevToolsRegistry`, which can be consumed by toolboxes to get a list of the available dev tools.
```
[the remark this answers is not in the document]
alice-i-cecile:
```suggestion
This adds them as a resource (just like the familiar `.init_resource` or `insert_resource`), but also registers their `ComponentId` with the central `DevToolsRegistry`, which can be consumed by toolboxes to get a list of the available dev tools.
```

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,658 @@
...
+`bevy_dev_tools` comes with two core traits: `ModalDevTool` (for tools that can be toggled) and `DevCommand` (for operations with an immediate effect on the game world).
+These are registered in the `DevToolsRegistry` via methods on `App`, allowing tool boxes (such as a dev console) to easily identify the available options and interact with them without the need for ad-hoc user glue code.
+
+## Motivation
+
+[`bevy_dev_tools`](https://github.com/bevyengine/bevy/tree/main/crates/bevy_dev_tools) was recently added to Bevy, giving us a home for first-party tools to ease the developer experience.
+However, since its inception, it has been plagued by debate over ["what is a dev tool?"](https://github.com/bevyengine/bevy/issues/12358)
+and every **tool** within both Bevy and its ecosystem follows its own ad hoc conventions about [how it might be enabled and configured](https://github.com/bevyengine/bevy/issues/12368).
+
+This is frustrating to navigate as an end user: every tool has its own way to configure it, with no rhyme, reason or way to set the same property to the same value across tools.
```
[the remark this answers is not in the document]
alice-i-cecile:
```suggestion
This is frustrating to navigate as a Bevy user: every tool has its own way to configure it, with no rhyme, reason or way to set the same property to the same value across tools.
```

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,658 @@
...
+
+Future possibilities:
+
+1. Over time, we can extend the `ModalDevTool` and `DevCommand` traits with more methods (mandatory and optional), enabling:
+   1. Gradual unification of more complex shared configuration (e.g. a `set_font` method).
+   2. Serializing and deserializing structs from a `.bsn` file, rather than simple CLI-style strings, allowing toolboxes to build abstractions for persistent config.
+2. Add support for a trait-queries-on-resources operation, and stop storing modal dev tools in the `DevToolsRegistry`.
+3. A derive macro for `DevCommand`, that aligns with `clap`'s calling conventions.
+4. Add specialized types of dev tools, likely each with their own trait to establish conventions and resolve conflicts between them.
+    1. For example, we may want to coordinate between various UI elements or overlays to avoid positioning conflicts, dev tools that change the materials of objects, or camera controllers.
```
alice-i-cecile, line 657:
```suggestion
    1. For example, we may want to coordinate between various UI elements or overlays to avoid positioning conflicts, dev tools that change the materials of objects, or camera controllers. See [Discussion #12589](https://github.com/bevyengine/bevy/discussions/12589) for some initial exploration.
```

rfcs/77-dev-tools-abstraction.md
naomijub: From my perspective, I like them separated. In my mind this goes well with the comment I made about they being +- equivalent to Resources and Events, so the separation makes them pretty clear

pablo-lua: The approach will change depending on what we do, if we have a function parsing both the commands and modals at the same block, I think we might need to unify the metadatas. At the same time, if we choose to have two functions: A command parsing method and a tool parsing method, this can be simplified and we can have two distinct metadatas.
But that will have significant differences in parsing, as we have to know beforehand if we are parsing a tool or a command.

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,658 @@
...
+/// 
+/// Their configuration is stored as a resource (the type that this trait is implemented for),
+/// and they can be enabled, disabled and reconfigured at runtime.
+/// 
+/// The documentation on this struct is reflected, and can be read by toolboxes to provide help text to users.
+trait ModalDevTool: Resource + Reflect + FromReflect + FromStr<Err=DevToolParseError> + Debug {
+    /// The name of this tool, as might be supplied by a command line interface.
+    fn name() -> &'static str {
+        Self::type_name().to_snake_case()
+    }
```
pablo-lua, line 111: I don't think we can have this trait as object safe at all, we can't have empty methods here, all them must accept &mut self or &self

Let alone those methods, we can't have FromStr and Debug as bounds because those traits aren't object safe, and I'm not sure if Resource is object safe at all, maybe we are forced to reflect all we can and have a FromStr reflected somehow.

pablo-lua: I think impl will be more easy if we have this as object safe, So we might have the name inside metadata and maybe call &self in metadata so we can be object safe?

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,658 @@
...
+    dev_commands: HashMap<String, DevCommandMetaData> 
+}
+```
+
+There are a few key operations that we will need to perform:
+
+```rust
+impl DevToolsRegistry {
+    /// Gets a reference to the specified modal dev tool by `ComponentId`.
+    fn get_tool_by_id(world: &World, component_id: ComponentId) -> Option<&dyn ModalDevTool> {
```
pablo-lua, line 473: As said above, we can't have this as object safe, but maybe we can have this return dyn Reflect?

rfcs/77-dev-tools-abstraction.md
rewin123, line 111:
```suggestion
    }
    
    fn short_description() -> Option<&'static str>;
```
Can we add an optional option to get a short text description of the dev tool? This would be very useful for different toolboxes, allowing to display a list of tools with their short descriptions. For example, if you type "tool list" into the console, I would expect such a list. Or, for example, it would allow to create a command palette like in VSCode

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,658 @@
...
+    }
+
+    /// The metadata for this modal dev tool.
+    fn metadata() -> DevToolMetaData {
+        DevToolMetaData {
+            name: self.name(),
+            type_id: Self::type_id(),
+            type_info: Self::type_info(),
+            // A function pointer, based on the std::str::from_str method
+            from_str_fn: <Self as FromStr>::from_str
```
rewin123, line 120:
```suggestion
            from_str_fn: <Self as FromStr>::from_str,
            short_description: Self::short_description()
```

rewin123 (commented):
Can we add an optional option to get a short text description of the dev tool? This would be very useful for different toolboxes, allowing to display a list of tools with their short descriptions. For example, if you type "tool list" into the console, I would expect such a list. Or, for example, it would allow to create a command palette like in VSCode.

~~Although I found some scattered mentions of a doc() function applied to commands, but there is no description of it in trait, nor is there an explicit description of how it should work.~~

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,658 @@
...
+trait ModalDevTool: Resource + Reflect + FromReflect + FromStr<Err=DevToolParseError> + Debug {
+    /// The name of this tool, as might be supplied by a command line interface.
+    fn name() -> &'static str {
+        Self::type_name().to_snake_case()
+    }
+
+    /// The metadata for this modal dev tool.
+    fn metadata() -> DevToolMetaData {
+        DevToolMetaData {
+            name: self.name(),
```
rewin123, line 116:
```suggestion
            name: Self::name(),
```
The current variant will cause a compilation error, it should be Self::name()

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,658 @@
...
+    // Clear the events, and act if at least one is found
+    if event_reader.drain().next().is_some() {
+        println!("Available modal dev tools:");
+        for (_component_id, modal_dev_tool) in dev_tools_registry.iter_modal_dev_tools() {
+            println!("{}", modal_dev_tool.type_name());
+            println!("{}", modal_dev_tool.docs());
+
+            for field in modal_dev_tools.fields(){
+                println!("{}", field.type_name());
+                println!("{}", field.docs())
```
rewin123, line 319: I feel like this method needs to be added to traits, otherwise where does it come from 
Perhaps this is a plan to add a short description to the tools?

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,658 @@
...
+this MVP should be sufficient to prove out the viability of the core architecture.
+
+## Implementation strategy
+
+### What metadata do toolboxes need?
+
+We can start simple:
+
+- the type name of the dev tool
+- any doc strings on the dev tool struct
```
rewin123, line 423: It also seems to be suggested to be implemented, but is mentioned in passing in two parts of the text. Would you like to add a doc() method to ModalDevTool and DevCommand? Or is it intended to retrieve documentation via reflection, as mentioned in the comments before traits? I think the doc part is very unclear in rfc

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,658 @@
...
+    }
+
+    /// The metadata for this dev command.
+    fn metadata() -> DevCommandMetadata {
+        DevCommandMetadata {
+            name: self.name(),
+            type_id: Self::type_id(),
+            type_info: Self::type_info(),
+            // A function pointer, based on the std::str::from_str method
+            from_str_fn: <Self as FromStr>::from_str
```
rewin123, line 244:
```suggestion
             from_str_fn: |s| {
                Self::from_str(s).map(|c| Box::new(c) as Box<dyn Reflect>)
            }
```

I'm not sure if we can store a typed from_str(&str) -> Self 
I'd suggest storing a lumbda function from this that returns Box<dyn Reflect>

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,658 @@
...
+    fn name() -> &'static str {
+        Self::type_name().to_snake_case()
+    }
+
+    /// The metadata for this dev command.
+    fn metadata() -> DevCommandMetadata {
+        DevCommandMetadata {
+            name: self.name(),
+            type_id: Self::type_id(),
+            type_info: Self::type_info(),
```
rewin123, line 242: The Self::type_id() and Self::type_info() methods actually require &self as an argument, so cannot be called this way. However, we can get this information via the GetTypeRegistration trait (for example ``` type_info: Self::get_type_registration().type_info() ```) which is automatically implemented via ``` #[derive(Reflect)]```. I suggest to add this trait to the list of parents of ModalDevTool and DevCommand traits

rfcs/77-dev-tools-abstraction.md
alice-i-cecile: I remember being able to do this with  reflection: we need to figure out how to actually access that information still.

rewin123: I just realised that maybe the RFC doesn't explicitly say this (or I missed it), but it is supposed to provide in near future a mechanism for automatic parsing of tool structures from comman line interface based on reflection?

rewin123: In terms of not just providing the FromStr mechanism, but being able to show the fields, their types, description in the CLI, and then from the standard CLI string atomatically fill in the fields of the tool's structure, bypassing FromStr<T>?

rewin123: > I remember being able to do this with reflection: we need to figure out how to actually access that information still.

This method does exist, oops, sorry for the incorrect question. Although it is not available by default, but only by non-default flag "documentation" in bevy_reflect crate.

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,661 @@
...
+Modal dev tools are registered via `app.init_modal_dev_tool::<D>()` (to use default config based on the `FromWorld` implementation) or via `app.insert_modal_dev_tool(config: D)`.
+This adds them as a resource (just like the familiar `.init_resource` or `insert_resource`), but also registers their `ComponentId` with the central `DevToolsRegistry`, which can be consumed by toolboxes to get a list of the available dev tools.
+
+To build your own modal dev tool, simply create a configuration resource, implement the `ModalDevTool` trait, and then register it in the app when the correct feature flag is enabled.
+
+```rust
+/// A flying camera controller that lets you disconnect your camera from the player to freely explore the environment.
+/// 
+/// When this mode is disabled
+#[derive(Resource, Reflect, FromReflect, Debug)]
```
rewin123, line 163:
```suggestion
#[derive(Resource, Reflect, Debug)]
```

#[derive(FromReflect)] is not using now

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,661 @@
...
+Keeping track of the registered dev tools without storing them all in a dedicated collection is quite challenging!
+How can we keep track of it under the hood?
+
+```rust
+#[derive(Resource)]
+struct DevToolsRegistry {
+    /// The stored collection of modal dev tools, tracked in a type-erased way using [`ComponentId`]
+    /// 
+    /// The key is the `name()` provided by the `ModalDevTool` trait.
+    modal_dev_tools: HashMap<String, ToolMetaData>,
```
matiqo15, line 463:
```suggestion
    modal_dev_tools: HashMap<String, DevToolMetadata>,
```

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,661 @@
...
+    /// 
+    /// The supplied name should match the `DevCommand::name()` method.
+    fn get_command_metadata(name: &str) -> Option<&DevCommandMetadata> {
+        self.dev_commands.get(name)
+    }
+
+    /// Gets the `DevToolMetadata` for a given dev tool by name.
+    /// 
+    /// The supplied name should match the `DevCommand::name()` method.
+    fn get_tool_metadata(name: &str) -> Option<&DevCommandMetadata> {
```
matiqo15, line 497:
```suggestion
    fn get_tool_metadata(name: &str) -> Option<&DevToolMetadata> {
```

rfcs/77-dev-tools-abstraction.md
alice-i-cecile: Yeah, this was simply pseudo-code: we need to figure out how to actually do this properly still.

afonsolage (commented):
I like the idea of having a common pattern for development tools. I'll read more on weekend but for now my first point is this RFC should decouple the parsing and keep `Reflect` as a generic requirement, since requiring `FromStr` will only be useful for "toolboxes" like `bevy_console`, for all others like having UI, remote debugging or even WASI integration, you won't use `&str` at all, you just want to serialize the command and process it or use `Reflect` meta info.

If `ModalDevTools: Reflect` this means any "toolbox" can parse a `String` without forcing the dev tools devs to do that.

```
my_awesome_command 10 15.5 enabled=true
```

Args without name would be used as positional (`Reflect::field_at(index)`) and args with name would be set with (`Reflect::field(name)`).

This way we shift the command parsing from `MoldaDevTools` itself and move to "toolbox" impls, making this even more flexible.

alice-i-cecile:
Yeah, the focus management thing is generally a big open question in Bevy as a whole: see https://github.com/bevyengine/bevy/issues/3570 for more context there. I think it can and should be tackled independently though, and done ad-hoc to start.

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,826 @@
...
+    fn get_tool_by_id(world: &World, component_id: ComponentId) -> Option<&dyn ModalDevTool> {
+        let resource = world.get_resource_by_id(component_id)?;
+        resource.downcast_ref()
+    }
+
+    /// Gets a mutable reference to the specified modal dev tool by `ComponentId`.
+    fn get_tool_mut_by_id(mut world: &mut World, component_id: ComponentId) -> Option<&mut dyn ModalDevTool> {
+        let resource = world.get_resource_mut_by_id(component_id)?;
+        resource.downcast_mut()
+    }
```
pablo-lua, line 645:
```rust
    pub fn get_tool_by_component_id<'w>(&self, cmp: ComponentId, world: &'w World) -> Option<&'w dyn Reflect> {
        let type_id = world.components().get_info(cmp)?.type_id()?;
        let type_registry = world.resource::<AppTypeRegistry>().read();
        type_registry.get(type_id)?.data::<ReflectResource>()?.reflect(world)
    }

    pub fn get_tool_by_component_id_mut<'w>(&self, cmp: ComponentId, world: &'w mut World) -> Option<Mut<'w, dyn Reflect>> {
        let type_id = world.components().get_info(cmp)?.type_id()?;
        let reflect_resource = {
            let type_registry = world.resource::<AppTypeRegistry>().read();
            a.get(type_id)?.data::<ReflectResource>()?.clone()
        };
        reflect_resource.reflect_mut(world)
    }
```
With this approach, we can return a dyn Reflect instead of dyn ModalDevTool while keeping component_id

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,826 @@
...
+    
+    /// Iterates over the list of registered modal dev tools.
+    fn iter_tools(&self, world: &World) -> impl Iterator<Item = (ComponentId, &dyn ModalDevTool)> { 
+        self.modal_dev_tools.iter().map(|&id| (id, self.get(world, id)))
+    }
+
+    /// Mutably iterates over the list of registered modal dev tools.
+    fn iter_tools_mut(&self, mut world: &mut World) -> impl Iterator<Item = (ComponentId, &mut dyn ModalDevTool)> { 
+        self.modal_dev_tools.iter_mut().map(|&id| (id, self.get(world, id)))
+    }
```
pablo-lua, line 686: A cleanup is probably needed here.

We should probably have the methods match the impl above (see my comment above), only the name and return will change.

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,826 @@
...
+
+    /// Iterates over the list registered dev commands, returning their name and `DevCommandMetadata`.
+    fn iter_commands(&self) -> impl Iterator<Item = (&str, `DevCommandMetadata)> {
+        self.dev_commands.iter()
+    }
+}
+```
+
+Once the user has a `dyn ModalDevTool`, they can perform whatever operations they'd like: enabling and disabling it, reading metadata and so on.
```
pablo-lua, line 695: Here we're talking about dyn ModalDevTool, but that's not possible, so this section should probably be changed.

pablo-lua (commented):
Some comments about object safety.

rfcs/77-dev-tools-abstraction.md
pablo-lua: I don't know for sure the implications of using ComponentId instead of TypeId, this can simplified if we don't need to get the type_id from components first.

rfcs/77-dev-tools-abstraction.md
```diff
@@ -0,0 +1,826 @@
+# Feature Name: `dev-tools-abstraction`
+
+## Summary
```
cart, line 3: Thank you for putting this together. I now understand where you all are coming from (for context: [my original comment pushing back against the "dev tool" concept](https://github.com/bevyengine/bevy/pull/12427#issuecomment-2005328288)).

I see the appeal of a common way to represent commands:

* Single searchable database of commands
* Common data model for command input

That being said, I think this is trying to be _too_ opinionated, and focusing on "dev tool" as a specific thing might be getting in our way:

1. Every "toolbox" will have a different way of interacting with these commands. Command serialization/deserialization is something the "toolbox" should be responsible for. We cannot just throw FromStr on the impl and let each dev tool trait impl choose what format to expect. We should rely on Reflect for this (with optional reflect-driven Serde support), similar to how we handle scene serialization. Leave things like the "command line format" to the "command line toolbox".
2. There is no reasonable way to solve the "Dev tool UI overlap" problem in a generic system. Each "toolbox" will define a different context for displaying and rendering "dev tools". The rendering/UX of a "dev tool" is _fully_ the responsibility of the toolbox, and most "interactive" functionality will be toolbox specific. A "dev-tool" that renders content inside of the Bevy Editor should define itself relative to the Bevy Editor's constraints (which are notably ... currently undefined). I do not think we should make solving this problem generically across toolboxes "future work". Instead it should be a non-goal.

I believe we should be fully focused on (1) a command-style abstraction and (2) queryable/categorized settings configured by commands. We should try very hard to build on what already exists when possible, and build reusable pieces when it isn't.

1. Reflect (with optional reflect-driven Serde) for all serialization
2. Use `Command` for "applying a command to a World" (I acknowledge that this is already part of the proposal)
3. Rely on existing registry tooling instead of building a new one. Some options:
    a. TypeRegistry enumeration of types registered with a specific reflected trait (ex: `DevCommand`)
    b. Make this Component / Query driven instead of Resource driven (ex: use a common DevCommand component)
4. Use Commands for configuring settings (including enabling / disabling features)

I think by shifting focus to these things, we sidestep the "dev tool is a nebulous concept" problem and focus on building the specific functionality we need. This also leaves the door open to other categories of commands / configuration.

For example (not a final design, just illustrating where my head is at):

```rust
// Reflect registry-driven commands

#[derive(Reflect, DevCommand, Debug)]
#[reflect(DevCommand)]
struct SetGold {
    amount: u64,
}

impl Command for SetGold {
    fn apply(self, world: &mut World){
        let mut current_gold = world.resource_mut::<Gold>();
        current_gold.0 = self.amount;
    }
}

app.register_type::<SetGold>()

// later, use the type registry to invoke / enumerate / serialize this command
type_registry.iter_with_data::<ReflectDevCommand>()
```

```rust
// Commands for enabling / disabling the settings of a dev tool

#[derive(Resource, Reflect, DevToolSettings, Debug)]
#[reflect(DevToolSettings)]
struct DevFlyCamera {
    enabled: bool,
    /// How fast the camera travels forwards, backwards, left, right, up and down, in world units.
    movement_speed: Option<f32>,
    /// How fast the camera turns, in radians per second.
    turn_speed: Option<f32>,
}

impl Toggleable for DevFlyCamera {
    fn is_enabled(&self) -> bool {
        self.enabled
    }
    
    fn set_enabled(&mut self, enabled: bool) {
        self.enabled = enabled;
    }
}

app.register_type::<DevFlyCamera>()
// Makes these Dev commands available (these could be put behind a helper function):
app.register_type::<Enable<DevFlyCamera>>()
app.register_type::<Disable<DevFlyCamera>>()
app.register_type::<Toggle<DevFlyCamera>>()
// this DevCommand writes a new value of DevFlyCamera on top of the existing value
app.register_type::<Configure<DevFlyCamera>>()

// later (ex: if you want to enumerate DevToolSettings types in a tool)
type_registry.iter_with_data::<DevToolSettings>()

/// Reusable Enable dev command for toggleable things
#[derive(DevCommand)]
pub struct Enable<T: Toggleable>();

impl<T: Toggleable> Command for Enable<T> {
    // TODO
}
```

alice-i-cecile: I can absolutely work with that! I'll do a clean up pass next week, reframing and simplifying to capture these ideas.

cart: Awesome!

Also note that `type_registry.iter_with_data` does not currently exist. You can do something similar with `.iter()` and filtering the results, but we could consider providing that ourselves (and _maybe_  building an index to accelerate, although building that index might be too detrimental to startup times and I'm not sure we need it.

matiqo15: Me and @alice-i-cecile noticed during the re-designing of the abstraction to use `TypeRegistry` a few issues:

1. `TypeRegistry` was not built for this. The `TypeRegistry` was designed to store the registration of types, not tools or metadata of them. If we would use this, we are potentially allowing usage of `TypeRegistry` for storing **anything**.
2. Using a dedicated resource (e.g. DevToolsRegistry) is a standard design pattern in the engine. For example, let's take a look at [`DiagnosticStore`](https://github.com/bevyengine/bevy/blob/5357e15966b7936d1134df44b90f347b0f8e1edb/crates/bevy_diagnostic/src/diagnostic.rs#L295-L299) resource that stores diagnostic information. Similarly, a `DevToolsRegistry` could be used to store and manage tools and commands, while hiding a lot of *magic* that could be intimidating to users.
3. In the `DevToolsRegistry` we would store metadata about a tool or a command, which is accessible by providing its name. While this still *could* be possible with `TypeRegistry` it would be a misuse of the registry, and would create additional overhead.
4. `TypeRegistry` limits our flexibility and extensibility of the abstraction in the future.

The use of `Reflect` instead of `FromStr` is great though!

cart: > TypeRegistry was not built for this. The TypeRegistry was designed to store the registration of types, not tools or metadata of them. If we would use this, we are potentially allowing usage of TypeRegistry for storing anything.

Tools and metadata are represented as types, and _whether_ a type is a tool (or command) should be encoded as a Rust Trait. From my perspective these line up perfectly with the TypeRegistry's functionality and purpose.

> In the DevToolsRegistry we would store metadata about a tool or a command, which is accessible by providing its name. While this still could be possible with TypeRegistry it would be a misuse of the registry, and would create additional overhead.

From my perspective the "dev tool name" should be the type name (short name by default with disambiguation possible via the full type path). Again this feels like a great fit (and the right one). I consider the overhead of querying the reflected trait registration for a type to be acceptable for this use case (and _just barely_ more expensive than the `HashMap<String, Data>` that a custom registry would use).  

I think the biggest missing piece is how DevCommands correlate to dev tools / DevToolSettings (which we could just shorten to DevTool if we want). For example: the DevTool trait could have a function that provides this correlation.

> TypeRegistry limits our flexibility and extensibility of the abstraction in the future.

This was intentional. My goal here was to put bounds on this abstraction in a way that prevents it from becoming a new unbounded "framework". I'd like to reconcile the requirements of the "dev tool" space with the rest of Bevy.

pablo-lua: I agree here: I don't really see a use for Metadata anymore if we can use the type registry for many things
For example, let's say we have the `Toggleable` trait, and the `ReflectToggleable` type and the 3 new commands: `Toggle<T>`, `Enable<T>` and `Disable<T>`, and the CLI tool as an example of functionality here. 
We can have the CLI tool be just a `HashMap<ToolName, TypeId>` (tool name would probably be of type `String` here). We can just have the CLI tool use the name here and lookup in type registry for the `ReflectToggleable` data. For inserting new tools with CLI, we would use the same approach, but with `ReflectDeserialize` instead. 
And so, this would lessen the uses for metadata. Note that with this approach, the CLI tool for example would have to use serde with deserialize, as we don't really have a way to store the given function to deserialize the tool without some metadata. But with the main commands this RFC proposes: Enable, disable and toggle, this will go alright without metadata.
At the same time, I don't see a way to go without a `ToolBox` resource that will store at least the TypeId and the Tool name.

matiqo15: > At the same time, I don't see a way to go without a ToolBox resource that will store at least the TypeIdand the Tool / command name.

That's why I believe we should have a registry. If we would ask every toolbox creator to "make your own storage for tools", we would end up with multiple resources that have the same purpose.

pablo-lua: > That's why I believe we should have a registry

Yeah, I believe we at can have a global resource to store what tools the app currently have, and rely more on the app registry for parse and creation of commands

matiqo15:
I've been experimenting with abstraction for the dev tools for some time now and:

1. I still don't think type registry is a good idea:

From my understanding `iter_with_data` would require `DevTool` to be object-safe otherwise, we would never be able to get it, even as a `&dyn Reflect`. This is because we would need to use a `ReflectDevTool` to check whether it is a dev tool. 

Even if we made `DevTool` object-safe, we would still run into problems getting `&dyn DevTool` from `ReflectDevTool` without type annotation.

```rust
    // New method on `TypeRegistry`
    pub fn iter_with_data<T: TypeData>(&self) -> impl Iterator<Item = (&TypeRegistration, &T)> {
        self.registrations.values().filter_map(|item| {
            let type_data = self.get_type_data::<T>(item.type_id());
            type_data.map(|data| (item, data))
        })
    }
    
    // Possible implementation of printing name of every tool
    for dev_tool in type_registry.iter_with_data::<ReflectDevTool>() {
        // This is the problem - we need to specify a tool here if we want to get a `&dyn DevTool`
        let reflected = Box::new(DevFlyCamera::default());
        if let Some(tool) = dev_tool.1.get(&*reflected) {
            println!("{:?}", tool.name());
        }
    }
```

2. Why shouldn't it be object-safe

We would need to drop `FromReflect`, and as a result of this, we would run into problems with executing dev commands.

3. DevToolsRegistry

You don't want dev tools to become a new unbounded framework - but actually, our abstraction can restrict what fits in it. With a type registry, you can always register a type, but whether you would be able to register a tool in our registry is going to be decided by the API of the registry. Would it be opinionated? - yes, but I think it should.

vladinator1000:
Y'all, am I the only one thinking this spec sounds exactly what an editor is for? Like a temporary solution that alleviates the lack of editor conventions.
> Regardless of this classification, dev tools are:
    not intended to be shipped to end users 
    highly application specific
    enabled or operated one at a time, on an as-needed basis, when problems arise
    intended to avoid interfering with the games ordinary operation, especially when disabled
    expansive and ad hoc

NthTensor:
Closing. Reincarnated as a project goal. The dev-tools abstractions are dead, long live the dev-tools abstractions.

PR #78, opened 2024-03-24
Entity Relations

Trashtalk217:
I politely ask you to refrain from commenting on this PR for now (just pretend like you didn't see it).
I've copied this verbatium from Boxy's earlier entity relations rfc and before I want to open it up for discussion, I'd like to first

- [ ] Give it a thorough review (hence the draft).
- [ ] Incorporate the relevant sections from Sanders [Roadmap to Entity Relations](https://medium.com/@ajmmertens/a-roadmap-to-entity-relationships-5b1d11ebb4eb) as a start to the implementation plan.
- [ ] Fill in some of the other left over gaps.

PR #79, opened 2024-04-01
Eliminating blockers for Minimal Fragmenting Relationships

james-j-obrien:
RFC with a focus on the technical blockers for a hypothetical minimal fragmenting relationships implementation. Less time has been spent on the summary and motivation aspects as those have been well explored by Sander's articles ([example](https://ajmmertens.medium.com/building-games-in-ecs-with-entity-relationships-657275ba2c6c)) as well as prior RFCs, PRs, discord discussions etc.

[RENDERED](https://github.com/james-j-obrien/rfcs/blob/minimal-fragmenting-relationships/rfcs/79-minimal-fragmenting-relationships.md)


Trashtalk217:
This looks great! And I think it's also a more reasonable size than #78 (which I should probably close for now). One thing I would like slightly more clarification on is the 'fragmenting' part of 'fragmenting' relationships. Even in #78, I don't think it's adequately explained. My current guess, based on context clues, is that adding relationships means that every (component, entity) acts like a new component for the archetype system, so (component, entity1), and (component, entity2) would be in two different tables, hence splitting the tables, hence fragmenting the tables, which makes iteration speed for queries slower (less predictable access patterns).

I think that's it, but wouldn't that also mean that when you don't use relations, you don't encounter any extra fragmentation, so you shouldn't see performance regressions?

james-j-obrien:
Your interpretation of the "fragmenting" part is correct, I took a quick edit to make it more explicit but it could use more detail. Iteration for queries could hypothetically be slower for a set of entities that previously would have occupied the same table however if the query is only iterating dense components we can still use dense iteration and shouldn't see any regressions. One advantage we have over flecs here is we already have split storage. Since this RFC doesn't cover data on the relationships iterating targets (which are stored in the archetype not component storage) will always be dense and when we do support them we opt for sparse storage when we don't want to impact iteration speed of dense queries that don't access the relationship.

Your also correct in that if you don't make use of relationships you won't have any increase to the fragmentation, however once you do it can increase quickly, especially if/when `bevy_hierarchy` uses them.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,126 @@
...
+
+Another analogy when considering the ECS as an in memory database is that relationships act as foreign keys. This unlocks a potential future where we can express more complex queries matching across multiple entities and relationship pairs. There are a myriad of additional features that can be built on top of the initial implementation described below. The main focus in this RFC will be the blocking issues that must be addressed before the initial implementation is achievable.
+## Motivation
+
+Fragmenting relationships has been a long desired bevy feature. The ability to express and traverse hierarchies and graphs within the ECS unlocks a huge amount of potential. Features such as breadth-first query traversal, immediate hierarchy clean-up, component inheritance, multi-target queries etc. all rely on or are simplified by the existence of fragmenting relationships.
```
james7132, line 12:
```suggestion
Relationships has been a long desired bevy feature. The ability to express and traverse hierarchies and graphs within the ECS unlocks a huge amount of potential. Features such as breadth-first query traversal, immediate hierarchy clean-up, component inheritance, multi-target queries etc. all rely on or are simplified by the existence of some form of relationships in the ECS. 
```
I don't think a specific implementation of relations was the target, just relations in general.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,126 @@
+# Feature Name: `minimal-fragmenting-relationships`
+
+## Summary
+
+The central idea behind fragmenting relationships is the ability to express a connection between two `Entity`s by adding a pair of `(ComponentId, Entity)` as if it were a component. For example if one had several entities representing different kinds of swords one could add the pair `(Equipped, sword_entity)` to the player to signify which is currently equipped. 
+
+Of course you could also just have created an `Equipped` component with an `Entity` field member to represent the same relationship. The advantage to "fragmenting" relationships is each unique pair is treated as if it were a unique component id, forming their own archetypes. This means we could also add `(Equipped, shield_entity)` to the same player and offer performant query APIs that express `(Equipped, *)`to find all equipped pairs.
```
james7132, line 7: This part is focusing on a surface level summary as to why users would want this. Making their own archetypes is useful, but to the average user, this takes the form of being able to express it as a part of the queries they're already used to working with.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,126 @@
...
+
+Relationship pairs can be inserted and removed like any other component. There are 3 ids associated with any relationship pair instance:
+1. The source entity that the pair is being added to
+2. The relationship component
+3. The target entity
+
+In the minimal implementation all relationships will take the form of `(ZST Component, entity)` pairs so the API surface will look quite familiar. All naming is subject to bikeshedding:
+```rust
+#[derive(Component)]
+struct Relationship;
```
james7132, line 23: Would it make sense to explicitly designate a component as a relationship component, or should we allow it to be used as a non-relationship context too?

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,126 @@
...
+	.build();
+
+query.single(&world); // == alice
+```
+
+## Implementation strategy
+
+There are a number of technical blockers before even a minimal version of relationships could be implemented in bevy. I've broken them down below:
+#### Archetype Cleanup
+In order to fit each pair into a single id we lose the entity generation (see this [article](https://ajmmertens.medium.com/doing-a-lot-with-a-little-ecs-identifiers-25a72bd2647) for more details). As a consequence when an entity is despawned we need to ensure that all the archetypes that included that entity as a target are also destroyed so that a future entity re-using that id is not misinterpreted as being the old target. This is non-trivial.
```
james7132, line 80: In terms of typing, is this the `Identifier` type? Another representation of `Entity`? This should be detailed as a part of an RFC, as an RFC must be implementable in some form.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,126 @@
...
+
+Luckily observers [#10839](https://github.com/bevyengine/bevy/pull/10839) can help solve this problem. By sending ECS events each time an archetype is created and creating an observer for each query we can target the updates to query caches so that we no longer need to iterate every new archetype. Similarly we can fire archetype deletion events to cause queries to remove the archetype from their cache.
+
+In order to implement such a mechanism we need some way for the observer to have access to the query state in order to update it. This can be done by moving the query state into a component on the associated entity (WIP branch here: [queries-as-entities](https://github.com/james-j-obrien/bevy/tree/queries-as-entities)) and will require further refactors to the cache to optimize for removal. However this is largely blocked by the deletion of entities in the render world tracked here: [#12144](https://github.com/bevyengine/bevy/issues/12144).
```
james7132, line 88: What isn't mentioned here is that matched storages are currently stored in Vecs, which presents a `O(n)` cost of removal, scaling with the number of tables/archetypes that already matched the query. Is that to remain the same, or do we need to restructure the internal caches of queries to support this?

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,126 @@
...
+##### Access Bitsets and Component SparseSets
+The `FixedBitSet` implementation used by `Access` inside of systems, system params, queries etc. relies on dense component ids to avoid using a large amount of memory. If we were to start inserting ids with arbitrarily large entity targets as well as component ids shifted into the upper bits each `FixedBitSet` would allocate non-trivial amounts of memory to have flags for all the intervening ids. In order to minimize the affects of this we could consider all `(R, *)` pairs as one id in accesses, reducing the amount we need to track at the expense of some theoretical ability to parallelise.
+
+Similarly the `SparseSet` implementation used in both table and sparse component storage also operate under the assumption that component ids will remain low and would allocate an even larger amount of memory storing indexes for all the intervening ids when storing relationship pairs on an entity.
+
+The most practical way to solve this is using paged versions of the appropriate data structures, for bitsets this can take the form of existing hierarchical or compressed bitsets like `roaring`.
```
james7132, line 96: If we're strictly talking about minimizing memory costs, `roaring` is probably OK, but it should be noted that this incurs a *very* heavy cost compared to uncompressed bitsets or even normal sparse set storage when working with random lookups.

This may be acceptable as relations are expected to support use cases like transform and visibility propagation via BFS, which are currently served by these point queries, but the general use case for them and `Query::iter_many` may pose a potential issue, particularly since the latter is heavily used in the renderer.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,126 @@
...
+Dynamic queries would also be able to match on specific pairs:
+```rust
+let query = QueryBuilder::new::<Entity>(&mut World)
+	.with_pair::<Eats>(apple)
+	.build();
+
+query.single(&world); // == alice
+```
+
+## Implementation strategy
```
james7132, line 76: A mention of how this interacts with intra-system parallelism is probably worth mentioning, since all of the user-facing examples are all single-threaded. A simple "this strictly relies on archetype matching and thus has no impact when used in conjunction with APIs like `Query::par_iter`" should be sufficient.

rfcs/79-minimal-fragmenting-relationships.md
james7132, line 96: Paging only resolves this within a fixed scalar factor. If we use 64-element pages, we incur 64x less memory used, but if the IDs are using bit shifted values, that's only going to account for 6 bits out of the 32 we're using at the very minimum, resulting in *very* large SparseSets being allocated. 

More care to ensure that encoded IDs in the sparse set is required (i.e. using morton encoding), and some acknowlegement of the potential costs of computing this when doing these lookups is warranted.

rfcs/79-minimal-fragmenting-relationships.md
james7132, line 88: How are we to handle all of the potential interactions that an entity supports when dealing with soundness?

 * A `Query` as a component being `std::mem::swap`'ed with another with a different component access or different `WorldId`. This would trivially enable unsound behavior without use of `unsafe`.
 * `Query` state components being queried by other queries, and then used to access entities that are not in the original component access set.
 * A `Query` has been despawned.

These need to be addressed in some way in this PR to ensure we're not introducing trivial unsoundness.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,126 @@
...
+- Refactor queries to become entities with caches updated via observers
+- Implement archetype deletion and the associated clean-up for systems and queries 
+- Replace `FixedBitSet` and `SparseSet` implementations with ones that support sparse ids
+
+After all the issues above are addressed the actual implementation of the feature can begin. This requires changes in 3 main areas:
+- Implementation of `WorldData` and `WorldFilter` implementations for accessing relationship targets
+- Transition to new `Identifier` implemented in [#9797](https://github.com/bevyengine/bevy/pull/9797) instead of `ComponentId` in all places where pairs could be expected: tables, archetypes etc. 
+- New methods for `EntityCommands`, `EntityMut`/`Ref`/`WorldMut` and `QueryBuilder`
+## Drawbacks
+- Increased archetype fragmentation when in use
```
james7132, line 110: Increasing archetype fragmentation has an impact on memory fragmentation and query iteration performance, both of which should be detailed here.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,126 @@
...
+Similarly the `SparseSet` implementation used in both table and sparse component storage also operate under the assumption that component ids will remain low and would allocate an even larger amount of memory storing indexes for all the intervening ids when storing relationship pairs on an entity.
+
+The most practical way to solve this is using paged versions of the appropriate data structures, for bitsets this can take the form of existing hierarchical or compressed bitsets like `roaring`.
+### Minimal Implementation
+Implementing the above in order of dependance could look like:
+- Resolve [#12144](https://github.com/bevyengine/bevy/issues/12144) so entities can be persisted in the render world.
+- Merge observers [#10839](https://github.com/bevyengine/bevy/pull/10839)
+- Refactor queries to become entities with caches updated via observers
+- Implement archetype deletion and the associated clean-up for systems and queries 
+- Replace `FixedBitSet` and `SparseSet` implementations with ones that support sparse ids
```
james7132, line 103: Due to how much of the internals of the ECS are involved in this effort, this needs to be explicitly ordered in a way that we can de-risk a blocker leaving us in a half-complete state.

rfcs/79-minimal-fragmenting-relationships.md
iiYese: @james-j-obrien I think the other possible implementations of relationships have been explored at this point & the nice doogfoody features that comes about from relations are in no small part a result of the implementation being doogfoody. Specifically it makes sense for archetype level capabilities like fast querying, cleanup & traversal to come about from an archetype level implementation in an archetypal ECS.

iiYese:
# "Fragmenting"
I'd like to add some comments about "fragmenting" since that's bound to receive the most questions. In short, we should stop calling it "fragmenting" it places too much emphasis on the potential downsides of the feature. A better description for this is "archetype level". This is a tool which can have downsides but has many desirable upsides akin to the archetype ECS model we already have.

## Other implementations
I do not think this RFC or any efforts towards this feature should deviate away from an archetype level implementation and generalizing to just "an implementation of relations" would be a mistake. The other big contenders for a relations implementation have been explored at this stage.
- Indexed
- "Non-fragmenting" (aery)
 
### Indexed
To my knowledge there's no specific crate like aery that offers this as a standalone feature but this already exists in every other crate. Many crates have an "index" of some sort in a resource that creates relationships between entities or data stored outside the ECS. The problem with all of these is indexing is a very different data model to archetypes. This presents syncing problems and an even bigger ergonomic problems. Different data models create boundaries in your app that you have to cross every time you access data. Think getting stuff in & out of egui contexts for widgets.

### Non-fragmenting
Aery was my attempt at this for my final year university project. It has less of the problems of indexed relations because it's closer to the archetype model (it just stores `Entity`s in components). It's serviceable for many usecases but it's very unoptimal for many more & has nowhere near the amount of query features I'd like. It originally started off as a fork of bevy but half way through I decided I didn't want it in bevy for two reasons.
- [The performance problems of highly fragmented scenes can be addressed](https://twitter.com/ajmmertens/status/1639725947258822656)
- A better version of "non-fragmenting" relationships can be implemented on top of archetype level relationships.

# TLDR
In short **it makes the most sense for an ECS relationships impl to use the same data model as the ECS**. Archetype level relationships are what we should aim for. The worst case scenario is you would have as many archetypes/tables as you have entities which is already how the vast majority of commercial games written in OO paradigms work.

rfcs/79-minimal-fragmenting-relationships.md
james-j-obrien: This is what I was referring to by "optimize for removal", will add more detail here.

james-j-obrien: I haven't spent time here expanding the details of queries as entities here, but part of that process would be taking `QueryState` private so that you can't run into those first two problem as a user.

Despawning a query is always going to be problematic but most users won't be querying for queries so unless they do a naive "despawn all entities" operation it's unlikely a user will run into this. If they do despawn a query that's used in a system it will likely result in a panic.

rfcs/79-minimal-fragmenting-relationships.md
james-j-obrien: I think it's not unlikely we will need a custom implementation optimized for our purposes so broadly agree here.

rfcs/79-minimal-fragmenting-relationships.md
james-j-obrien: I think having a derive specific to relationship components is fine, there are potential cases where someone might want to use the same type for both, but they can always newtype.

No strong feelings here.

rfcs/79-minimal-fragmenting-relationships.md
SanderMertens: For clarification, I don't use sparse sets anywhere in Flecs where a pair can be used as key, as even with a paged sparse set that could allocate lots of pages because of how wide the range of possible values is.

I'm using hashmaps wherever pairs are used as keys, sometimes combined with an array for the lower id ranges where performance matters.

rfcs/79-minimal-fragmenting-relationships.md
SanderMertens: > If they do despawn a query that's used in a system 

Simple solution for this is to also associate the system with the same entity (which is what I do), so that when you delete the query entity you're guaranteed to also delete the system.

rfcs/79-minimal-fragmenting-relationships.md
james-j-obrien: That's true. For access bitsets it should be feasible when collapsing all `(R, *)` as we effectively only double our id space, we just have to deal with them being sparse.

james-j-obrien: From what I understand the way lookups stay fast in `flecs` without needing the sparse lookup in the table is using the component index so I suppose I should also detail that here.

rfcs/79-minimal-fragmenting-relationships.md
james-j-obrien: We have multiple queries per system but we can use relationships to associate them eventually. 

Users may encounter some footguns with despawning a query by accident and having arbitrary systems stop running, but that's a debate to be had when we get there, in the mean time I think a panic is fine.

rfcs/79-minimal-fragmenting-relationships.md
SanderMertens: The impact of fragmentation is complex:
- there is a wide range of different scenarios with wildly different performance characteristics
- there are many scenarios where fragmentation can improve query performance
- component fragmentation can be kept in check because of the table/archetype split
- there's prior art for a [non-fragmenting relationship implementation](https://www.flecs.dev/flecs/md_docs_2Relationships.html#union-property) which can coexist with a fragmenting one

The flecs vs. bevy hierarchy benchmarks (which highlights just one narrow aspect of relationships) show that the impact is not uniformly good or bad, with narrow trees being faster without fragmenting relationships, and wide trees being (much) faster:

Screenshot 2024-04-03 at 9 43 42 AM

(flecs left, bevy right, number is flecs / bevy)

tl;dr when calling this out it's easy to focus on the negatives of fragmentation which are easy to understand without balancing that out with the positives, which are harder to understand and measure without an implementation, but do exist and have been proven.

rfcs/79-minimal-fragmenting-relationships.md
james7132: I'm not asking for the RFC to be generalized here, but just to relax the wording for the average user of the feature. This is supposed to be a user facing summary of the motivation, not the exact technical reasons why we're doing this over other implementations.

SanderMertens:
> The other two big contenders for a relations implementation in the context of an archetypal ECS have been explored at this stage:

There's also [Flecs union relationships](https://www.flecs.dev/flecs/md_docs_2Relationships.html#union-property), which are a non-fragmenting relationships implementation that are widely used by flecs users in combination with fragmenting relationships.

A fourth non-fragmenting implementation has just finished and will land soon, which'll let you use component members as targets of relationships.

rfcs/79-minimal-fragmenting-relationships.md
james7132: > I haven't spent time here expanding the details of queries as entities here, but part of that process would be taking `QueryState` private so that you can't run into those first two problems as a user.

This probably deserves mentioning. Any potential soundness hole in the implementation is a hard blocker here.

Another potential item that might need fully exploring in this RFC is how queries as entities integrates with existing system accesses. An `Access<ComponentId>` will always conflict on the query component unless something is changed in the access model, or the conflict is explicitly ignored. `Access<ArchetypeComponentId>` may also need to factor this in as well.

james-j-obrien: Since `QueryState` doesn't need to be mutated to iterate the query systems can all have read access to the `QueryState` component without causing issues for soundness or parallelism.

james7132: That's true within a system, but the schedule executor needs to mutate them to update their archetypes before the  systems that use them can run, potentially while other systems are running. Trivial in single threaded executors, but a way to prevent aliasing in the multithreaded executor is needed.

james-j-obrien: As part of queries as entities we would also be transitioning to relying on observers to update the query caches, so those updates would be applied during command application.

Is there some other mutation that needs to be done during scheduling?

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,128 @@
...
+
+Another analogy when considering the ECS as an in memory database is that relationships act as foreign keys. This unlocks a potential future where we can express more complex queries matching across multiple entities and relationship pairs. There are a myriad of additional features that can be built on top of the initial implementation described below. The main focus in this RFC will be the blocking issues that must be addressed before the initial implementation is achievable.
+## Motivation
+
+Relationships has been a long desired bevy feature. The ability to express and traverse hierarchies and graphs within the ECS unlocks a huge amount of potential. Features such as breadth-first query traversal, immediate hierarchy clean-up, component inheritance, multi-target queries etc. all rely on or are simplified by the existence of relationships as first class ECS primitive.
```
Trashtalk217, line 12:
```suggestion
Relationships have been a long desired bevy feature. The ability to express and traverse hierarchies and graphs within the ECS unlocks a huge amount of potential. Features such as breadth-first query traversal, immediate hierarchy clean-up, component inheritance, multi-target queries etc. all rely on or are simplified by the existence of relationships as first class ECS primitive. 
```

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,128 @@
...
+
+In order to implement such a mechanism we need some way for the observer to have access to the query state in order to update it. This can be done by moving the query state into a component on the associated entity (WIP branch here: [queries-as-entities](https://github.com/james-j-obrien/bevy/tree/queries-as-entities)) and will require further refactors to the cache. For example currently the collection of tables and archetypes is a vector but that doesn't provide an efficent method for removal without scanning the entiriety of the vector so this data structure will have to replaced with one that performs reasonably for both iteration and insertion/removal. `QueryState` would then be taken private in order to prevent unsafe accesses and users would refer to queries either by their type signature or entity id.
```
Trashtalk217, line 89:
```suggestion
In order to implement such a mechanism we need some way for the observer to have access to the query state in order to update it. This can be done by moving the query state into a component on the associated entity (WIP branch here: [queries-as-entities](https://github.com/james-j-obrien/bevy/tree/queries-as-entities)) and will require further refactors to the cache. For example, currently the collection of tables and archetypes is a vector but that doesn't provide an efficient method for removal without scanning the entirety of the vector so this data structure will have to be replaced with one that performs reasonably for both iteration and insertion/removal. `QueryState` would then be taken private in order to prevent unsafe accesses and users would refer to queries either by their type signature or entity id. 
```

Trashtalk217 (commented):
A couple of language things that can be found with any appropriately smart spell checker.

Anyway, I can say that I understand the document now, which is good news. One thing I do like to clear up on thing about the order of operations. All of the caching improvements (component index, access, the `ArchetypeComponentId` cache) technically don't need to wait on #12144, right?

Trashtalk217 (approved):
The language is still quite dense, some things could've easily been expanded upon to make it read clearer, and there probably are still a couple of spelling errors in there.

BUT with a topic as deep and complex as an ecs, we could keep adding to it forever. And for the good of starting development, I think this rfc is in a good state.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,128 @@
+# Feature Name: `minimal-fragmenting-relationships`
+
+## Summary
```
alice-i-cecile, line 3: The writing here is pretty solid, but it's all motivation. I'd like to see a bullet-point list of "here are the steps we need to take" here.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,128 @@
...
+
+Another analogy when considering the ECS as an in memory database is that relationships act as foreign keys. This unlocks a potential future where we can express more complex queries matching across multiple entities and relationship pairs. There are a myriad of additional features that can be built on top of the initial implementation described below. The main focus in this RFC will be the blocking issues that must be addressed before the initial implementation is achievable.
+## Motivation
+
+Relationships have been a long desired bevy feature. The ability to express and traverse hierarchies and graphs within the ECS unlocks a huge amount of potential. Features such as breadth-first query traversal, immediate hierarchy clean-up, component inheritance, multi-target queries etc. all rely on or are simplified by the existence of relationships as a first class ECS primitive. 
+## User-facing explanation
+
+Relationship pairs can be inserted and removed like any other component. There are 3 ids associated with any relationship pair instance:
```
alice-i-cecile, line 15: This needs a bit more introduction. What is a "relationship pair"?

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,128 @@
...
+2. The relationship component
+3. The target entity
+
+In the minimal implementation all relationships will take the form of `(ZST Component, entity)` pairs so the API surface will look quite familiar. All naming is subject to bikeshedding:
+```rust
+#[derive(Component)]
+struct Relationship;
+
+world.entity_mut(source_entity_a)
+	.insert_pair::<Relationship>(target_entity_b)
```
alice-i-cecile, line 26: I'm not keen on the naming here. A "pair" is a new concept / terminology that I don't think actually adds any expressive value over `insert_relation`.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,128 @@
...
+	.with_pair::<Eats>(apple)
+	.build();
+
+query.single(&world); // == alice
+```
+As relationship pairs are effectively regular components this doesn't require any changes to the more complex query iteration APIs such as `par_iter`.
+
+## Implementation strategy
+
+There are a number of technical blockers before even a minimal version of relationships could be implemented in bevy. They're individually broken down below:
```
alice-i-cecile, line 79:
```suggestion
There are a number of technical blockers before even a minimal version of relationships could be implemented in Bevy. They're individually broken down below:
```

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,128 @@
...
+- New methods for `EntityCommands`, `EntityMut`/`Ref`/`WorldMut` and `QueryBuilder`
+## Drawbacks
+- Increased archetype fragmentation when used, this can have impacts on memory fragmentation and query iteration performance, however the impact is non-trivial and varies from case to case
+- Increase in internal and API complexity
+## Prior art
+- Prior [RFC](https://github.com/BoxyUwU/rfcs/blob/min-relations/rfcs/min-relations.md)
+- [flecs](https://github.com/SanderMertens/flecs/blob/master/docs/Relationships.md)
+- Partial draft implementation: [#9123](https://github.com/bevyengine/bevy/pull/9123)
+## Unresolved questions
+- How to address [#12144](https://github.com/bevyengine/bevy/issues/12144)
```
alice-i-cecile, line 119:
```suggestion
- How to address [the render world blocking ECS developments](https://github.com/bevyengine/bevy/issues/12144)
```

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,128 @@
...
+## Future possibilities
+- Data on relationship pairs is a trivial extension:
+  ```rust
+  world.entity_mut(source).insert_pair(Relationship {}, target);
+  ```
+- Clean-up policies to allow specifying recursive despawning, this is important to allow porting `bevy_hierarchy` to use relationships
+- More advanced traversal methods: up, BFS down, etc.
+- More expressive query types: multi-target, grouping, sorted etc.
+- With component as entities we unlock full `(entity, entity)` relationships
```
alice-i-cecile, line 128: This needs more explanation.

alice-i-cecile (changes requested):
I think I'm broadly on board with the technical implementation strategy, but the writing here needs work.

There are two key problems in this document (and the surrounding discussion):

1. Jargon from `flecs` is used freely without definition. "pairs" and (entity, entity) relationships are the key issues here.
2. We jump straight to a solution, without clearly laying out the problems to be solved, or how the problems are related to the solution. This is most evident in the section on Archetype Cleanup, but needs work throughout.

We can reduce the amount of space spent talking about "relations are great" (just link to my issue), but a clear "here are the problems, here's how we can solve them" structure will go a long way to making this easy for others to understand.

In terms of API design, I have two points of feedback:

1. Can we avoid introducing the "pairs" terminology completely?
2. I would like to cover the basic API by which ZST relations can be queried (or explicitly mark it as out of scope).

rfcs/79-minimal-fragmenting-relationships.md
james-j-obrien: Very fair, this is me bringing in baggage from other languages that defined two item tuples as pairs. I think the main reason I went that way over `insert_relation` is just how long that is for a very common operation that already needs a type parameter.

rfcs/79-minimal-fragmenting-relationships.md
james-j-obrien: Yes this could be more clear, it's only implicitly defined in the summary.

rfcs/79-minimal-fragmenting-relationships.md
SanderMertens: Fwiw the reason I picked pair is because it's less confusing/opinionated:
- Does relation(ship) refer to the first element of the pair or both the relationship and the target?
- The feature is called relationships, but it really implements the ability to have composite component ids which can be used for relationships. I can use it to add a `(Dog, Food)` pair, where `Dog` isn't really a relationship and `Food` isn't really a relationship target.

alice-i-cecile: I can get behind that. The key thing here is clearly laying out (and motivating) the terminology EDIT: I've changed my mind here.

zesterer:
If it's useful in any way, we have some prior art on this over at [Veloren](www.veloren.net) that's been in use for a number of years and has proven itself under a number of different situations.

We call it the ['link system'](https://gitlab.com/veloren/veloren/-/blob/master/common/src/link.rs?ref_type=heads), and it's a generic framework for defining and maintaining relationships between ECS entities. It supports both 1:1 and 1:many relationships.

Currently, it is used to define:

- Rider/mount relationships (like a player riding a horse)
- Volume entity riding relationships (many players sat on seats in an airship, moving through the air)
- Tethering (physics-based ropes that attach entities to one-another, or to the terrain)

I'm happy to answer any questions about it.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,145 @@
+# Feature Name: `minimal-fragmenting-relationships`
+
+## Summary
+Users often run into difficulty while trying to model entity hierarchies in Bevy. The tools we currently have are not expressive enough to meet their needs. See this issue for more discussion [#3742](https://github.com/bevyengine/bevy/issues/3742).
+
+This RFC proposes a minimal design that would unblock a large amount of users while also leaving avenues for the engine to further expand these features. In short:
+- Define a new variant of `Identifier` that combines a `ComponentId` (the relationship type) and an `Entity` (the target) that can be added to an entity (the source) to model a relationship of that type between the source and the target.
```
alice-i-cecile, line 7: I'm leary of this design in general: I've found discussions about flecs features that lean on this feature (and components-as-entities) rather confusing to follow. While I've gotten more comfortable with it over time, I worry that new users will still struggle to understand what's going on.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,145 @@
...
+- Implement command, world and query APIs to support adding, removing and reading these new relationships.
+
+A large amount of internal refactoring has to be done in order to support this API so a significant focus will be on eliminating these technical blockers. For example:
+- Replacing internal `SparseSet` usage where `Identifier` would be used as keys as they are no longer dense.
+- Allowing archetypes to be deleted when an entity that was used as part of a pair in that archetype is despawned. This involves updating several internal caches that are used for correctness and/or safety.
+
+## Motivation
```
alice-i-cecile, line 14: This section needs to be rewritten to take into account the newly merged non-fragmenting relations.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Terminology
+
+- **[Identifier](https://docs.rs/bevy_ecs/latest/bevy_ecs/identifier/struct.Identifier.html)**: A globally unique ECS id represented as a `u64`.
+- **[Entity](https://docs.rs/bevy_ecs/latest/bevy_ecs/entity/struct.Entity.html)**: A variant of `Identifier` representing an entity.
+- **Entity Id**: The lower 32 bits of an `Entity`, no two living entities should share the same id.
+- **Generation**: The upper 32 bits of an `Entity`, represents how many times the id has been re-used, important to prevent interpreting a recycled `Entity` as the original.
+- **Relationship**: A `Component` type used to model a relationship between two entities i.e. `Eats`, takes up the upper 32 bits of a `Pair`
+- **Target**: An entity used as a target of a relationship, it's id takes up the lower 32 bits of a `Pair`
+- **Pair**: A variant of `Identifier` representing a relationship to a target entity i.e. `(Eats, apple)` where `apple` is the target entity
```
alice-i-cecile, line 27: I will veto this name. It is extremely unclear and too generic.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,145 @@
...
+- **Source**: The entity a `Pair` is added to as a component i.e. if `(Eats, apple)` is added to entity `alice`, `alice` would be the source of that relationship.
+- **ZST**: Zero-sized type, these are types with `std::mem::size_of::<T>() == 0` and hence carry no data.
+- **Wildcard**: A special id used to match all any id, similar to a glob i.e. `(Eats, *)` would match all ids where the relationship type was `Eats`
+## User-facing explanation
```
alice-i-cecile, line 31:
```suggestion

## User-facing explanation
```

alice-i-cecile, line 31: This also needs to be rewritten in the wake of merging non-fragmenting relations.

alice-i-cecile: There's a bunch of missing newlines around section headers actually :)

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,145 @@
...
+##### Query and System Caches
+Query's caches of matching tables and archetypes are updated each time the query is accessed by iterating through the archetypes that have been created since it was last accessed. As the number of archetypes in bevy is expected to stabilize over the runtime of a program this isn't currently an issue. However with the increased archetype fragmentation caused by fragmenting relationships and the new wrinkle of archetype deletion the updating of the query caches is poised to become a performance bottleneck.
+
+An appropriate solution would allow us to expose new/deleted archetypes to only the queries that may be interested in iterating them. Luckily observers [#10839](https://github.com/bevyengine/bevy/pull/10839) could cleanly model this pattern. By sending ECS triggers each time an archetype is created and creating an observer for each query we can target the updates so that we no longer need to iterate every new archetype. Similarly we can fire archetype deletion events to cause queries to remove the archetype from their cache.
```
alice-i-cecile, line 102: No objections to this: supporting archetype cleanup is a valuable optimization no matter what we do.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,145 @@
...
+
+In order to implement such a mechanism we need some way for the observer to have access to the query state. Currently `QueryState` is stored in the system state, so in order to update it we would need to expose several additional APIs, even then it's quite difficult to reach into the system state tuple to access the correct query without some macros to handle tuple indexing. An alternative would be to move the query state into a component on an associated entity (WIP branch here: [queries-as-entities](https://github.com/james-j-obrien/bevy/tree/queries-as-entities)) this provides additional benefits in making it easier manage dynamic queries, but they are largely tangential to this RFC.
```
alice-i-cecile, line 104: I think that moving queries and systems out onto their own entities is likely to be a useful long-term route. The ECS is very powerful and expressive, and improving flexibility and introspection is valuable.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,145 @@
...
+
+Regardless this will also require further refactors to the query cache implementation. For example, currently the collection of tables and archetypes is a vector but that doesn't provide an efficient method for removal without an O(n) scan, so this data structure will have to be replaced with one that performs reasonably for both iteration and insertion/removal. Similarly systems cache their `ArchetypeComponentId` accesses for the purposes of multithreading, we also need to update theses caches to remove deleted ids presenting many of the same problems. 
+
+Using observers globally or moving queries to entities are both blocked by the deletion of entities in the render world tracked here: [#12144](https://github.com/bevyengine/bevy/issues/12144).
```
alice-i-cecile, line 108: No longer true!

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,145 @@
...
+The `FixedBitSet` implementation used by `Access` inside of systems, system params, queries etc. relies on dense ids to avoid allocating a large amount of memory. Since pairs have component ids in the upper bits the `u64` they are represented as is extremely large, each `FixedBitSet` would allocate non-trivial amounts of memory to have flags for all the intervening ids. In order to minimize the affects of this we could consider all `(R, *)` pairs as one id in accesses, reducing the amount we need to track at the expense of some theoretical ability to parallelise.
+
+Similarly the `SparseSet` implementation used in both table and sparse component storage also operate under the assumption that ids will remain low and would allocate an even larger amount of memory storing pairs on an entity. In order to circumvent this we can implement a "component index", this index would track for each `Identifier` the tables and archetypes it belongs to as well as the column in that table the id inhabits (if any). This removes the need for the sparse lookup as well as providing opportunities for other optimizations related to creation of observers and queries.
```
alice-i-cecile, line 112: I think this is worth pursuing.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Minimal Implementation
+The component index could be implemented at any time, however the other pieces require some ordering to satisfy dependencies or prevent half implemented features.
+Implementing the remainder in an appropriate order of dependance would work as follows:
```
alice-i-cecile, line 116: This needs to be updated.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,145 @@
...
+- Merge observers [#10839](https://github.com/bevyengine/bevy/pull/10839)
+- Resolve [#12144](https://github.com/bevyengine/bevy/issues/12144) so entities can be persisted in the render world.
+- Refactor queries to become entities with caches updated via observers
+- Implement archetype deletion and the associated clean-up for systems and queries, this is the only change that pays a complexity cost for no user benefit until relationships are realised
```
alice-i-cecile, line 120: IMO we should expose a manual "cleanup empty archetypes method" for power users even at this stage.

rfcs/79-minimal-fragmenting-relationships.md
```diff
@@ -0,0 +1,145 @@
...
+- How to address [the render world blocking ECS developments](https://github.com/bevyengine/bevy/issues/12144)
+## Future possibilities
+- Data on relationship pairs is a trivial extension:
+  ```rust
+  world.entity_mut(source).insert_pair(Relationship {}, target);
+  ```
+- Clean-up policies to allow specifying recursive despawning, this is important to allow porting `bevy_hierarchy` to use relationships
+- More advanced traversal methods: up, BFS down, etc.
+- More expressive query types: multi-target, grouping, sorted etc.
+- With component as entities we unlock the ability `(entity, entity)` pairs, where either entity could also be a component. This makes it much easier to express dynamic relationships where you don't necessarily want to define your relationship type statically or go through the trouble of creating a dynamic component, as well as cases where you want a component as the target.
```
alice-i-cecile, line 143: I am broadly opposed to this feature in the proposed form. I don't think that the clarity/complexity to utility tradeoff is worth it over more narrow forms.

rfcs/79-minimal-fragmenting-relationships.md
iiYese: That makes an unrealistic assumption users will be introduced to all of the ECS at once. Instead of say developing an understanding of components + entities & later learning that "oh components are just entities too". If anything I don't think it's a "new user" issue I think it takes more effort for you to grasp *precisely because* you have far more activity with bevy. I like many others, with far less development time on bevy have no issues grasping the feature.

alice-i-cecile:
Closing this out: I don't think that editing this is a particularly good way to get to an implementable design doc. Too much has changed, there's too much history here, and there's too many implicit assumptions about the path we should take. I'm not at all opposed to archetypal relations (and actively want to support things like large numbers of archetypes and clearing empty tables and archetypes), but I want to see a user-focused path / design for how and why Bevy should get there.

PR #83, opened 2024-09-19
Add `reflect_generic_id` rfc

JohnTheCoolingFan:
Rendered: https://github.com/JohnTheCoolingFan/bevy-rfcs/blob/reflect_generic_id/rfcs/reflect_generic_id.md

Make type identification in the Reflection system generic, allowing multiple identification data types to coexist.

### DIFF
--- a/rfcs/reflect_generic_id.md
+++ b/rfcs/reflect_generic_id.md
@@ -1,0 +1,168 @@
+# Feature Name: `reflect_generic_id`
+
+## Summary
+
+Make type identification in the Reflection system generic, allowing multiple identification data types to coexist.
+
+## Motivation
+
+Various use cases of reflection have different requirements for dynamic type identification methods and different types have different ups and downs.
+
+## User-facing explanation
+
+> Explain the proposal as if it was already included in the engine and you were teaching it to another Bevy user. That generally means:
+> 
+> - Introducing new named concepts.
+> - Explaining the feature, ideally through simple examples of solutions to concrete problems.
+> - Explaining how Bevy users should *think* about the feature, and how it should impact the way they use Bevy. It should explain the impact as concretely as possible.
+> - If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+> - If applicable, explain how this feature compares to similar existing features, and in what situations the user would use each one.
+
+Comparisons of previously used identification methods and their pos and cons:
+- A UUID would be universally unique, but require manually specifying one for each reflected type in source code. Automatically creating one at compile time results in equivalency with TypeId, if a random one is generated. If some information about a type such as name and fields/variants is used as randomness seed, it would be somewhat equivalent to type name or type path.
+- A type name can be automatically derived but can cause collisions when types from multiple modules are combined in a registry.
+- A `core::any::TypeId` identifies a type uniquely, but only for a single compiled binary, meaning it could change when source code is changed or in dynamically loaded binaries.
+
+## Implementation strategy
+
+> This is the technical portion of the RFC.
+> Try to capture the broad implementation strategy,
+> and then focus in on the tricky details so that:
+> 
+> - Its interaction with other features is clear.
+> - It is reasonably clear how the feature would be implemented.
+> - Corner cases are dissected by example.
+> 
+> When necessary, this section should return to the examples given in the previous section and explain the implementation details that make them work.
+> 
+> When writing this section be mindful of the following [repo guidelines](https://github.com/bevyengine/rfcs):
+> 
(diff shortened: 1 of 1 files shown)


SkiFire13:
I don't think this will solve the problem. Even if you manage to get some `TypeData` for a type, if the `TypeId` does not match the `TypeData` will fail to downcast and call the given method on the concrete type. And if you use anything not based on `TypeId` (e.g. UUIDs) in there too, then it will likely be unsound, since you have no way to guarantee that if a type is changed (even transitively, e.g. one of its field types is changed) then its UUID is changed too.

JohnTheCoolingFan:
These requirements would probably be part of the trait implementation requirements. `TypeInfo` and `TypeData` sets could probably be hashed and used to compare, but also there are places where using only `TypeId` for identification is not going to work. For example, if deserializing a set of data, a user wants to have something similar to an enum, but the set of keys is dynamic and based on what types are registered in a registry used for the de/serialization operation. Custom key types would be handy, and restricting it to type names and/or paths isn't making it much better.
`TypeRegistration` already contains a `TypeId` and could be used for checks. And if `TypeId`s don't match but the registry keys do, that might mean that the registration is from another binary/compilation and a conversion could be attempted using `FromReflect` and dynamic structs: `clone_dynamic` to get a dynamic struct from source data, `FromReflect` to convert to concrete type with expected `TypeId`.

Apologies if this comes off a bit chaotic. I might have only a set of specific applications in mind and also maybe some misunderstandings about the current reflect implementation and uses/connections.

JohnTheCoolingFan:
`TypeId` is a sound identification type in one library, and this guarantee is provided by the rust compiler. But, when loading from another binary (plugin) or loading data from a serialized data set (scene), other identification types are needed. It may be possible to still keep these identifications sound, have checks. For example, by using some kind of footprint, that is encoding the properties of the reflected type. A stable type id that hashes even package versions, features and all of the compilation context would encode too much for the use case.
And it may be possible to create mappings from one id type to another. For example, to identify the type that is basically identical but in two binaries has a different typeid and potentially different data layout. If there is an identification type that works for signaling across these two binaries, a mapping could be made for each one. And when data inevitably has to go from one to the other, a serialized common representation of that data can be sent, then deserialized on the other end.

SkiFire13:
> `TypeInfo` and `TypeData` sets could probably be hashed and used to compare

You likely don't want to hash `TypeData`s, since they don't have a canonical representation (e.g. the function pointers inside them might be duplicated in different codegen units, so two `TypeData`s for the same type might actually compare different).

`TypeInfo` contains `TypeId` so you're going to have the same problems as using `TypeId`.

> And if `TypeId`s don't match but the registry keys do, that might mean that the registration is from another binary/compilation and a conversion could be attempted using `FromReflect` and dynamic structs

This cannot be done automatically on-the-fly when a `TypeData` is accessed, since it might be returning borrowed data or mutating the data passed in, and those won't work if you pass a temporary object that just got converted.

It might work if you perform this conversion the moment you create some type, essentially canonicalizing that type, but then you're just performing deserialization using reflection and type paths should be enough for that.

> It may be possible to still keep these identifications sound, have checks. For example, by using some kind of footprint, that is encoding the properties of the reflected type. A stable type id that hashes even package versions, features and all of the compilation context would encode too much for the use case.

My point is that for many usecases this cannot be just a possibility, it has to be a guarantee or you have unsoundness.

rfcs/83-reflect_generic_id.md
```diff
@@ -0,0 +1,168 @@
...
+        core::any::type_name::<T>()
+    }
+}
+```
+
+## Drawbacks
+
+> Why should we *not* do this?
+
+This introduces a concept of multiple type identifications which might have no connection between them. There would be no singular central type registry, but instead a number of registries, each with a different id type and for a different purpose.
```
MrGVSV, line 125: I think this is probably one of the biggest issues with this approach. If we end up going with a generic type identification system, we need a way to interop between them such that we can have a central registry.

If we don't have that, then the onus to use the correct registry/identification falls on the user.

MrGVSV:
> I don't think this will solve the problem. Even if you manage to get some `TypeData` for a type, if the `TypeId` does not match the `TypeData` will fail to downcast and call the given method on the concrete type. And if you use anything not based on `TypeId` (e.g. UUIDs) in there too, then it will likely be unsound, since you have no way to guarantee that if a type is changed (even transitively, e.g. one of its field types is changed) then its UUID is changed too.

Personally, I'd like to know more about how we intend to support cross-binary types. Are we passing around raw pointers, utilizing the dynamic types, allowing specially-made custom dynamic types, serializing-deserializing, or something else? Depending on how this is done, "downcasting"[^1] should be possible through type data.

In any case, it would be good to identify these particular methods so we can properly assess whether a generic identification system truly covers those cases, or if another solution needs to be explored.

[^1]: I put "downcasting" in quotes because—depending on how things are done—downcasting might be to the dynamic type and not actually the represented type.

PR #85, opened 2025-06-23
Add `bevy_audio_graph` RFC

eira-fransham:
[Rendered](https://github.com/eira-fransham/bevy-rfcs/blob/bevy-audio-graph/rfcs/audio-graph.md).

This RFC proposes a new system to handle audio, leveraging relationships and the ECS to define how audio is routed, and the Web Audio API (potentially implemented on desktop with the [`web-audio-api`](https://crates.io/crates/web-audio-api) crate) to ensure high performance both natively and on the brower.

### DIFF
--- a/rfcs/audio-graph.md
+++ b/rfcs/audio-graph.md
@@ -1,0 +1,385 @@
+# Feature Name: `bevy_audio_graph`
+
+## Summary
+
+A more-advanced graph-based audio system for Bevy, leveraging relationships to define the graph, a
+separate schedule to build and drive the audio, and the Web Audio API to create parity between
+desktop and web applications.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


rparrett:
> Once a sound has been queued for playback, you lose control over it. You can stop the sound by deleting
the `AudioPlayer` component, but there is no way to change the speed of playback, pause and resume,
set a new location to play from, et cetera.

This section doesn't ring true: https://bevy.org/examples/audio/audio-control/

> alternatives

The "[better audio working group](https://discord.com/channels/691052431525675048/1236113088793677888)" is cooking up the following, under the same premise that abandoning rodio is the way forward.

https://github.com/corvusprudens/bevy_seedling
https://github.com/BillyDM/firewheel

eira-fransham:
@rparrett Thank you! I'm now switching my project to Seedling and in communication with the developer to start contributing.

